PR #30, opened 2024-01-21
Update README.md

### LINKED ISSUE #29: Error in code in README
If you follow all of the instructions in the README you will get an error like "callModel is not defined". This is because the function is used but not defined.

Following [this example](https://github.com/langchain-ai/langgraphjs/blob/main/examples/chat_agent_executor_with_function_calling/base.ipynb), if I add the following code it works as expected.

```ts
// Define the function that calls the model
const callModel = async (state: { messages: Array<BaseMessage> }) => {
  const { messages } = state;
  const response = await newModel.invoke(messages);
  // We return a list, because this will get added to the existing list
  return {
    messages: [response],
  };
};
```

logan-anderson:
Fixes #29

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,16 @@ const _getAction = (state: { messages: Array<BaseMessage> }): AgentAction => {
   };
 };
 
+// Define the function that calls the model
+const callModel = async (state: { messages: Array<BaseMessage> }) => {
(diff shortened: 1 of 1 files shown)


bracesproul (approved):
Thanks for adding this and updating the langsmith link!

PR #41, opened 2024-02-01
Replace Tool by StructuredTool in tool_executor.ts

### LINKED ISSUE #40: ToolExecutor should probably use an array of StructuredTool instances rather than Tool instances
While trying to use a `DynamicStructuredTool` instance with a `ToolExecutor`, I noticed that TS was yelling at me.

```ts
const tools = [
  new DynamicStructuredTool({
    name: "search",
    description: "Search the web",
    schema: z.object({
      input: z.string(),
    }),
    callbacks: [toolRunCollector],
    async func({ input }, runManager) {
      console.log("searching for", input);
      return Promise.resolve("whatever");
    },
  }),
];

const toolExecutor = new ToolExecutor({
  tools // TypeScript is yelling at me here
});
```

I suspect that the `ToolExecutorArgs` interface should be:

https://github.com/langchain-ai/langgraphjs/blob/8ee245f992443c8d7efc0b505451a6cd5cc75504/langgraph/src/prebuilt/tool_executor.ts#L10-L16

```ts
import { StructuredTool } from "@langchain/core/tools";
// ...
export interface ToolExecutorArgs {
    tools: Array<StructuredTool>;
    /**
     * @default {INVALID_TOOL_MSG_TEMPLATE}
     */
    invalidToolMsgTemplate?: string;
}
```

As [`Tool` extends `StructuredTool`](https://api.js.langchain.com/classes/langchain_core_tools.StructuredTool.html) (which is quite weird in my book).

zeropaper:
Because the `Tool` abstract class extends the `StructuredTool` abstract class, using a `DynamicStructuredTool` would not be usable in the `ToolExecutor`.

https://api.js.langchain.com/classes/langchain_core_tools.StructuredTool.html 

Fixes #40 

~~**Note:** I did not clone the repo locally or tested that.~~

### DIFF
--- a/langgraph/src/prebuilt/chat_agent_executor.ts
+++ b/langgraph/src/prebuilt/chat_agent_executor.ts
@@ -1,4 +1,4 @@
-import { Tool } from "@langchain/core/tools";
+import { StructuredTool } from "@langchain/core/tools";
 import { convertToOpenAIFunction } from "@langchain/core/utils/function_calling";
 import { AgentAction } from "@langchain/core/agents";
 import { FunctionMessage, BaseMessage } from "@langchain/core/messages";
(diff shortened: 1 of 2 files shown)


bracesproul (changes requested):
Hey, just making these changes will cause type errors in many places in the codebase. It would be great if you could fork and get it all working locally, then push up a branch! Thank you

zeropaper:
@bracesproul my bad, I was in the middle of something else, sorry. 
I did it the right way this time, hope that it floats this time.

bracesproul (approved):
Sorry for the delay, getting this merged in now! Thanks!

PR #46, opened 2024-02-02
Pass runnable config through lambda calls for tracing

jacoblee93:


jacoblee93:
CC @bracesproul @nfcampos

nfcampos (approved):

PR #83, opened 2024-03-10
The retrieve method adds the missing parameter config

luocongqiu:
image
Missing parameter config


bracesproul (approved):
Good catch! Thank you!

PR #92, opened 2024-03-21
fix: default INVALID_TOOL_MSG_TEMPLATE

ddosakura:
The default template in tool_executor is incorrect and has not properly replaced availableToolNamesString.


jacoblee93:
Thank you!

PR #93, opened 2024-03-21
Added "should handle checkpoints correctly" Unit Tests + Fix 2 Bugs That Were Preventing Them From Passing

janvi-codaio:
The `should handle checkpoints correctly` unit test was initially skipped because it didn't pass. 
This PR fixes the two bugs that were preventing it from passing. 

1. `BinaryOperatorAggregate` Channel did not implement `empty()` (aka - snapshot) correctly. The issue was that the old checkpoint's value was not being saved as the initial value leading to a lossy channel that stored state in correctly 

2. `Checkpoint` was not implemented correctly, because it would only update the checkpoint's channel value if it had never been set before. Updated the code such that it _always_ update's the checkpoint value based on the latest channel value 

Now that these have been fixed, add the checkpointing test back in!

### DIFF
--- a/langgraph/src/channels/base.ts
+++ b/langgraph/src/channels/base.ts
@@ -93,7 +93,7 @@ export async function createCheckpoint<Value>(
       newCheckpoint.channelValues[k] = await channels[k].checkpoint();
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
     } catch (error: any) {
-      if ("name" in error && error.name === EmptyChannelError.name) {
+      if (error.name === EmptyChannelError.name) {
         // no-op
       } else {
         throw error; // Rethrow unexpected errors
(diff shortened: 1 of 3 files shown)


langgraph/src/channels/binop.ts
```diff
@@ -61,7 +64,7 @@ export class BinaryOperatorAggregate<Value> extends BaseChannel<
   }
 
   public checkpoint(): Value {
-    if (!this.value) {
+    if (this.value === undefined) {
```
janvi-codaio, line 67: nit for consistency with the rest of the file

nfcampos (approved):

langgraph/src/channels/base.ts
```diff
@@ -88,17 +88,15 @@ export async function createCheckpoint<Value>(
...
-          // no-op
-        } else {
-          throw error; // Rethrow unexpected errors
-        }
+  for (const k of Object.keys(channels)) {
+    try {
+      newCheckpoint.channelValues[k] = await channels[k].checkpoint();
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } catch (error: any) {
+      if ("name" in error && error.name === EmptyChannelError.name) {
```
jacoblee93, line 96: Worth simplifying into one conditional?

janvi-codaio: good point - done!

PR #101, opened 2024-03-26
Add diagram for agent executor example

ric9176:
This PR adds a basic diagram for the agent executor example. I've used the same colors and style as in other diagrams and lined up the naming with the code.

I just thought I'd PR this as I sketched it while working through the examples. 

Using a diagrams as code solution such as [mermaid](https://mermaid.js.org/) could be a good way to iterate more easily on diagrams and make them all uniform in terms of styling 😀

Really enjoying the work the team is doing here!


jacoblee93:
Neat, thank you!

PR #109, opened 2024-04-06
Type-safety of Channel.subscribeTo

functorism:
Ensure `Channel.subscribeTo(["input"], { key: "key" })` which throws at runtime is a type-error.


bracesproul (commented):
This looks reasonable to me. cc @nfcampos for a second look

nfcampos (approved):
Thanks for this

PR #110, opened 2024-04-06
Include Pregel PDF link in CONTRIBUTING.md

functorism:
Just make it one tiny bit easier for people curious about Pregel to read the paper, since the current Google link only goes to ACM which requires access, but CMU provides the paper.


bracesproul (approved):
thanks for this!

Issue #88, opened 2024-03-19
LangGraph throws an error when adding edges that don't exist, but that makes writing the code a bit messy


davidfant:
Here LangGraph throws an error when trying to add an edge between nodes that don't exist:
https://github.com/langchain-ai/langgraphjs/blob/56d3d28f45748ee4254a3e658d4f0db450a8d64d/langgraph/src/graph/graph.ts#L90-L95

That **makes sense** but it makes my code that constructs the graph a bit confusing to read. This isn't the case for addConditionalEdge (unless providing a mapping).

Example for how I ideally want to set up the graph
```ts
agent.addNode("first", ...);
agent.addEdge("first", "second");

agent.addNode("second", ...);
```

Counter point: why can't I set up edges after all nodes. I can, but if a graph gets big, it's annoying and confusing to have to scroll a lot to figure out how everything is connected. I want to read the logic/paths from top to bottom, rather than just nodes and then just edges. Maybe this is just me and no one else having this problem.

What are your thoughts reg moving this validation to the compile step instead?

jacoblee93:
@nfcampos to weigh in!

nfcampos:
@davidfant see improved typings added in #128 which would potentially make what you want harder? That said I'll be moving the runtime validation itself to compile() to allow you to build graphs "out of order" if you opt out of the static validation

andrewnguonly:
> What are your thoughts reg moving this validation to the compile step instead?

@jacoblee93, what's your opinion?

I'm slightly against this, but I don't know why and I don't have a strong reason. Pushing business logic validation to a compile/static check does not seem desirable IMHO if it's at the expense of increased developer maintenance. The implementation may get more complex when supporting more complicated graph features (dynamic graphs).

I think a better solution would be to allow the developer to add nodes and edges in any order (i.e. just aggregate them in a list). Then build the graph internally in topological sorted order.

hinthornw:
It could be nice to have something where it creates the node and adds an edge at the same time (for simple edges), like `node.to(other_node)`, however.

jacoblee93:
Hey sorry for missing the ping!

@andrewnguonly I don't feel super strongly but I have seen adjacency lists in the wild as a common way to declare graphs in general which naturally imply declaring edges before declaring all nodes:

```ts
1: 2, 3
2: 1
3: 1
```

It does increase the likelihood of something being misrouted if a name changes, but I lean towards @davidfant's comments here around convenience.

[...]

PR #112, opened 2024-04-06
Add a graph builder

functorism:
This PR shows one possible way to construct a graph builder.

This graph builder turns what is now run-time errors during graph construction into type-errors. Meaning users get the same error feedback directly in their IDE as type errors without having to run the code. There is also the added benefit that they get some level of "intellisense" while writing the code.

The builder has the same API as `Graph`, with the addition of a `.done()` method and stricter input types.

One can peruse the [test file](https://github.com/functorism/langgraphjs/blob/graph-builder/langgraph/src/tests/graph-builder.test.ts) for examples of how to use the builder and what type of errors it catches at "compile-time".

The builder works via the fact that methods return the `this` instance with additional type information:

```typescript
  // Adds a node to the graph and returns a builder that's aware of that nodes existence
  addNode<K extends string>(
    this: GraphBuilder<RunInput, RunOutput, Nodes>,
    key: K,
    action: RunnableLike<RunInput, RunOutput>
  ): GraphBuilder<RunInput, RunOutput, Nodes | K> {
    super.addNode(key, action);
    return this;
  }
```

Which allows us to catch simple mistakes like:

```typescript
new GraphBuilder()
  .addNode("a", () => {})
  .addNode("b", () => {})
  // This is fine
  .addEdge("a", "b")
  // @ts-expect-error - this must type-error
  .addEdge("b", "c");
```

```typescript
new GraphBuilder()
  .addNode("a", () => {})
  .addNode("b", () => {})
  .addConditionalEdges("b", () => "next", {
    continue: "b",
 // ^^^^^^^^ - The type error will be localized here
 // Error: Object literal may only specify known properties,
 // and 'continue' does not exist in type 'Record<"next", "b">'
  });
```

The builder doesn't guarantee perfect safety due to limitations of what TypeScript's type system can express; but it does make it a lot easier to _do the right thing_.

A clear example of these limitations is that we have no way to "invalidate" references to "old" builder instances; if one creates ["forks" of the builder](https://github.com/functorism/langgraphjs/blob/graph-builder/langgraph/src/tests/graph-builder.test.ts#L77-L90), you'll run into unexpected behavior.

The code duplication for `GraphBuilder` and `StateGraphBuilder` is inelegant, but can be improved if this idea is deemed appealing!

Having access to a graph builder that provides "at construction time checking" sweetens the deal if one would want to move the runtime checks that exists in `Graph` into the `compile` function entirely. As suggested in this issue https://github.com/langchain-ai/langgraphjs/issues/88. If construction time checking is a type-level concern and compile time checking a runtime one; then it's possible to support both use cases. Making it possible to support construction methods that only have to be valid at compile time - then order of node and edge inserts does not matter. It only matters that the state is correct when calling compile. And for those wanting construction time validation (and "intellisense"), then the builder is available.

Thanks @davidfant for reviewing and helping to find a bug prior to opening this.

### DIFF
--- a/langgraph/src/graph/builder.ts
+++ b/langgraph/src/graph/builder.ts
@@ -1,0 +1,252 @@
+import { RunnableLike } from "@langchain/core/runnables";
+import { END, Graph } from "./graph.js";
+import { Pregel } from "../pregel/index.js";
+import { BaseCheckpointSaver } from "../checkpoint/index.js";
+import { START, StateGraph } from "./state.js";
+
+/**
+ * @experimental This class is experimental
+ */
+export class GraphBuilder<
+  RunInput = unknown,
+  RunOutput = unknown,
+  const Nodes extends string = typeof END,
+  Compiled extends true | void = undefined
+> extends Graph<RunInput, RunOutput> {
+  private strictCompiled: Compiled;
+
+  private compileResult: Compiled extends true ? Pregel : never;
+
+  // Adds a node to the graph and returns a builder that's aware of that nodes existence
+  addNode<K extends string>(
+    this: GraphBuilder<RunInput, RunOutput, Nodes>,
+    key: K,
+    action: RunnableLike<RunInput, RunOutput>
+  ): GraphBuilder<RunInput, RunOutput, Nodes | K> {
+    super.addNode(key, action);
+    return this;
+  }
+
+  // `startKey` and `endKey` must be existing node keys
+  addEdge<SK extends Nodes, EK extends Nodes>(
+    this: GraphBuilder<RunInput, RunOutput, Nodes>,
+    startKey: SK,
+    endKey: EK
+  ): GraphBuilder<RunInput, RunOutput, Nodes> {
+    super.addEdge(startKey, endKey);
+    return this;
+  }
+
+  // `startKey` and return key of `condition` must be existing node keys
+  addConditionalEdges<SK extends Nodes>(
+    this: GraphBuilder<RunInput, RunOutput, Nodes>,
+    startKey: SK,
+    condition: (...args: unknown[]) => PromiseLike<Nodes> | Nodes
+  ): GraphBuilder<RunInput, RunOutput, Nodes>;
+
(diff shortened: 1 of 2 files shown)


davidfant:
Nice! I've kind of done something similar locally, but having better types be part of LG would be a better solution

bracesproul (commented):
This looks super interesting! I'll try to get around to reviewing it by EOW

Thanks for making a PR for this!

nfcampos:
@functorism thanks a lot for this, can we just fold this into the main Graph class? and instead of `done()`, we already have `compile()`?

functorism:
@nfcampos That sounds like a good idea to me overall, but I guess that could leave the ability to build the graph out of order - on the table so to say. Supporting both might be a understandable maintenance burden however.

I guess the main reason to consider this is if one would imagine that being able to build algorithms that constructs graphs could be a powerful feature. Like program synthesis. However, due to an implementation detail of the current graph builder, this could be achieved already. If the graph builder is instantiated with the type `Nodes = string`, then you would release the constraints currently enforced and could build out of order. But the fact that this works is potentially already a fragile weakness of the current formulation, albeit perhaps pragmatic for this very reason.

nfcampos:
Not sure I understand why this would be a downside in the main Graph class and not the exact same downside in a separate GraphBuilder class?

functorism:
I guess the setup could be that the graph builder is what gives you validation _while_ constructing the graph, and you don't get that in the Graph class. Thus `Graph` allows building out of order.

* Remove runtime checks for the methods that builds the class in `Graph`, and have them in the builder or rely only on the type level.
* Compile always performs a full validation
* If you want to build graphs out of order you use `Graph`
* If you want checking while building your graph you can use `GraphBuilder`

But as I also mentioned, both can be had in one class with the caveat I mentioned above about `Nodes = string`.

But thinking more about it, I agree trying to solve these different use cases still makes more sense with one unified approach, so folding it into Graph makes perfect sense to me now.

nfcampos:
Superseded by #127

PR #114, opened 2024-04-09
fix: state graph new node validation

dimatill:
Fixes a new node key validation for StateGraph.

In the previous implementation, it validated existing nodes instead of new ones.

Python implementation reference: https://github.com/langchain-ai/langgraph/blob/main/langgraph/graph/state.py#L48


nfcampos (approved):

nfcampos:
Thanks a lot for this @dimatill

PR #118, opened 2024-04-17
Rename `empty()` to `fromCheckpoint()` to match LangGraph Python

andrewnguonly:
### Summary
All files in the `channels` directory are updated to match the corresponding naming and implementation in LangGraph Python.

### Implementation
1. Rename `empty()` to `fromCheckpoint()` in `BaseChannel` class.
2. Update implementation of `update()` methods to return last element of `values` array.

### To Do
- [x] Figure out if `BaseChannel.createCheckpoint()` is equivalent to the Python implementation.
- [x] Do we want to remove `initialValueFactory` API from `BaseChannel`? The API doesn't exist in the Python implementation. Static function `StateGraph._getChannels()` depends on this interface.

### Future
1. Remove `guard` API from `EphemeralValue`. It's not needed anymore.
2. Make `NamedBarrierValue.update() ` transactional (`this.seen` can be partially updated).

I'm keeping track of these items separately.


langgraph/src/channels/any_value.ts
```diff
@@ -31,7 +31,7 @@ export class AnyValue<Value> extends BaseChannel<Value, Value, Value> {
     }
 
     // eslint-disable-next-line prefer-destructuring
-    this.value = values[0];
+    this.value = values[values.length - 1];
```
andrewnguonly, line 34: Python implementation: https://github.com/langchain-ai/langgraph/blob/main/langgraph/channels/any_value.py#L55

langgraph/src/channels/ephemeral_value.ts
```diff
@@ -36,7 +36,7 @@ export class EphemeralValue<Value> extends BaseChannel<Value, Value, Value> {
     }
 
     // eslint-disable-next-line prefer-destructuring
-    this.value = values[0];
+    this.value = values[values.length - 1];
```
andrewnguonly, line 39: Python implementation: https://github.com/langchain-ai/langgraph/blob/main/langgraph/channels/ephemeral_value.py#L63

langgraph/src/channels/last_value.ts
```diff
@@ -30,7 +30,7 @@ export class LastValue<Value> extends BaseChannel<Value, Value, Value> {
     }
 
     // eslint-disable-next-line prefer-destructuring
-    this.value = values[0];
+    this.value = values[values.length - 1];
```
andrewnguonly, line 33: Python implementation: https://github.com/langchain-ai/langgraph/blob/main/langgraph/channels/last_value.py#L57

langgraph/src/channels/base.ts
```diff
@@ -18,9 +18,8 @@ export abstract class BaseChannel<
    * @param {CheckpointType | undefined} initialValue
    * @returns {this}
    */
-  abstract empty(
-    checkpoint?: CheckpointType,
-    initialValueFactory?: () => CheckpointType
```
andrewnguonly, line 23: `initialValueFactory` is removed from `BaseChannel`, but it remains in `BinaryOperatorAggregate`.

langgraph/src/channels/base.ts
```diff
@@ -85,9 +102,9 @@ export async function createCheckpoint<ValueType>(
   const newCheckpoint: Checkpoint = {
     v: 1,
     ts: new Date().toISOString(),
-    channelValues: { ...checkpoint.channelValues },
-    channelVersions: { ...checkpoint.channelVersions },
-    versionsSeen: { ...checkpoint.versionsSeen },
+    channelValues: deepCopy(checkpoint.channelValues),
```
andrewnguonly, line 105: `checkpoint.channelValues` and `checkpoint.channelVersions` don't need to be deep copied. They can be set to an empty object so previous channel values aren't carried over. This is equivalent to Python implementation.

nfcampos (approved):

PR #119, opened 2024-04-19
Implement `serde`, `getTuple()`, and `list()` in `BaseCheckpointSaver`

andrewnguonly:
### Summary
To get checkpointing functionality up to parity with LangGraph's Python implementation, the `BaseCheckpointSaver` class needs to be updated to define `serde`, `getTuple()` and `list()`. Downstream dependencies (e.g. `MemorySaver`) are updated accordingly.

### Implementation
1. Move `deepCopy()` to `/checkpoint/base.ts`.
2. Update `BaseCheckpointSaver`: Define `serde` attribute. Define `getTuple()` and `list()` methods. Update methods to return `Promise`.
3. Update `MemorySaver`: Update `storage` attribute type. Update method implementations.
4. Implement `SqliteSaver`.

### To Do
- [x] Implement `serde` in `MemorySaver`.
- [x] Implement SQLite checkpointer and tests.
- [x] Add unit tests for `getTuple()` and `list()`.
- [x] Clean up the code a bit.
- [x] Update SQLite schema to use `camel_case`.
- [ ] ~Add changes for `SqliteSaver` from this PR: https://github.com/langchain-ai/langgraph/pull/332~


langgraph/src/tests/checkpoints.test.ts
andrewnguonly, line 1: This file is named `checkpoints.test.ts` instead of `checkpoint.test.ts` because the `jest --testPathIgnorePatterns=\\.int\\.test.ts` config in `package.json` is ignoring the later file name for some reason. The regex is correct, but there seems to be a bug with `jest`.

I'm not sure what the best solution is. I tried updating the regex to require at least one character before the first dot (`.`), but I couldn't get `jest` to pick up any files.

langgraph/src/checkpoint/base.ts
```diff
@@ -65,14 +83,67 @@ export const enum CheckpointAt {
...
 }
 
+export interface CheckpointTuple {
+  config: RunnableConfig;
+  checkpoint: Checkpoint;
+  parentConfig?: RunnableConfig;
+}
+
+const CheckpointThreadId: ConfigurableFieldSpec = {
+  id: "threadId",
```
andrewnguonly, line 93: @nfcampos, general question: The JS implementation serializes `threadId` and `threadTs` in `camelCase` (as opposed to `snake_case` in Python). Is there any value in making them both one way or the other?

Hypothetical scenario: Python graph checkpoints state, pauses execution, and JS graph (or some other runtime) resumes execution.

nfcampos: yep would be nice to use same keys

andrewnguonly: I'll make the changes in this PR.

langgraph/src/checkpoint/base.ts
```diff
@@ -65,14 +83,67 @@ export const enum CheckpointAt {
...
+
+export interface SerializerProtocol {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  dumps(obj: any): any;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  loads(data: any): any;
+}
+
 export abstract class BaseCheckpointSaver {
   at: CheckpointAt = CheckpointAt.END_OF_RUN;
```
andrewnguonly, line 121: Change this to `CheckpointAt.END_OF_STEP`.

langgraph/src/checkpoint/base.ts
```diff
@@ -65,14 +83,67 @@ export const enum CheckpointAt {
...
+  description:
+    "Pass to fetch a past checkpoint. If None, fetches the latest checkpoint.",
+  default: null,
+  isShared: true,
+  dependencies: null,
+};
+
+export interface SerializerProtocol {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  dumps(obj: any): any;
```
andrewnguonly, line 115: Not sure what the right input/output types are for `dumps()` and `loads()`.

nfcampos: I think in JS the most common would be `any -> string -> any`

langgraph/src/checkpoint/sqlite.ts
```diff
@@ -0,0 +1,195 @@
...
+  db: DatabaseType;
+
+  constructor(
+    connStringOrLocalPath: string,
+    serde?: SerializerProtocol,
+    at?: CheckpointAt
+  ) {
+    super(serde, at);
+    this.db = new Database(connStringOrLocalPath);
+    this.setup().catch((error) =>
```
nfcampos, line 43: isn't there a race condition here? you could end up calling one of the methods before setup finishes

langgraph/src/checkpoint/sqlite.ts
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  static fromConnString(connStringOrLocalPath: string): SqliteSaver {
+    return new SqliteSaver(connStringOrLocalPath);
+  }
+
+  private async setup(): Promise<void> {
+    try {
+      this.db.exec(`
+CREATE TABLE IF NOT EXISTS checkpoints (
```
nfcampos, line 55: prob worth enabling wal mode? I know I haven't done that in py, prob should there too

langgraph/src/checkpoint/sqlite.ts
andrewnguonly: Yes, you're right. Let me fix.

I reproduced the race condition by adding a sleep statement.
```typescript
private async setup(): Promise<void> {
  // sleep for 3 seconds
  await new Promise((resolve) => setTimeout(resolve, 3000));
  ...
}
```
The unit tests failed.

langgraph/src/checkpoint/sqlite.ts
andrewnguonly: > I know I haven't done that in py, prob should there too

Added this to my To Do list for later.

langgraph/src/checkpoint/base.ts
andrewnguonly: Let me know if you like this approach better: https://github.com/langchain-ai/langgraphjs/pull/119/commits/348572c62397da2f41ad97680dc8d7cdce7cad63 (commit)

I implemented generics on the `BaseCheckpointSaver` class and the `SerializerProtocol` interface. This ensures that a type incompatible serializer cannot be passed to the checkpoint saver constructor. For example, the `SqliteSaver` requires the checkpoint to be serialized to a `string`. A serializer that doesn't implement this correctly will raise an error. It's also nice to help developers implement `dumps()` and `loads()` with the correct types.

That being said, I'm not sure how much value this provides.

langgraph/src/checkpoint/base.ts
```diff
@@ -65,14 +83,65 @@ export const enum CheckpointAt {
...
+  isShared: true,
+  dependencies: null,
+};
+
+export interface SerializerProtocol<D, L> {
+  dumps(obj: D): L;
+  loads(data: L): D;
+}
+
+export abstract class BaseCheckpointSaver<D, L> {
```
nfcampos, line 118: D here should always be Checkpoint

andrewnguonly: Makes sense. I removed `D` and hardcoded `Checkpoint` in the internal implementation of `BaseCheckpointSaver`.

nfcampos (approved):

PR #125, opened 2024-05-01
Update static functions in `pregel/index.ts` to match corresponding Python implementation

andrewnguonly:
### Summary
This PR updates static functions in `pregel/index.ts` to the match the corresponding Python implementation. These functions are used in the core `stream()` and `invoke()` methods of the `Pregel` class.

### Implementation
1. Implement `_shouldInterrupt()` (currently unused).
2. Update `_applyWrites()` to match Python implementation.
3. Remove `_applyWritesFromView()`. Removing the function did not break any tests. Not sure if this needs to be migrated.
4. Update `_prepareNextTasks()` to match Python implementation. Note: The side effect from this function is removed and the calling code is updated to ensure that all existing tests pass.
5. Implement `_localRead()` (replaced old implementation of `read`).
6. Update `ChannelRead` and `ChannelWrite` to extend from `RunnableCallable` (this was a regression).
7. Update `@langchain/core` and update `RunnableCallable.invoke()` to call `mergeConfigs()` (this was a regression).
8. Implement `streamChannelsList` getter.

### Next Steps
1. Update `_default()`, `stream()`, and `invoke()` in `Pregel` class.

### DIFF
--- a/langgraph/src/channels/base.ts
+++ b/langgraph/src/channels/base.ts
@@ -77,15 +77,15 @@ export function emptyChannels(
   return newChannels;
 }
 
-export async function createCheckpoint<ValueType>(
+export function createCheckpoint<ValueType>(
   checkpoint: Checkpoint,
   channels: Record<string, BaseChannel<ValueType>>
-): Promise<Checkpoint> {
+): Checkpoint {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const values: Record<string, any> = {};
   for (const k of Object.keys(channels)) {
     try {
-      values[k] = await channels[k].checkpoint();
+      values[k] = channels[k].checkpoint();
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
     } catch (error: any) {
       if (error.name === EmptyChannelError.name) {
--- a/langgraph/src/pregel/index.ts
+++ b/langgraph/src/pregel/index.ts
@@ -595,77 +595,85 @@ export function _prepareNextTasks(
 
     // If any of the channels read by this process were updated
     if (
-      proc.triggers.some(
-        (chan) => checkpoint.channelVersions[chan] > (seen[chan] ?? 0)
-      )
+      proc.triggers
+        .filter(
+          (chan) =>
+            readChannel(channels, chan, true, true) !== EmptyChannelError
+        )
+        .some((chan) => checkpoint.channelVersions[chan] > (seen[chan] ?? 0))
     ) {
-      // If all channels subscribed by this process have been initialized
-      try {
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        let val: Record<string, any> = {};
-        if (Array.isArray(proc.channels)) {
-          // eslint-disable-next-line no-unreachable-loop
-          for (const chan of proc.channels) {
-            val[chan] = readChannel(channels, chan);
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      let val: any;
+
+      // If all trigger channels subscribed by this process are not empty
+      // then invoke the process with the values of all non-empty channels
+      if (Array.isArray(proc.channels)) {
+        let emptyChannels = 0;
+        for (const chan of proc.channels) {
+          try {
+            val = readChannel(channels, chan, false);
             break;
-          }
-        } else {
-          for (const [k, chan] of Object.entries(proc.channels)) {
-            val[k] = readChannel(channels, chan);
-          }
-        }
-
-        // Processes that subscribe to a single keyless channel get
-        // the value directly, instead of a dict
-        if (Array.isArray(proc.channels)) {
-          // eslint-disable-next-line no-unreachable-loop
-          for (const chan of proc.channels) {
-            val = val[chan];
-            break;
-          }
-        } else if (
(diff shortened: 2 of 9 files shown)


langgraph/src/channels/base.ts
```diff
@@ -77,15 +77,15 @@ export function emptyChannels(
   return newChannels;
 }
 
-export async function createCheckpoint<ValueType>(
```
andrewnguonly, line 80: This function needs to be non-async in order for `_localRead()` to be patched in `RunnableConfig` and eventually called so it doesn't return a `Promise`.

Need to double check if this is ok.

langgraph/src/pregel/index.ts
```diff
@@ -326,16 +341,25 @@ export class Pregel
...
+      const [nextCheckpoint, nextTasks] = _prepareNextTasks(
+        checkpoint,
+        processes,
+        channels,
+        true
+      );
+
+      // Reassign nextCheckpoint to checkpoint because the subsequent implementation
+      // relies on side effects applied to checkpoint. Example: _applyWrites().
+      checkpoint = nextCheckpoint as Checkpoint;
```
andrewnguonly, line 361: This is temporary. When `stream()` is reimplemented, I think this code will be removed.

langgraph/src/pregel/index.ts
```diff
@@ -347,21 +371,30 @@ export class Pregel
 
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       const pendingWrites: Array<[string, any]> = [];
+      // A copy of the checkpoint is created because `checkpoint` is defined with `let`.
+      // `checkpoint` can be mutated during loop execution and when used in a function,
+      // may cause unintended consequences.
+      const checkpointCopy = copyCheckpoint(checkpoint);
```
andrewnguonly, line 377: This might be temporary. When `stream()` is reimplemented, this code might be removed.

langgraph/src/pregel/index.ts
```diff
@@ -384,10 +417,6 @@ export class Pregel
 
       if (stepOutput) {
         yield stepOutput;
-
-        if (typeof outputKeys !== "string") {
-          _applyWritesFromView(checkpoint, channels, stepOutput);
```
andrewnguonly, line 389: Removing this had no effect (i.e. all tests passed). Need to double check if I need to migrate it.

langgraph/src/pregel/index.ts
```diff
@@ -484,7 +513,58 @@ async function executeTasks<RunOutput>(
...
+  writes: Array<[string, any]>,
+  select: Array<string> | string,
+  fresh: boolean = false
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+): Record<string, any> | any {
+  if (fresh) {
+    const newCheckpoint = createCheckpoint(checkpoint, channels);
+
+    // create a new copy of channels
+    const newChannels = Object.entries(channels).reduce(
```
andrewnguonly, line 545: Double check: This implementation is ported from a Python implementation that uses a context manager. What's the equivalent/preferred implementation in JS/TS?

langgraph/src/pregel/read.ts
```diff
@@ -232,7 +230,10 @@ export class PregelNode<
         triggers: this.triggers,
         mapper: this.mapper,
         writers: [...this.writers, coerceable as ChannelWrite],
-        bound: this.bound.pipe(coerceable),
```
andrewnguonly, line 235: This was a regression from a previous PR. It was only possible to do this once `_prepareNextTasks()` was updated.

langgraph/src/utils.ts
```diff
@@ -1,4 +1,8 @@
-import { Runnable, RunnableConfig } from "@langchain/core/runnables";
+import {
+  mergeConfigs,
```
andrewnguonly, line 2: `mergeConfigs()` was exported in `@langchain/core` `0.1.61`.

langgraph/src/channels/base.ts
nfcampos: yep I think this doesn't have to be async

langgraph/src/utils.ts
```diff
@@ -48,9 +52,13 @@ export class RunnableCallable extends Runnable {
 
     // TODO: mergeConfigs() from @langchain/core is not exported
```
nfcampos, line 53: remove comment

package.json
```diff
@@ -31,7 +31,7 @@
   "author": "LangChain",
   "license": "MIT",
   "resolutions": {
-    "@langchain/core": "0.1.51"
```
nfcampos, line 34: why do we need this in both package json files?

langgraph/src/pregel/index.ts
nfcampos: ah this is an api that used to exist in py but we removed, so good to remove here too

package.json
andrewnguonly: Not 100% sure, but my assumption is because the `@langchain/core` dependencies are also needed in the `examples` directory, which is at the same level as `langgraph`.

langgraph/src/pregel/index.ts
nfcampos: this is fine

nfcampos (approved):

PR #142, opened 2024-05-15
Rename thread_ts to checkpoint_id

nfcampos:
- Also rework any objects that will be saved in db / server over api to use snake_case keys (to match py)

Fixes # (issue)

### DIFF
--- a/langgraph/src/checkpoint/sqlite.ts
+++ b/langgraph/src/checkpoint/sqlite.ts
@@ -169,7 +169,7 @@ CREATE TABLE IF NOT EXISTS checkpoints (
     try {
       const row = [
         config.configurable?.thread_id,
-        checkpoint.ts,
+        checkpoint.id,
         config.configurable?.checkpoint_id,
         this.serde.stringify(checkpoint),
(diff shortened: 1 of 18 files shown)


langgraph/src/checkpoint/sqlite.ts
```diff
@@ -122,31 +122,31 @@ CREATE TABLE IF NOT EXISTS checkpoints (
...
             },
             checkpoint: this.serde.parse(row.checkpoint) as Checkpoint,
             metadata: this.serde.parse(row.metadata) as CheckpointMetadata,
-            parentConfig: row.parent_ts
+            parentConfig: row.parent_id
               ? {
                   configurable: {
-                    threadId: row.thread_id,
-                    threadTs: row.parent_ts,
+                    thread_id: row.thread_id,
```
hinthornw, line 148: so we're officially dropping support for pascal case then?

nfcampos: In things that are meant to be sent over the wire / could one day be shared between js and py yes

PR #169, opened 2024-05-18
Implement createReActAgent

hinthornw:


langgraph/package.json
```diff
@@ -43,7 +43,7 @@
   "devDependencies": {
     "@jest/globals": "^29.5.0",
     "@langchain/community": "^0.0.43",
-    "@langchain/openai": "^0.0.23",
+    "@langchain/openai": "latest",
```
hinthornw, line 46: to get the bindTools for testing

langgraph/src/prebuilt/react_agent_executor.ts
```diff
@@ -0,0 +1,187 @@
...
+    ) {
+      return END;
+    } else {
+      return "continue";
+    }
+  };
+
+  const callModel = async (state: AgentState) => {
+    const { messages } = state;
+    // TODO: Stream
```
hinthornw, line 102: If we wanted to support streamEvents, i think i need to actually stream here and concatenate, though i need to check. been a while since ir eviewed the js chat model implementations

langgraph/src/prebuilt/react_agent_executor.ts
```diff
@@ -0,0 +1,187 @@
...
+import { MessagesState } from "../graph/message.js";
+import { CompiledStateGraph, StateGraphArgs } from "../graph/state.js";
+import { All } from "../pregel/types.js";
+import { ToolNode } from "./tool_node.js";
+
+export interface AgentState {
+  messages: BaseMessage[];
+  // TODO: This won't be set until we
+  // implement managed values in LangGraphJS
+  is_last_step: boolean;
```
nfcampos, line 32: Then let's comment it out?

langgraph/src/prebuilt/react_agent_executor.ts
nfcampos: Because in js we don't auto promote invoke to stream like we do in py?

hinthornw: afaict

PR #174, opened 2024-05-20
Raise error if cond edge returns null/missing destination

nfcampos:
Fixes # (issue)


langgraph/src/prebuilt/react_agent_executor.ts
```diff
@@ -111,7 +111,7 @@ export function createReactAgent(
     .addEdge(START, "agent")
     .addConditionalEdges("agent", shouldContinue, {
       continue: "tools",
-      end: END,
+      [END]: END,
```
hinthornw, line 114: ah thanks!

hinthornw (approved):

PR #188, opened 2024-05-30
Update token streaming example for new @langchain/core

jacoblee93:
@hinthornw @nfcampos


hinthornw (approved):

jacoblee93:
Oh, @nfcampos suggested getting rid of the comments. Will open another PR.

PR #191, opened 2024-05-30
[Docs] Run TS compiloer on HOWTOs + README

hinthornw:


_scripts/run_notebook.py
```diff
@@ -0,0 +1,85 @@
...
+                options.append(str(value))
+    return options
+
+
+def compile_ts_code(ts_code: str, tsconfig_options: list) -> str:
+    print("Compiling TypeScript code...")
+    temp_file_path = "temp.ts"
+    with open(temp_file_path, "w") as f:
+        f.write(ts_code)
+    cmd = ["npx", "tsc", temp_file_path] + tsconfig_options
```
nfcampos, line 43: Can't you just pass the config file to tsc?

_scripts/run_notebook.py
```diff
@@ -0,0 +1,85 @@
...
+    with open(compiled_js_path, "r", encoding="utf-8") as js_file:
+        compiled_js_code = js_file.read()
+
+    return compiled_js_code, compiled_js_path
+
+
+def run_js_code(js_code_path: str) -> str:
+    print("Running JavaScript code...")
+    process = subprocess.Popen(
+        ["npx", "node", js_code_path],
```
nfcampos, line 62: Why use npx everywhere?

hinthornw: Because i have no clue what i'm doing lol

_scripts/run_notebook.py
hinthornw: You would think. Maybe you can. Doesn't seem like it though
1. https://www.typescriptlang.org/docs/handbook/compiler-options.html
2. https://arc.net/l/quote/lvbifppw

>  When input files are specified on the command line, tsconfig.json files are ignored.

PR #193, opened 2024-05-30
[Docs] Compile more tutorials

hinthornw:
Couple of the RAG ones. Fixup, check types, etc.


_scripts/run_notebook.py
```diff
@@ -40,7 +40,7 @@ def compile_ts_code(ts_code: str, tsconfig_options: list) -> str:
     temp_file_path = "temp.ts"
     with open(temp_file_path, "w") as f:
         f.write(ts_code)
-    cmd = ["npx", "tsc", temp_file_path] + tsconfig_options
+    cmd = ["npx", "tsc", temp_file_path] + tsconfig_options + ["--skipLibCheck"]
```
hinthornw, line 43: This is so it stops raising errors when i import cheerio ; maybe something to do with esm module vs nodenext vs idc

Issue #81, opened 2024-03-07
Async_hooks cannot be used in browser


Hehongyuanlove:
I'm using [chat_agent_executor_with_function_calling](https://github.com/langchain-ai/langgraphjs/blob/main/examples/chat_agent_executor_with_function_calling/base.ipynb)

The same code deno can run, but react cannot run in the browser

### node -v  
v21.7.0

### expo 
C:\xxxxx\node_modules\@langchain\langgraph\dist\setup\async_local_storage.js:5 Uncaught TypeError: _async_hooks.AsyncLocalStorage is not a constructor
    at initializeAsyncLocalStorageSingleton (C:\xxxxx\node_modules\@langchain\langgraph\dist\setup\async_local_storage.js:5:2)
    at new Pregel (C:\xxxxxt\node_modules\@langchain\langgraph\dist\pregel\index.js:159:51)
    at StateGraph.compile (C:\xxxxx\node_modules\@langchain\langgraph\dist\graph\state.js:90:13)
    at C:\xxxxx\page\tool\index.tsx:164:21
    at loadModuleImplementation (C:\xxxxx\node_modules\metro-runtime\src\polyfills\require.js:282:19)
    at guardedLoadModule (C:\xxxxx\node_modules\metro-runtime\src\polyfills\require.js:182:1)
    at metroRequire (C:\xxxxx\node_modules\metro-runtime\src\polyfills\require.js:91:43)
    at C:\xxxxx\App.js:4:44
    at loadModuleImplementation (C:\xxxxx\node_modules\metro-runtime\src\polyfills\require.js:282:19)
    at guardedLoadModule (C:\xxxxx\node_modules\metro-runtime\src\polyfills\require.js:182:1)

bracesproul:
cc @jacoblee93 related to `AsyncLocalStorage`

jacoblee93:
Oh shoot 😬

demchuk-alex:
Hey @jacoblee93 having the same issue, tring to use langgraphjs in react-native app, and nextjs.
Sorry for nudging is there any news or work arounds ?

jacoblee93:
Thank you for nudging - will try my best to look this week!

davidfant:
@jacoblee93 I started getting this issue after upgrading Langchain core. Before I was using langgraph 0.0.10 successfully with langchain core 0.1.47, but since upgrading to 0.1.51 this started happening

demchuk-alex:
[...]

PR #211, opened 2024-06-12
Adds web entrypoint

### LINKED ISSUE #208: "AsyncLocalStorage" is not exported by "__vite-browser-external"
Trying to build the langgraph project using vite, getting this error

"AsyncLocalStorage" is not exported by "__vite-browser-external", imported by "node_modules/@langchain/langgraph/dist/setup/async_local_storage.js".

jacoblee93:
Creates separate entrypoint for LangGraph classes that does not include async local storage.

I'm not sure why the initialize call was so nested - I think may have been unnecessary, tracing seems fine.

We will want to document this carefully and also update prebuilt components to explicitly pass config around to avoid the need for a `prebuilt/web` entrypoint.

CC @nfcampos @andrewnguonly @dqbd

Fixes #208 and #81


andrewnguonly:
What's the impact of this? Does this means LangGraph classes can be used in a browser context?

jacoblee93:
> What's the impact of this? Does this means LangGraph classes can be used in a browser context?

Exactly, that's the big one that doesn't support `async_hooks`.

For prebuilt components, we should make sure we pass `config` manually and avoid importing from the primary `index.js` entrypoint from now on.

nfcampos (approved):

adamgit:
Fresh install of langsmith, following the API examples word for word, crashes apparently because of the same issue this was created to resolve.

No langgraph is being imported directly - it's all being brought-in by langsmith. But there's no mention in the langsmith docs that I can see for how to workaround this.

Was this workaround only created for direct use of langgraph? i.e. langsmith etc needs to be manually patched to make it work in browser?

jacoblee93:
Depends on what you're importing but if it's `traceable` then it will not work in the browser due to async hooks usage.

We've been careful to isolate it for LangGraph/LangChain and could do the same for LangSmith but figured it was more niche since in most cases you wouldn't want to leak your API key on the frontend - what is your use case?

adamgit:
Personally: local development / VPN, and evaluation. If it's known / 
deliberate that it can't work in browser then it would be good if someone 
documented that - ideally in the features pages but at least: in the 
install docs.

Scra3:
Hi, I'm using the web entry point, and I'm wondering how to use the “interrupt” function as it's not exported from “web”? Thanks 🙏

jacoblee93:
It's currently unsupported in `web` as it requires AsyncLocalStorage features :(

Scra3:
Thank you for the answer, do you know if there is an alternative?

jacoblee93:
You can use a `NodeInterrupt` and then update state before resuming:

https://langchain-ai.github.io/langgraphjs/how-tos/#human-in-the-loop

https://langchain-ai.github.io/langgraphjs/how-tos/dynamic_breakpoints/

CC @benjamincburns

PR #235, opened 2024-06-27
Fix - Updated uuid pkg to support v6 generation

### LINKED ISSUE #234: Bug in checkpoint id generation when converting  UUID v1 to UUID v6.
When the graph is generating a UUID v6 from the starting point the generated id looks like : `6fbb4270-3461-11ef-undefineda0-a9b17340a0aa`. 

Note the 'undefined' in the UUID generated.

This happens because the code used to generate the `id` is called with a negative number. (-1) when the graph does not have previous checkpoints.  

UUID v1 to generate the ID here:  https://github.com/langchain-ai/langgraphjs/blob/main/langgraph/src/checkpoint/id.ts#L30

After the v1 generation, there is a conversion to a UUID v6. It be easy to solve if the use uuid v6, from start. A proposed code bellow. 

```
function uuid6(clockseq: number) {
  // using uuid pkg
  const node = crypto.getRandomValues(new Uint8Array(6));

  return  v6({ node, clockseq });
}
```

@nfcampos ?

chentschel:
Fixes #234

### DIFF
--- a/langgraph/package.json
+++ b/langgraph/package.json
@@ -50,7 +50,7 @@
     "@swc/jest": "^0.2.29",
     "@tsconfig/recommended": "^1.0.3",
     "@types/better-sqlite3": "^7.6.9",
-    "@types/uuid": "^9",
+    "@types/uuid": "^10",
     "@typescript-eslint/eslint-plugin": "^6.12.0",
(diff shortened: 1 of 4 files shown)


nfcampos (dismissed):
Thanks!

nfcampos (approved):

PR #236, opened 2024-06-28
feat: vercelKV checkpointer (WIP)

chentschel:
This PR adds VercelKV checkpointer support. 

Working OK tested against vercelKV. It needs proper mockup for the vercelKV service before merging. @nfcampos

### DIFF
--- a/langgraph/package.json
+++ b/langgraph/package.json
@@ -53,6 +53,7 @@
     "@types/uuid": "^10",
     "@typescript-eslint/eslint-plugin": "^6.12.0",
     "@typescript-eslint/parser": "^6.12.0",
+    "@vercel/kv": "^2.0.0",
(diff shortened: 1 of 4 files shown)


jacoblee93:
Ooh very cool!

jacoblee93:
Could you open this/put this in `@langchain/community` instead here though?

https://github.com/langchain-ai/langchainjs

chentschel:
@jacoblee93 Sure I can. 

I think it be good  to export `SerializerProtocol` and `CheckpointTuple` from this pkg to allow these contributions there. Is it OK?

jacoblee93:
Yeah makes sense. We'll also need to add an optional dep on LangGraph to community, let me know if I can help with that.

PR #237, opened 2024-07-01
fix: export checkpointer interfaces

chentschel:
This PR exports `CheckpointTuple` and `SerializerProtocol` interfaces to allow checkpointer contributions from 3rd party pkgs. @jacoblee93


jacoblee93:
Thank you! I don't think we needed to export from `checkpoints/index`, removed that for now.

jacoblee93 (approved):

PR #241, opened 2024-07-04
Adds Mermaid graph support

jacoblee93:
CC @nfcampos


langgraph/package.json
```diff
@@ -37,8 +37,9 @@
   "author": "LangChain",
   "license": "MIT",
   "dependencies": {
-    "@langchain/core": ">0.1.61 <0.3.0",
-    "uuid": "^10.0.0"
+    "@langchain/core": ">=0.2.14 <0.3.0",
+    "uuid": "^10.0.0",
+    "zod": "^3.23.8"
```
nfcampos, line 42: Why do we need an extra dependency now?

langgraph/src/utils.ts
```diff
@@ -89,3 +94,40 @@ export class RunnableCallable<I = unknown, O = unknown> extends Runnable<I, O> {
     return returnValue;
   }
 }
+
+export class DrawableGraph extends Graph {
+  override extend(graph: Graph, prefix: string = "") {
```
nfcampos, line 99: I think this might not be necessary anymore (in python too), isn't this the same as the implementation in core now?

langgraph/package.json
jacoblee93: Very lightly used here to supply a schema - but it's a dependency of `@langchain/core` anyway so figured it wasn't a big deal

langgraph/src/utils.ts
jacoblee93: You are correct, thought it was doing some different prefixing but no.

jacoblee93:
Screenshot 2024-07-09 at 9 13 43 AM

PR #266, opened 2024-07-17
Adds automatic error handling for ToolNode, RemoveMessage support

jacoblee93:
This PR also changes the params of the `ToolNode` constructor, but the other params were undocumented and unlikely to be used. I think worth fixing now.


jacoblee93:
CC @vbarda

langgraph/src/prebuilt/tool_node.ts
```diff
@@ -22,13 +28,16 @@ export class ToolNode<
 
   tools: (StructuredToolInterface | RunnableToolLike)[];
 
+  handleToolErrors = true;
+
   constructor(
     tools: (StructuredToolInterface | RunnableToolLike)[],
-    name: string = "tools",
-    tags: string[] = []
+    options?: ToolNodeOptions
```
bracesproul, line 35: Breaking change?

jacoblee93: Yeah see my comment above

PR #280, opened 2024-07-23
tslab instructions - Update README.md

justinlevi:
I was very familiar with Jupyter notebooks but hadn't used tslab before. Took me a minute to figure out how to run the examples. This may not be the best approach, but got me up and running and thought it might help someone else.

Fixes # (issue)


jacoblee93:
Thanks for this! `yarn build` makes sense but I'm surprised you needed to add that `tsconfig.json` file 🤔

PR #286, opened 2024-07-26
docs[minor]: Port concepts pages from PY docs

bracesproul:


docs/docs/concepts/low_level.md
```diff
@@ -0,0 +1,427 @@
...
+          reducer: messagesStateReducer,
+          default: () => [],
+        },
+      },
+    });
+  }
+}
+```
+
+What this is doing is creating a `State` with a single key `messages`. This is a list of `BaseMessage`s, with [`messagesStateReducer`](https://langchain-ai.github.io/langgraphjs/reference/functions/index.messagesStateReducer.html) as a reducer. `messagesStateReducer` basically adds messages to the existing list (it also does some nice extra things, like convert from OpenAI message format to the standard LangChain message format, handle updates based on message IDs, etc).
```
austinm911, line 122: Sorry this is bit confusing. 

Isn't BaseMessage imported from '@langchain/core/messages'?

It also reads like `MessageState` is supposed to imported and used? But it's not imported in the example.

The docs use the pattern below all over the place but this seems like it introduces a new implementation which is confusing.

I'm not sure if MessageState is WIP somewhere and I'm missing it. 

```ts
// this pattern is used throughout the docs
// Define the state interface
interface AgentState {
  messages: BaseMessage[];
}

// Define the graph state
const graphState: StateGraphArgs<AgentState>["channels"] = {
  messages: {
    reducer: (x: BaseMessage[], y: BaseMessage[]) => x.concat(y),
  },
};
```

How is the graph constructed after this?
```typescript
interface State extends MessagesState {
  documents: Array<string>
}
```

Again it seems like the expected pattern would be?
```ts
const graphState: StateGraphArgs<AgentState>["channels"] = {
  messages: {
    reducer: (x: BaseMessage[], y: BaseMessage[]) => x.concat(y),
  },
  documents: {
     reducer: `implementation here`
  },
};
```

PR #289, opened 2024-07-31
Define StateGraph State in a single delcaration, from which both reducers and state and update interfaces are inferred

nfcampos:
This more closely matches python, where a single declaration gives you all three

Fixes # (issue)

### DIFF
--- a/langgraph/src/graph/state.ts
+++ b/langgraph/src/graph/state.ts
@@ -430,15 +430,16 @@ export class CompiledStateGraph<
   }
 }
 
+function isBaseChannel(obj: unknown): obj is BaseChannel {
+  return obj != null && typeof (obj as BaseChannel).lc_graph_name === "string";
+}
+
 function isStateDefinition(obj: unknown): obj is StateDefinition {
   return (
     typeof obj === "object" &&
     obj !== null &&
     !Array.isArray(obj) &&
     Object.keys(obj).length > 0 &&
-    Object.values(obj).every(
-      // eslint-disable-next-line no-instanceof/no-instanceof
-      (v) => typeof v === "function" || v instanceof BaseChannel
-    )
+    Object.values(obj).every((v) => typeof v === "function" || isBaseChannel(v))
   );
 }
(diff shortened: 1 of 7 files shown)


jacoblee93:
Game changing thank you!

langgraph/src/graph/state.ts
```diff
@@ -53,19 +94,34 @@ export interface StateGraphArgs<Channels extends object | unknown> {
 }
 
 export class StateGraph<
-  State extends object | unknown,
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  Update extends object | unknown = Partial<Record<keyof State, any>>,
+  SD extends StateDefinition | unknown,
```
jacoblee93, line 97: I generally don't like abbreviations but it's fine

langgraph/src/graph/state.ts
```diff
@@ -374,3 +429,16 @@ export class CompiledStateGraph<
     }
   }
 }
+
+function isStateDefinition(obj: unknown): obj is StateDefinition {
```
jacoblee93, line 433: May be nice to export this?

langgraph/src/graph/state.ts
```diff
@@ -374,3 +429,16 @@ export class CompiledStateGraph<
...
+  return (
+    typeof obj === "object" &&
+    obj !== null &&
+    !Array.isArray(obj) &&
+    Object.keys(obj).length > 0 &&
+    Object.values(obj).every(
+      // eslint-disable-next-line no-instanceof/no-instanceof
+      (v) => typeof v === "function" || v instanceof BaseChannel
```
jacoblee93, line 441: nit: Use presence/absence of `fromCheckpoint` or `lc_graph_name` instead to avoid bundler issues?

jacoblee93 (dismissed):

langgraph/src/graph/state.ts
nfcampos: I did this to avoid this being same name as the name the user would likely use in their code, which I think could be a little confusing

jacoblee93:
Will merge and follow up with docs

PR #294, opened 2024-08-05
Use config.signal to control cancellation

nfcampos:
Fixes # (issue)


jacoblee93 (approved):
Thank you!

Pckool:
Unfortunately this doesn't work properly; when passing a signal to it, the chain doesn't get past the __start__ node #319

PR #304, opened 2024-08-09
Spelling

hinthornw:


jacoblee93:
Weird failure - I think old version of core is sneaking in 😢

hinthornw:
fun ha

jacoblee93:
Need to fix this core issue ASAP

PR #323, opened 2024-08-16
feat(debug): add checkpoint debug events

dqbd:
Fixes # (issue)

### DIFF
--- a/langgraph/src/pregel/algo.ts
+++ b/langgraph/src/pregel/algo.ts
@@ -427,7 +427,7 @@ export function _prepareNextTasks<
   // Check if any processes should be run in next step
   // If so, prepare the values to be passed to them
   for (const [name, proc] of Object.entries<PregelNode>(processes)) {
-    const hasUpdatedChannels = proc.triggers
+    const updatedChannels = proc.triggers
       .filter((chan) => {
         try {
           readChannel(channels, chan, false);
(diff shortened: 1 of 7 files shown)


langgraph/src/pregel/algo.ts
```diff
@@ -417,17 +418,14 @@ export function _prepareNextTasks<
         });
       }
     } else {
-      taskDescriptions.push({
-        name: packet.node,
-        input: packet.args,
-      });
+      taskDescriptions.push({ id: taskId, name: packet.node });
```
jacoblee93, line 421: No need for `input` here? Was a mistake before?

langgraph/src/pregel/algo.ts
jacoblee93, line 1: This all looks like just pure refactoring right?

langgraph/src/pregel/debug.ts
```diff
@@ -128,13 +142,51 @@ export function* mapDebugTaskResults<
...
   channels: Record<string, BaseChannel>,
   streamChannels: string | string[],
-  metadata: CheckpointMetadata
+  metadata: CheckpointMetadata,
+  tasks: readonly PregelExecutableTask<N, C>[],
+  pendingWrites: CheckpointPendingWrite[]
 ) {
+  function formatConfig(config: RunnableConfig) {
+    // https://stackoverflow.com/a/78298178
+    type CamelToSnake<
```
jacoblee93, line 159: Nice

jacoblee93 (approved):

langgraph/src/pregel/algo.ts
dqbd: Haven't seen usage of `input` in both JS nor Python, thus removed it in favour of matching Python

langgraph/src/pregel/algo.ts
dqbd: The goal was to add the `tasks` prop in debug checkpoint event, which required us to update the `PregelTaskDescription` with the `id`

PR #324, opened 2024-08-16
fix(algo): only return channels which actually triggered the task

dqbd:
Fixes # (issue)


langgraph/src/pregel/algo.ts
```diff
@@ -421,7 +423,7 @@ export function _prepareNextTasks<
           const metadata = {
             langgraph_step: extra.step,
             langgraph_node: name,
-            langgraph_triggers: proc.triggers,
+            langgraph_triggers: updatedChannels,
```
nfcampos, line 426: ha good fix

PR #333, opened 2024-08-20
langgraph[minor]: Expose MessagesState annotation

bracesproul:
cc @jacoblee93 for thoughts


langgraph/src/graph/message.ts
```diff
@@ -76,6 +77,11 @@ export class MessageGraph extends StateGraph<
   }
 }
 
-export interface MessagesState {
-  messages: BaseMessage[];
-}
+export const createMessagesState = () => Annotation.Root({
```
jacoblee93: Why don't we just expose the created state directly?

bracesproul: side effects, causes tree shaking to fail

jacoblee93: I think that's ok in this case?

bracesproul: I'm not aware of a way to suppress the warnings though, if you're okay with having them then I can update

jacoblee93: Look at https://github.com/langchain-ai/langgraphjs/blob/main/langgraph/src/index.ts

I would put this in its own file so that we're not ignoring tree shaking warnings for everything in this file

bracesproul: Okay, updated it to only be accessible via `@langchain/langgraph/messages`

jacoblee93: No you can still export it from the main entrypoint - just put it in its own file so that you don't have to annotate the other one.

bracesproul: There isn't a way to export it from `/web` then without adding the `__LC_ALLOW_ENTRYPOINT_SIDE_EFFECTS__` comment to that file too (not possible unless your code in `@langchain/scripts` which checks for that comment is updated, as it only checks if the entrypoint file includes that comment)

PR #386, opened 2024-08-27
langgraph[patch]: Adds inputDefinition and outputDefinition properties to compiled graph

jacoblee93:
This is a stopgap until we figure out a good way to implement schema support in core - unclear if we should use Zod or if there's even a way to generate Zod schema from a `StateDefinition`.

Maybe the best long term thing is to even just use a type.

Also does a bit of housecleaning (removes Pregel as an input to the loop to match Python and remove the implicit circular dep)


libs/langgraph/src/pregel/loop.ts
```diff
@@ -287,24 +297,19 @@ export class PregelLoop {
...
-        ? mapOutputUpdates(outputKeys, this.tasks).next().value
-        : mapOutputValues(outputKeys, writes, this.channels).next().value;
       await this._putCheckpoint({
         source: "loop",
-        writes: metadataWrites ?? null,
+        writes:
+          mapOutputUpdates(this.outputKeys, this.tasks).next().value ?? null,
       });
-      // after execution, check if we should interrupt
+      // after execution, check if we should inte`rrupt
```
bracesproul, line 312:
```suggestion
      // after execution, check if we should interrupt
```
nit

libs/langgraph/src/graph/state.ts
```diff
@@ -180,6 +187,28 @@ export class StateGraph<
...
+    for (const [key, val] of Object.entries(stateDefinition)) {
+      let channel;
+      if (typeof val === "function") {
+        channel = val();
+      } else {
+        channel = val;
+      }
+      if (this.channels[key] !== undefined) {
+        if (this.channels[key].lc_graph_name !== channel.lc_graph_name) {
+          if (channel.lc_graph_name !== "LastValue") {
```
nfcampos, line 200: I don't think we have this error like this in py?

jacoblee93: Fixed

PR #387, opened 2024-08-27
fix(prebuilt): use messagesStateReducer to support more formats

dqbd:
Fixes # (issue)


libs/langgraph/src/prebuilt/chat_agent_executor.ts
```diff
@@ -124,7 +124,7 @@ export function createFunctionCallingExecutor<Model extends object>({
   // So we annotate the messages attribute with operator.add
   const schema: StateGraphArgs<FunctionCallingExecutorState>["channels"] = {
     messages: {
-      value: (x: BaseMessage[], y: BaseMessage[]) => x.concat(y),
```
nfcampos, line 127: isn't this one deprecated?

PR #403, opened 2024-08-28
doc links

isahers1:


.markdown-link-check.config.json
```diff
@@ -0,0 +1,4 @@
+{
+  "aliveStatusCodes": [200, 206, 402],
```
bracesproul, line 2: Is it possible to do ranges? We prob want all 200's, all 300's, etc

.github/workflows/link_check.yml
```diff
@@ -0,0 +1,85 @@
...
+        with:
+          fetch-depth: 1
+
+      - name: Check README.md is in sync
+        run: |
+            if ! diff -q README.md libs/langgraph/README.md >/dev/null; then
+            echo "README.md is out of sync with libs/langgraph/README.md"
+            diff -C 3 README.md libs/langgraph/README.md
+            exit 1
+            fi
```
bracesproul, line 85: This already exists [here](https://github.com/langchain-ai/langgraphjs/blob/e298739291958be27cd50d8ffa4d0d5c9ca4bd28/.github/workflows/ci.yml#L105)

.github/workflows/link_check.yml
```diff
@@ -0,0 +1,85 @@
...
+on:
+  pull_request:
+    branches:
+      - main
+  push:
+    branches:
+      - main
+  schedule:
+    - cron: "0 5 * * *"
+  workflow_dispatch:
```
bracesproul, line 12: From the code below, it looks like you only want `.ipynb` files, in that case do this
```suggestion
on:
  push:
    branches: ["main"]
  pull_request:
    paths:
      - '**.ipynb'
  workflow_dispatch:
```
and no need for scheduling, just have it run in PRs and on commits to main

bracesproul: Ah, unless this is because we want to have a recurring check. Prob makes sense. In that case ya keep the scheduling.

.github/workflows/link_check.yml
```diff
@@ -0,0 +1,85 @@
...
+        run: |
+          yarn install --frozen-lockfile
+
+      - name: Check links in notebooks
+        env:
+          LANGCHAIN_API_KEY: test
+        run: |
+          if [ "${{ github.event_name }}" == "schedule" ] || [ "${{ github.event_name }}" == "workflow_dispatch" ] || ([ "${{ github.event_name }}" == "push" ] && [ "${{ github.ref }}" == "refs/heads/main" ]); then
+            echo "Running link check on all notebooks in examples directory..."
+            yarn run pytest -v --check-links-ignore "https://(api|web)\.smith\.langchain\.com/.*" --check-links-ignore "https://x.com/.*" --check-links examples
```
bracesproul, line 56: Not sure where the `pytest` yarn script is coming from. Maybe you forgot to commit a new dependency you added?

.github/workflows/link_check.yml
```diff
@@ -0,0 +1,85 @@
...
+      
+      - name: Install dependencies
+        run: |
+          yarn install --frozen-lockfile
+
+      - name: Check links in notebooks
+        env:
+          LANGCHAIN_API_KEY: test
+        run: |
+          if [ "${{ github.event_name }}" == "schedule" ] || [ "${{ github.event_name }}" == "workflow_dispatch" ] || ([ "${{ github.event_name }}" == "push" ] && [ "${{ github.ref }}" == "refs/heads/main" ]); then
```
bracesproul, line 54: prob dont need this since we have the condition at the top which will only run this on a pr push, push to main, workflow dispatch, or scheduling

.github/workflows/link_check.yml
bracesproul: also, is this `--check-links-ignore "https://(api|web)\.smith\.langchain\.com/.*"` saying don't check langsmith links? If so, why?

.markdown-link-check.config.json
isahers1: Based on my brief chat-gpt research it's not easily doable programatically. I just copied this from the main langgraph one so I assume someone smarter than me had a good reason for selecting these codes but idk

.github/workflows/link_check.yml
isahers1: langsmith links stuff was copied over, not sure why but assume there was a reason. tried to fix the pytest stuff in next push

benjamincburns:
@isahers1 this one is quite old now - still worth pursuing or should we close?

PR #427, opened 2024-08-30
langgraph[patch]: Adds node-level input schema to support private state

jacoblee93:


libs/langgraph/src/graph/state.ts
```diff
@@ -417,7 +440,7 @@ export class CompiledStateGraph<
   attachNode(key: N, node: StateGraphNodeSpec<S, U>): void;
 
   attachNode(key: N | typeof START, node?: StateGraphNodeSpec<S, U>): void {
-    const stateKeys = Object.keys(this.builder._inputDefinition);
+    const stateKeys = Object.keys(this.builder.channels);
```
jacoblee93, line 443: A little bit suspicious of doing this but I think it's ok since we're filtering later down the line by `inputDefinition` keys

@nfcampos

PR #445, opened 2024-09-03
fix(checkpointers): Avoid storing Sends twice in memory checkpointer

jacoblee93:
Will probably want to do the same for the other implementations


libs/checkpoint/src/serde/jsonplus.ts
```diff
@@ -140,6 +140,12 @@ function _default(_key: string, obj: any): any {
     return _encodeConstructorArgs(RegExp, undefined, [obj.source, obj.flags]);
   } else if (obj instanceof Error) {
     return _encodeConstructorArgs(obj.constructor, undefined, [obj.message]);
+    // TODO: Remove special case
+  } else if (obj?.lg_name === "Send") {
```
nfcampos, line 144: how was this not handled already?

jacoblee93: I am actually not sure, I suppose the default JSON.stringify did something good. It serialized with `lg_name` though so this just removes that to preserve immutability.

PR #458, opened 2024-09-04
langgraph[minor]: Log warning if NodeInterrupt is thrown in conditional edge

bracesproul:


libs/langgraph/src/graph/graph.ts
```diff
@@ -65,8 +65,20 @@ export class Branch<IO, N extends string> {
   ) {
     return ChannelWrite.registerWriter(
       new RunnableCallable({
-        func: (input: IO, config: RunnableConfig) =>
-          this._route(input, config, writer, reader),
+        func: async (input: IO, config: RunnableConfig) => {
+          try {
+            return await this._route(input, config, writer, reader);
```
bracesproul, line 70: Must await before returning or the err won't be caught

libs/langgraph/src/graph/graph.ts
```diff
@@ -65,8 +65,20 @@ export class Branch<IO, N extends string> {
...
+          } catch (e: any) {
+            // Detect & warn if NodeInterrupt is thrown in a conditional edge
+            if (e.name === "NodeInterrupt") {
```
jacoblee93, line 73: This unfortunately won't work in minified envs

You unfortunately need to do:

```ts
if (e.name === NodeInterrupt.unminifiable_name())
```

PR #460, opened 2024-09-05
[wip]: adding Managed Value channels

isahers1:


libs/langgraph/src/tests/pregel.test.ts
```diff
@@ -2094,12 +2173,50 @@ describe("StateGraph", () => {
...
+    }
+  }
+
   const AgentAnnotation = Annotation.Root({
     input: Annotation<string>,
     agentOutcome: Annotation<AgentAction | AgentFinish | undefined>,
     steps: Annotation<Step[]>({
       reducer: (x: Step[], y: Step[]) => x.concat(y),
     }),
+    context: Annotation<MyContextModel>(Context(makeContext)),
```
jacoblee93, line 2219: I think I would make this use the `Context` constructor directly rather than a function

It is nice to not do this for `Annotation` so that it more closely mimics TypeScript type declarations but for things that you're actually passing in I'd say constructor is more intuitive

libs/langgraph/src/channels/context.ts
```diff
@@ -0,0 +1,5 @@
+import { Context as ContextManagedValue } from "../managed/context.js";
+
+const Context = ContextManagedValue.of;
```
jacoblee93, line 3: At a glance this is a bit odd to me

libs/langgraph/src/managed/context.ts
```diff
@@ -0,0 +1,39 @@
+import { EmptyChannelError } from "../errors.js";
+import { ManagedValue, ConfiguredManagedValue } from "./base.js";
+
+// Type for a synchronous context manager
+export type ContextManager<T> = {
+    enter: () => T;
```
jacoblee93, line 6: Can you have multiple values that need to be exposed to `exit()`?

Wondering if it actually would be cleaner to use an iterator...

PR #471, opened 2024-09-08
feat(checkpoint-postgres): Postgres checkpointer

l4b4r4b4b4:
# Add PostgreSQL Checkpoint Saver Implementation

This PR introduces a PostgreSQL-based checkpoint saver implementation for LangGraph.js. This new `PostgresSaver` class provides an alternative to the existing `SqliteSaver`, allowing users to leverage PostgreSQL databases for storing and managing checkpoints in their LangGraph applications.

## Key Changes

1. Added `PostgresSaver` class that implements the `BaseCheckpointSaver` interface.
2. Implemented PostgreSQL-specific versions of `setup`, `getTuple`, `list`, `put`, and `putWrites` methods.
3. Updated the checkpointer to use PostgreSQL connection pooling for improved performance and resource management.
4. Added support for `checkpoint_ns` (namespace) in all relevant methods and database operations.
5. Ported unit tests from SqliteSaver for the new `PostgresSaver` class.

## Benefits

- Allows users to integrate LangGraph.js with existing PostgreSQL infrastructure.
- Provides better scalability and concurrent access compared to SQLite for larger applications.
- Maintains feature parity with the existing `SqliteSaver` implementation.

## Testing

The PR includes a full suite of unit tests that cover all major functionalities of the `PostgresSaver` class. These tests ensure that the PostgreSQL implementation behaves consistently with the existing SQLite implementation.

Start up PosgreSQL test instance by running the following command in the module directory.

```bash
docker-compose up -d && docker-compose logs -f
yarn && yarn test
```
grafik
grafik

## Usage

Users can create a `PostgresSaver` instance like this:

```typescript
import { PostgresSaver } from "@langchain/langgraph-checkpoint";

const postgresSaver = PostgresSaver.fromConnString("postgresql://user:password@localhost:5432/dbname");

### CODE REVIEW

**Comment by l4b4r4b4b4** [comment]
Was able to install my langgraph fork locall in ma containerised nextjs web and hooked it up to a defined graph. The graph executed successfully and the state checkpoints have been successfully stored in the PostgreSQL DB :partying_face:

However on the db side I keep getting the following error. Got an idea where that might come from?
`app-db-1  | 2024-09-10 12:39:54.200 UTC [78] FATAL:  role "root" does not exist`

File: libs/checkpoint-postgres/src/index.ts [hunk 1]
```diff
@@ -0,0 +1,330 @@
...
+          checkpoint_ns TEXT NOT NULL DEFAULT '',
+          checkpoint_id TEXT NOT NULL,
+          parent_checkpoint_id TEXT,
+          type TEXT,
+          checkpoint BYTEA NOT NULL,
+          metadata BYTEA NOT NULL,
+          PRIMARY KEY (thread_id, checkpoint_ns, checkpoint_id)
+        )`);
+      await client.query(`
+        CREATE TABLE IF NOT EXISTS writes (
```
vanb, line 71: Perhaps a more specific table name like `checkpoint_writes`? Would also be nice if names were configurable like MongoDbSaver.

jacoblee93 (commented):
Sorry for the delay!I think this is a great start - just had a closer look and the SQLite one

libs/checkpoint-postgres/src/index.ts
l4b4r4b4b4: Iirc, I just used the names as in sqlite checkpointer. But can sure give the tables a more telling name.

l4b4r4b4b4:
> Sorry for the delay!
> 
> I think this is a great start - just had a closer look and we should actually have a third table to take into account checkpoint versioning:
> 
> https://github.com/langchain-ai/langgraph/tree/main/libs/checkpoint-postgres/langgraph/checkpoint/postgres
> 
> I will try to have a look at adapting this tomorrow.

Interesting. So the sqlite one is not complete yet either? 🤔 

Gotta add to the earlier statement, the checkpointer successfully persists the state at break points and from all I can see resumes as wanted on user input.

brianpursley:
> take into account checkpoint versioning:

By this, are you referring to the database migrations (schema versioning)?

Defined here...
https://github.com/langchain-ai/langgraph/blob/29932bb59bad3684ad576d1721629e1e8e42c113/libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py#L18-L57

And applied here...
https://github.com/langchain-ai/langgraph/blob/29932bb59bad3684ad576d1721629e1e8e42c113/libs/checkpoint-postgres/langgraph/checkpoint/postgres/aio.py#L84-L106

If so, then @l4b4r4b4b4 I don't think it is so much that the sqlite checkpointer is incomplete, but that the python postgres checkpointer supports migrations, which allows the database to be schema to be "upgraded".

Whether that is needed in the js implementation or not, is up to the maintainers I suppose. I wouldn't stop the initial version from working, but would be helpful in the future if the schema needs to be changed.

brianpursley:
> take into account checkpoint versioning:

By this, are you referring to the database migrations (schema versioning)?

Defined here...
https://github.com/langchain-ai/langgraph/blob/29932bb59bad3684ad576d1721629e1e8e42c113/libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py#L18-L57

And applied here...
https://github.com/langchain-ai/langgraph/blob/29932bb59bad3684ad576d1721629e1e8e42c113/libs/checkpoint-postgres/langgraph/checkpoint/postgres/aio.py#L84-L106

If so, then @l4b4r4b4b4 I don't think it is so much that the sqlite checkpointer is incomplete, but that the python postgres checkpointer supports migrations, which allows the database to be schema to be "upgraded".

Whether that is needed in the js implementation or not, is up to the maintainers I suppose. I wouldn't stop the initial version from working, but would be helpful in the future if the schema needs to be changed.

jacoblee93:
> Edit: Or are you referring to the [checkpoint_blobs table](https://github.com/langchain-ai/langgraph/blob/29932bb59bad3684ad576d1721629e1e8e42c113/libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py#L36), which the python implementation has?

> (I am also interested in a postgres checkpoint saver, so I'm trying to understand what it would take for this to be implemented)

The `blobs` table is important for prod in Postgres for performance - channel values can be arbitrarily large.

jacoblee93:
Thank you very much for your patience!

PR #475, opened 2024-09-11
langgraph[minor]: Stores

bracesproul:
Adds stores:
- `BaseStore`
- `MemoryStore`
- `AsyncBatchStore`

Updates `PregelLoop.initialize` to set the store inside `config.configurable` under the key `__pregel_store` (aliased via `CONFIG_KEY_STORE` constant)

This will be used inside the managed values (memory) implementation to put/list values

@jacoblee93 / @nfcampos

### DIFF
--- a/libs/langgraph/src/pregel/loop.ts
+++ b/libs/langgraph/src/pregel/loop.ts
@@ -49,6 +49,7 @@ import {
 } from "./debug.js";
 import { PregelNode } from "./read.js";
 import { BaseStore } from "../store/base.js";
+import { AsyncBatchedStore } from "../store/batch.js";
 
 const INPUT_DONE = Symbol.for("INPUT_DONE");
 const INPUT_RESUMING = Symbol.for("INPUT_RESUMING");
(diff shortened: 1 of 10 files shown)


libs/langgraph/src/pregel/loop.ts
```diff
@@ -162,13 +169,17 @@ export class PregelLoop {
     this.streamKeys = params.streamKeys;
     this.nodes = params.nodes;
     this.skipDoneTasks = this.config.configurable?.checkpoint_id === undefined;
+    this.store = params.store;
   }
 
   static async initialize(params: PregelLoopInitializeParams) {
+    const configForManaged = patchConfigurable(params.config, {
+      [CONFIG_KEY_STORE]: params.store,
+    });
```
bracesproul, line 178: For reviewer: This is the main implementation of this PR (besides the store classes, however those have tests).

This is essentially supposed to port the `__enter__` method of `(Sync|Async)PregelLoop`and some code from `ChannelsManager` (setting the config).

PR #476, opened 2024-09-11
feat(langgraph): Adds stores & managed values (memory)

bracesproul:
TODO: add tests that verifies managed values get passed through, but aren't stored in checkpointer


libs/langgraph/src/managed/shared_value.ts
```diff
@@ -0,0 +1,122 @@
...
+    await instance.tick();
+    return instance as unknown as ManagedValue<Value>;
+  }
+
+  static on(scope: string): ConfiguredManagedValue<Value> {
+    return {
+      cls: SharedValue,
+      params: {
+        scope,
+        key: {},
```
bracesproul, line 70: in py this is typed as a string, but defaults to an object. figure out which it should be.

libs/langgraph/src/channels/base.ts
```diff
@@ -6,6 +6,10 @@ import {
 } from "@langchain/langgraph-checkpoint";
 import { EmptyChannelError } from "../errors.js";
 
+export function isBaseChannel(obj: unknown): obj is BaseChannel {
+  return obj != null && typeof (obj as BaseChannel).lc_graph_name === "string";
```
jacoblee93, line 10: I don't think we should do this as a rule (`Send` has this populated) - can we add some other field?

libs/langgraph/src/channels/base.ts
```diff
@@ -75,11 +79,15 @@ export function emptyChannels<Cc extends Record<string, BaseChannel>>(
   channels: Cc,
   checkpoint: ReadonlyCheckpoint
 ): Cc {
+  const filteredChannels = Object.fromEntries(
```
jacoblee93, line 82: This removes managed values?

examples/how-tos/shared-state.ipynb
```diff
@@ -0,0 +1,779 @@
...
+    "    throw new Error(\"Expected an AI message with tool calls.\");\n",
+    "  }\n",
+    "\n",
+    "  return (lastMessage as AIMessage).tool_calls?.length ? \"update_memory\" : END;\n",
+    "};\n",
+    "\n",
+    "const updateMemory = (\n",
+    "  state: typeof AgentAnnotation.State\n",
+    "): Partial<typeof AgentAnnotation.State> => {\n",
+    "  const toolCalls: ToolMessage[] = [];\n",
```
jacoblee93, line 369: These aren't tool calls, should be called `toolResponseMessages` or something

examples/how-tos/shared-state.ipynb
```diff
@@ -0,0 +1,779 @@
...
+   "cell_type": "code",
+   "execution_count": 2,
+   "id": "337ee88f",
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Result from running graph with input \"hi\":\n",
```
jacoblee93, line 71: This below example could use a few more comments

examples/how-tos/shared-state.ipynb
```diff
@@ -0,0 +1,779 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "b8c416fa-086a-491d-a7d3-57091f6413e3",
+   "metadata": {},
+   "source": [
+    "## Run graph on a different thread\n",
+    "\n",
+    "We can now run the graph on a different thread and see that it remembers facts about the user (specifically that the user likes pepperoni pizza and lives in SF):"
```
jacoblee93, line 635: Isn't this what we just did above?

libs/langgraph/src/managed/base.ts
```diff
@@ -0,0 +1,187 @@
...
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export abstract class ManagedValue<Value = any> {
+  runtime: boolean = false;
+
+  config: RunnableConfig;
+
+  private _promises: Promise<unknown>[] = [];
+
+  ls_is_managed_value = true;
```
jacoblee93, line 15: Call this `lg_is_managed_value` - `ls` is LangSmith

libs/langgraph/src/managed/base.ts
```diff
@@ -0,0 +1,187 @@
...
+
+  constructor(config: RunnableConfig, _params?: ManagedValueParams) {
+    this.config = config;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  static async initialize<Value = any>(
```
jacoblee93, line 22: Just make this an abstract method?

libs/langgraph/src/managed/shared_value.ts
```diff
@@ -0,0 +1,123 @@
...
+      this.ns = null;
+    } else if (config.configurable?.[this.scope]) {
+      const scopeValue = config.configurable[this.scope];
+      let scopedValueString = "";
+      try {
+        scopedValueString =
+          typeof scopeValue === "string"
+            ? scopeValue
+            : JSON.stringify(scopeValue);
+      } catch (_) {
```
jacoblee93, line 45: When would this fail?

libs/langgraph/src/managed/base.ts
```diff
@@ -0,0 +1,187 @@
...
+    _config: RunnableConfig,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    _args?: any
+  ): Promise<ManagedValue<Value>> {
+    throw new Error("Not implemented");
+  }
+
+  abstract call(step: number): Value;
```
jacoblee93, line 30: Why not make this a runnable?

libs/langgraph/src/pregel/algo.ts
```diff
@@ -143,7 +189,10 @@ export function _applyWrites<Cc extends Record<string, BaseChannel>>(
   tasks: WritesProtocol<keyof Cc>[],
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   getNextVersion?: (version: any, channel: BaseChannel) => any
-): void {
+): Record<string, PendingWriteValue[]> {
+  const filteredChannels = Object.fromEntries(
```
jacoblee93, line 193: See above, I *think* this is filtering out managed values but more clarity around naming would be nice.

libs/langgraph/src/pregel/algo.ts
```diff
@@ -356,22 +419,33 @@ export function _prepareNextTasks<
...
+                    step,
+                    (items: [keyof Cc, unknown][]) => writes.push(...items),
+                    processes,
+                    channels,
+                    managed,
+                    writes_
+                  ),
+                [CONFIG_KEY_READ]: (
+                  select_: Array<keyof Cc> | keyof Cc,
+                  fresh_: boolean = false
```
jacoblee93, line 434: Why are these params suffixed?

Also what do they represent?

libs/langgraph/src/pregel/index.ts
```diff
@@ -775,7 +875,12 @@ export class Pregel<
       await runManager?.handleChainError(e);
       throw e;
     } finally {
-      await Promise.all(loop?.checkpointerPromises ?? []);
+      this.store?.stop();
+      loop?.store?.stop();
```
jacoblee93, line 879: Why would there be two different values here?

libs/langgraph/src/store/batch.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  /**
+   * Runs the task that processes queued operations in batches.
+   * This method runs continuously until the store is stopped,
+   * or the process is terminated.
+   * @returns {Promise<void>} A promise that resolves when the task is complete.
+   */
+  private async runTask(): Promise<void> {
+    while (this.running) {
+      await new Promise((resolve) => {
```
jacoblee93, line 125: What is this for?

jacoblee93: If it is necessary, add a comment explaining why. But it seems like a race condition waiting to happen.

libs/langgraph/src/store/batch.ts
```diff
@@ -0,0 +1,162 @@
...
+    this.running = false;
+  }
+
+  /**
+   * Runs the task that processes queued operations in batches.
+   * This method runs continuously until the store is stopped,
+   * or the process is terminated.
+   * @returns {Promise<void>} A promise that resolves when the task is complete.
+   */
+  private async runTask(): Promise<void> {
```
jacoblee93, line 123: This is a bad name, you're basically calling it `doThing()`

libs/langgraph/src/store/batch.ts
```diff
@@ -0,0 +1,162 @@
...
+    while (this.running) {
```
jacoblee93, line 124: Is there really no better way to do this? For example, can the loop call this at the end of each tick?

jacoblee93:
All tests have been ported from Python?

jacoblee93 (commented):
x

libs/langgraph/src/pregel/loop.ts
```diff
@@ -232,6 +257,14 @@ export class PregelLoop {
     this.checkpointerPromises.push(this._checkpointerChainedPromise);
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  protected async updateMv(key: string, values: any[]) {
```
jacoblee93, line 261: nit: avoid abbreviations like this - just `updateManagedValues`

libs/langgraph/src/managed/base.ts
bracesproul: Can't make static methods abstract

libs/langgraph/src/managed/shared_value.ts
bracesproul: Good catch, I don't think it will. removed

libs/langgraph/src/pregel/algo.ts
bracesproul: it is, will rename

libs/langgraph/src/pregel/algo.ts
bracesproul: suffix to show we shouldn't do anything with them besides passing args through. They're the two args which we aren't populating, but this func requires so we create a `partial` func. No change from what we had before, except a new type error occurred with the existing implementation. (plus I think naming conflicts)

libs/langgraph/src/store/batch.ts
bracesproul: ya thats a good idea, pushed fix

libs/langgraph/src/store/batch.ts
bracesproul: renamed to `processBatchQueue`

libs/langgraph/src/pregel/index.ts
bracesproul: fixed

libs/langgraph/src/managed/base.ts
```diff
@@ -0,0 +1,187 @@
...
+            throw error;
+          }
+        }
+      }
+    }
+  }
+}
+
+export function isManagedValue(value: unknown): value is typeof ManagedValue {
+  if (typeof value === "object" && value && "lg_is_managed_value" in value) {
```
jacoblee93, line 169: Don't forget to update name here

PR #488, opened 2024-09-17
Add link to langchain in the overview

Masstronaut:


jacoblee93:
You'll need to apply this to the inner README as well to pass CI

Issue #474, opened 2024-09-11
Subgraph Issues: Invalid Update For Channel query


justinlevi:
I'm trying to build a RAG graph with a few extra nodes to experiment with a few ideas and I thought it would be interesting to try and create a subgraph for the rag pipeline. Unfortunately I can't figure out where this error is coming from and how to solve it. Any direction is greatly appreciated. 

I've tried everything I can think of to return the state from the subgraph to match the parent graph state as well as even just return an empty state in the last node in the graph, but nothing seems to work. I feel like the issue is with the handoff back to the parent graph, but I'm not sure I fully understand how to return properly. 

I'm seeing the following error: 

```
[Nest] 51582  - 09/10/2024, 8:37:25 PM   ERROR [ExceptionsHandler] Invalid update for channel query. Values: How can I create a managed cluster?,What are the step-by-step instructions to create a managed cluster?,What files are needed to create a managed cluster?,What code is required to create a managed cluster?,How can I deploy a managed cluster to a specific bare metal host?,What are the step-by-step instructions to deploy to a specific bare metal host?,What files are needed to deploy to a specific bare metal host?,What code is required to deploy to a specific bare metal host?

Error:
InvalidUpdateError: Invalid update for channel query. Values: How can I create a managed cluster?,What are the step-by-step instructions to create a managed cluster?,What files are needed to create a managed cluster?,What code is required to create a managed cluster?,How can I deploy a managed cluster to a specific bare metal host?,What are the step-by-step instructions to deploy to a specific bare metal host?,What files are needed to deploy to a specific bare metal host?,What code is required to deploy to a specific bare metal host?

Error:
    at _applyWrites (/Users/justinwinter/projects/docbot/backend/node_modules/@langchain/langgraph/dist/pregel/index.cjs:652:27)
    at CompiledStateGraph._transform (/Users/justinwinter/projects/docbot/backend/node_modules/@langchain/langgraph/dist/pregel/index.cjs:475:17)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async CompiledStateGraph._transformStreamWithConfig (/Users/justinwinter/projects/docbot/backend/node_modules/@langchain/core/dist/runnables/base.cjs:308:30)
    at async CompiledStateGraph.transform (/Users/justinwinter/projects/docbot/backend/node_modules/@langchain/langgraph/dist/pregel/index.cjs:536:26)
[...]

PR #499, opened 2024-09-20
feat: Adds support for subgraph interrupts, streaming, update/get state, and send API

jacoblee93:


Masstronaut:
There's a case from #474 we should make sure to cover & test for.

PR #512, opened 2024-09-24
WIP: Implement types for Pregel.stream using recursive type

nfcampos:
Fixes # (issue)

### DIFF
--- a/libs/langgraph/src/pregel/index.ts
+++ b/libs/langgraph/src/pregel/index.ts
@@ -49,8 +49,10 @@ import {
   PregelExecutableTask,
   PregelInterface,
   PregelParams,
+  SingleStreamMode,
   StateSnapshot,
   StreamMode,
+  StreamOutput,
 } from "./types.js";
 import {
   GraphRecursionError,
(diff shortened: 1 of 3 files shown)


libs/langgraph/src/pregel/index.ts
```diff
@@ -198,12 +198,13 @@ export type PregelInputType = any;
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type PregelOutputType = any;
 
-export class Pregel<
-    Nn extends StrRecord<string, PregelNode>,
-    Cc extends StrRecord<string, BaseChannel | ManagedValueSpec>
+export class Pregel<Nodes extends NodesType, Channels extends ChannelsType>
+  extends Runnable<
+    PregelInputType,
+    PregelOutputType,
```
Masstronaut, line 204: We need a way to evaluate `StreamOutput<Mode, Nodes, Channels, Schema>` here. This has 2 problems:
1. We don't know the mode until the user does `invoke` or `stream`
2. Can we deduce `Schema` from the Channels/Nodes? Currently the usage in `stream`'s return type is being passed any.

PR #516, opened 2024-09-26
fix: Fix message serialization in prebuilt reducer, add subgraph docs

bracesproul:
Fork of `Jacob/subgraphs`

Added three subgraph docs from python (atm only `subgraph.ipynb` has been ported)

Notes:
- Return value when specifying `subgraphs: true` is weird, not sure what the different array items should represent
  - Note on `PregelOptions.subgraphs` is `Whether to stream subgraphs.` but it lets me pass it to `.invoke` calls (and it's not returning an iterator). Is this a bad jsdoc, bug, something else? Not really sure what passing this field does.


libs/checkpoint/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@langchain/langgraph-checkpoint",
-  "version": "0.0.6",
+  "version": "0.0.7",
```
bracesproul, line 3: npm released version is `0.0.7`, I didn't manually make this change so not sure how it got here?

PR #518, opened 2024-09-26
fix(langgraph): Update store api

bracesproul:


libs/checkpoint/src/store/base.ts
```diff
@@ -0,0 +1,210 @@
...
+   */
+  createdAt: Date;
+  /**
+   * Timestamp of last update.
+   */
+  updatedAt: Date;
+  /**
+   * Timestamp of last access.
+   */
+  lastAccessedAt: Date;
```
hinthornw, line 39: We are removing this

libs/checkpoint/src/store/base.ts
```diff
@@ -0,0 +1,210 @@
...
+   * The stored data as an object. Keys are filterable.
+   */
+  value: Record<string, any>;
+  /**
+   * Relevance scores for the item.
+   * Keys can include built-in scores like 'recency' and 'relevance',
+   * as well as any key present in the 'value' object. This allows
+   * for multi-dimensional scoring of items.
+   */
+  scores: Record<string, number>;
```
hinthornw, line 17: And this

libs/checkpoint/src/store/memory.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export class MemoryStore extends BaseStore {
+  private data: Map<string, Map<string, Item>> = new Map();
+
+  async batch<Op extends Operation[]>(
+    operations: Op
+  ): Promise<OperationResults<Op>> {
+    const results = [];
+
+    for (const op of operations) {
```
hinthornw, line 16: Will want to do the listNamespaces operation here

libs/checkpoint/src/store/base.ts
```diff
@@ -0,0 +1,259 @@
...
+    };
+
+    const batchResults = await this.batch<[ListNamespacesOperation]>([op]);
+    return batchResults[0];
+  }
+
+  /**
+   * Stop the store. No-op if not implemented.
+   */
+  stop(): void {
```
hinthornw, line 249: This is all for batching? What checkpointer uses this?

libs/checkpoint/src/store/base.ts
```diff
@@ -0,0 +1,259 @@
...
+export abstract class BaseStore {
+  /**
+   * Execute a batch of operations.
+   * @param _operations An array of operations to execute.
+   * @returns A promise that resolves to the results of the operations.
+   */
+  async batch<Op extends Operation[]>(
+    _operations: Op
+  ): Promise<OperationResults<Op>> {
+    throw new Error("Method not implemented.");
```
hinthornw, line 141: Would be nice to have a better error message for this

libs/checkpoint/src/store/base.ts
bracesproul: `AsyncBatchedStore` needs it to stop the `while` loop at the end of an execution (if not, it will continue looping until the process exits.

PR #530, opened 2024-09-28
Remove StateGraphArgs

steren:
StateGraphArgs is not used in the code

And it anyway throws an error: 

```
import { StateGraph, StateGraphArgs } from "@langchain/langgraph";
                     ^^^^^^^^^^^^^^
SyntaxError: The requested module '@langchain/langgraph' does not provide an export named 'StateGraphArgs'
```


jacoblee93:
Agh, nice catch thanks!

bracesproul (changes requested):
Thank you! However this PR will fail ci since we have two readmes that must be in sync. Could you push up a commit updating the other readme file with this change?

jacoblee93:
I got it, thanks!

PR #543, opened 2024-10-02
Allowing checkpointers' serializers to be asynchronous

wpoynter:
Updated the SerializationProtocol interface to allow `dumpsTyped` and `loadsTyped` to be run synchronously or asynchronously.
Applied `await` to calls of each function throughout the existing checkpoint savers and tests.

This should allow future users to provide asynchronous serialization and deserialization for their savers, without breaking any current implementations.

### DIFF
--- a/libs/checkpoint-mongodb/src/index.ts
+++ b/libs/checkpoint-mongodb/src/index.ts
@@ -214,9 +214,12 @@ export class MongoDBSaver extends BaseCheckpointSaver {
         `The provided config must contain a configurable field with a "thread_id" field.`
       );
     }
-    const [checkpointType, serializedCheckpoint] =
-      this.serde.dumpsTyped(checkpoint);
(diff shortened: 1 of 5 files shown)


jacoblee93:
This will break folks using custom checkpointers - it's early days but I'd prefer not to.

What's the use case?

wpoynter:
> This will break folks using custom checkpointers - it's early days but I'd prefer not to. Will think through a backwards compatible way to do this.

@jacoblee93 Not a problem, but can you explain how this will break existing users with custom checkpointers? I think I am missing something that it would be worth me learning.

jacoblee93:
I mean this will be breaking if you've implemented your own custom checkpointer (which I know a few have) since they'll need to change their code to `await` serde methods.

benjamincburns:
Closing this for now, per above comments.

Issue #541, opened 2024-10-01
Feature request: Create an acceptance test suite for checkpointers


benjamincburns:
As @hgoona mentioned in [this comment](https://github.com/langchain-ai/langgraphjs/issues/536#issuecomment-2384548941_), the requirements for custom checkpointers aren't abundantly obvious from the docs, or even the LangChain-provided checkpointer implementations.

It would be helpful if a black-box acceptance test suite existed for checkpointers, ideally one that could be used both by the LangGraph team, and authors of third party checkpointer implementations.

The suite would need to exercise all of the requirements of the checkpointer contract, such that if an arbitrary checkpointer passes this suite, the author and community would have high confidence that the checkpointer in question fulfills all of LangGraph's requirements for storing and retrieving persisted state.

Ideally this would also exist for stores. I'm happy to open another issue for those if the maintainers would prefer to track these separately.

vbarda:
@benjamincburns thanks for the feedback! this is a great point and we actually do plan to do this for both Python and JS, we'll follow up here once we have more details

benjamincburns:
@vbarda are you guys interested in external contributions for this? I'd imagine it'd be a riskier contribution for a third party to make given that anyone external to the team would essentially be reverse-engineering the interface specification, not to mention making some assumptions around how you guys would want to structure the test suite, what tooling to use, etc.

That said, I'm finding that I need something like this for my own project, so I'm likely to be building something of this ilk soon. If the thing that I come up with looks like it might be more broadly useful, I'd be happy to submit it as a PR. I'd just expect that I'd need some help from the LangGraph team to massage it into a shape that fits nicely in this repository.

hgoona:
@benjamincburns Noted! Though, I don't have access to that Slack. Are you on the LangChain Discord server to talk further?

hgoona:
@benjamincburns In case you're interested, I've posted a draft Saver for SurrealDB here where I've just tried to mimic the Sqlite version: https://github.com/langchain-ai/langgraphjs/issues/545#issuecomment-2392531170

I've annotated the test a bit better than the original from SqliteSaver so it may be useful if applied to your case. Currently my `SdbSaver` is failing two tests. Would love any help I can get on it!

benjamincburns:
@vbarda I've started an acceptance suite that I intend to contribute. So far it's just a pile of unit tests, but I intend to fill out some integration tests in the `scenarios`, as well.

[...]

PR #555, opened 2024-10-07
fix: return correct checkpointId in parentConfig in MemorySaver.getTuple

benjamincburns:
Found while working on #541.

Prior to this fix, when `MemorySaver.getTuple` was called to fetch a specific `checkpoint_id` that was a child of some other checkpoint, the value of `parentConfig.configurable.checkpoint_id` in the returned `CheckpointTuple` would be assigned to the child's `checkpoint_id` rather than the parent's.

### DIFF
--- a/libs/checkpoint/src/memory.ts
+++ b/libs/checkpoint/src/memory.ts
@@ -99,7 +99,7 @@ export class MemorySaver extends BaseCheckpointSaver {
             configurable: {
               thread_id,
               checkpoint_ns,
-              checkpoint_id,
+              checkpoint_id: parentCheckpointId,
             },
           };
         }


jacoblee93:
Oof good catch, thanks and thanks for looking at that!

I'm working on updating #471 as well, would be very useful for that.

benjamincburns:
@jacoblee93 ah, that's serendipitous! Feel free to DM me on the LangChain community slack or leave comments on #541 if you have feedback about the test suite. Details in [this comment](https://github.com/langchain-ai/langgraphjs/issues/541#issuecomment-2395885866). It's very WIP and I intend to do some clean-up, so I wouldn't give it more than a quick glance to see if there's anything about the high-level structure or strategy that you think would be difficult to accept in a PR.

Either way, I'm hoping to be able to submit it by the end of the week.

PR #561, opened 2024-10-08
fix: make MemorySaver.put throw if thread_id not passed

benjamincburns:
Found while working on #541.

Prior to this change, the value of `config.configurable.thread_id` was unchecked in `MemorySaver.getTuple`. Admittedly this is a minor problem, without this change it returns `undefined` rather than throwing, but throwing seems more appropriate here, as it better informs the caller that they forgot to supply a `thread_id`.

I also see that this value is unchecked in `checkpoint-sqlite` and `checkpoint-mongodb`, and it looks like they may behave similarly (return `undefined` rather than throwing), although I haven't run them against my tests to check just yet. Happy to submit a PR for those (or include them in this one) if you want this behaviour change everywhere).

### DIFF
--- a/libs/checkpoint/src/memory.ts
+++ b/libs/checkpoint/src/memory.ts
@@ -60,6 +60,12 @@ export class MemorySaver extends BaseCheckpointSaver {
     const checkpoint_ns = config.configurable?.checkpoint_ns ?? "";
     let checkpoint_id = config.configurable?.checkpoint_id;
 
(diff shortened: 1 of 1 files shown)


benjamincburns:
Closing this for now, as I think it will be a better approach if I hold off on these smaller PRs until I've got a draft PR together for #541. That way we can discuss in the PR the best way to tackle the inconsistencies between the various LangChain provided checkpoint saver implementations.

PR #570, opened 2024-10-09
feat(langgraph): Allow option to prevent subgraphs inheriting checkpointer

jacoblee93:
@hinthornw

### DIFF
--- a/libs/langgraph/src/errors.ts
+++ b/libs/langgraph/src/errors.ts
@@ -111,6 +111,15 @@ export class MultipleSubgraphsError extends Error {
 
 /**
  * Used for subgraph detection.
- * TODO: Move to globalThis for bundlers
  */
-export const _SEEN_CHECKPOINT_NS = new Set();
+export const getSubgraphsSeenSet = () => {
+  if (
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (globalThis as any)[Symbol.for("LG_CHECKPOINT_SEEN_NS_SET")] === undefined
+  ) {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (globalThis as any)[Symbol.for("LG_CHECKPOINT_SEEN_NS_SET")] = new Set();
+  }
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  return (globalThis as any)[Symbol.for("LG_CHECKPOINT_SEEN_NS_SET")];
+};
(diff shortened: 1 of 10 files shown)


libs/langgraph/src/pregel/loop.ts
```diff
@@ -311,7 +315,13 @@ export class PregelLoop {
       // Start the store. This is a batch store, so it will run continuously
       store.start();
     }
-
+    if (checkSubgraphs && isNested && params.checkpointer !== undefined) {
+      if (_SEEN_CHECKPOINT_NS.has(config.configurable?.checkpoint_ns)) {
+        throw new MultipleSubgraphsError();
```
hinthornw, line 320: Would be nice to have a better error message here (if we know the name where this is occurring, recommendation on what to do)

libs/langgraph/src/errors.ts
```diff
@@ -97,3 +97,20 @@ export class InvalidUpdateError extends Error {
...
+  }
+
+  static get unminifiable_name() {
+    return "MultipleSubgraphError";
+  }
+}
+
+/**
+ * Used for subgraph detection.
+ * TODO: Move to globalThis for bundlers
```
hinthornw, line 114: ooc: what's blocking us from doing now?

libs/langgraph/src/pregel/retry.ts
```diff
@@ -167,6 +167,16 @@ async function _runWithRetry(
           2
         )} seconds (attempt ${attempts}) after ${errorName}: ${error}`
       );
+      // Clear checkpoint_ns seen (for subgraph detection)
+      const checkpointNs = pregelTask.config?.configurable?.checkpoint_ns;
+      if (checkpointNs) {
+        _SEEN_CHECKPOINT_NS.delete(checkpointNs);
```
hinthornw, line 173: does this raise an error if not found?

hinthornw (commented):
thank you! I think it makes sense? Any tests we're missing?

libs/langgraph/src/errors.ts
jacoblee93: Feels premature and don't want to rely on globals if possible, maybe would be a good idea.

libs/langgraph/src/pregel/retry.ts
jacoblee93: No, should be idempotent:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/delete

Issue #550, opened 2024-10-05
Feature Request: Type Safe ConfigAnnotation


debkanchan:
If I define a `ConfigAnnotation` like this
```ts
const ConfigAnnotation = Annotation.Root({
  model: Annotation<string>({
    reducer: (_, current) => current,
    default: () => "gpt-4o",
  }),
});
```
I should be able to do this
```ts
const agent = async (
  state: typeof MessagesAnnotation.State,
  config?: RunnableConfig,
): Promise<Partial<typeof MessagesAnnotation.State>> => {
  const response = await researchPromptTemplate
    .pipe(llms[config.configurable.model].bindTools(tools))
    .invoke(
      {
        messages: state.messages,
      },
      config
    );

  return {
    messages: [response],
  };
};
```
Without facing `Element implicitly has an 'any' type because expression of type 'any' can't be used to index type '{ "gpt-4o": AzureChatOpenAI…`

jacoblee93:
Should definitely support this, thanks for flagging!

debkanchan:
Nice! How do I acess it inside a node? For example what should I change within this node for the type hints to take effect?

```ts
const agent = async (
  state: typeof GraphStateAnnotation.State,
  config?: RunnableConfig,
): Promise<Partial<typeof GraphStateAnnotation.State>> => {
  const response = await researchPromptTemplate
    .pipe(llms["gpt-4o"].bindTools(tools))
    .invoke(
      {
        messages: state.messages,
      },
      config,
    );

  return {
    messages: [response],
  };
};
```

jacoblee93:
This should now work:

https://langchain-ai.github.io/langgraphjs/how-tos/configuration/#config-schema

PR #572, opened 2024-10-10
feat(langgraph): Allow configurable typing in nodes and conditional edges

jacoblee93:
Fully addresses #550


libs/langgraph/src/graph/graph.ts
```diff
@@ -29,22 +29,32 @@ import {
...
-export class Branch<IO, N extends string> {
+export class Branch<
+  IO,
+  N extends string,
+  CallOptions extends LangGraphRunnableConfig = LangGraphRunnableConfig
+> {
   condition: (
     input: IO,
-    config?: RunnableConfig
+    config: CallOptions
```
nfcampos, line 57: Now it's a required arg?

nfcampos: Ah this is an internal function, ignore me

nfcampos (dismissed):

libs/langgraph/src/graph/graph.ts
jacoblee93: I usually do!

PR #576, opened 2024-10-10
fix(checkpoint-sqlite): fix invalid SQL syntax on list

benjamincburns:
Prior to this change, if no filters (including those defined by `config.configurable`) were passed to the `SqliteSaver.list` method, it would throw due to there being a `WHERE` keyword with an empty predicate list in the prepared `SELECT` query. This change builds the query iteratively to avoid this error.

This change also fixes a potential (albeit unlikely) SQL injection attack vector in the same query. Prior to this change, the `limit` value was being directly concatenated into the `sql` string, rather than being sanitized first.

Found while working on #541, thanks to [this test](https://github.com/benjamincburns/langgraphjs/blob/d95f4aafac23d2f39c12a7c1c0a2f2ede0815ba8/libs/checkpoint-validation/src/tests/spec/list.ts#L148-L170).

### DIFF
--- a/libs/checkpoint-sqlite/src/index.ts
+++ b/libs/checkpoint-sqlite/src/index.ts
@@ -168,12 +168,26 @@ CREATE TABLE IF NOT EXISTS writes (
     const { limit, before } = options ?? {};
     this.setup();
     const thread_id = config.configurable?.thread_id;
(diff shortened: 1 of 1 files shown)


benjamincburns:
Going to reopen this with a slightly wider scope (other fixes to the `list` method).

PR #577, opened 2024-10-10
fix(checkpoint-postgres): store channel_values correctly

benjamincburns:
Prior to this change, `PostgresSaver` wouldn't store `channel_values` if `newVersions` wasn't passed to the `put` method. This is due to a [filter check in `_dumpBlobs`](https://github.com/langchain-ai/langgraphjs/blob/45e59e64175a89a7cf782015b011b7d348ef835f/libs/checkpoint-postgres/src/index.ts#L186) that checks each channel name to ensure it appears in the channel versions object prior to serializing it for storage.

Upon inspecting the other `CheckpointSaver` types, I noticed that [`MemorySaver`](https://github.com/langchain-ai/langgraphjs/blob/45e59e64175a89a7cf782015b011b7d348ef835f/libs/checkpoint/src/memory.ts#L270-L274), [`MongoDBSaver`](https://github.com/langchain-ai/langgraphjs/blob/45e59e64175a89a7cf782015b011b7d348ef835f/libs/checkpoint-mongodb/src/index.ts#L204-L208), and [`SqliteSaver`](https://github.com/langchain-ai/langgraphjs/blob/45e59e64175a89a7cf782015b011b7d348ef835f/libs/checkpoint-sqlite/src/index.ts#L217-L221)) all ignore [the `newVersions` argument to the `BaseCheckpointSaver.put` method](https://github.com/langchain-ai/langgraphjs/blob/45e59e64175a89a7cf782015b011b7d348ef835f/libs/checkpoint/src/base.ts#L148), so I _think_ this is the correct fix. From what I can see by searching references, no test code (apart from the tests this PR changes) passes the `newVersions` argument.

That said, the **one** reference to `BaseCheckpointSaver.put` that I found in actual shipped code is in [`PregelLoop._putCheckpoint`](https://github.com/langchain-ai/langgraphjs/blob/45e59e64175a89a7cf782015b011b7d348ef835f/libs/langgraph/src/pregel/loop.ts#L780-L785) (via [`_checkpointerPutAfterPrevious`](https://github.com/langchain-ai/langgraphjs/blob/45e59e64175a89a7cf782015b011b7d348ef835f/libs/langgraph/src/pregel/loop.ts#L368)), and this _does_ pass the `newVersions` argument.

So I have to ask, **are all of the other `CheckpointSaver` implementations, and all of the existing tests implemented incorrectly, or is this the correct fix?**

Found this while working on #541, thanks to [the `getTuple` tests](https://github.com/benjamincburns/langgraphjs/blob/d95f4aafac23d2f39c12a7c1c0a2f2ede0815ba8/libs/checkpoint-validation/src/tests/spec/getTuple.ts) (all of them failed when I [wired them up to `PostgresSaver`](https://github.com/benjamincburns/langgraphjs/blob/d95f4aafac23d2f39c12a7c1c0a2f2ede0815ba8/libs/checkpoint-validation/src/tests/postgres.spec.ts#L64)).

### DIFF
--- a/libs/checkpoint-postgres/src/index.ts
+++ b/libs/checkpoint-postgres/src/index.ts
@@ -298,7 +298,9 @@ export class PostgresSaver extends BaseCheckpointSaver {
    * @param config The config to use for retrieving the checkpoint.
    * @returns The retrieved checkpoint tuple, or undefined.
    */
-  async getTuple(config: RunnableConfig): Promise<CheckpointTuple | undefined> {
+  async getTuple(
+    config: RunnableConfig
+  ): Promise<CheckpointTuple | undefined> {
     const {
       thread_id,
       checkpoint_ns = "",
@@ -416,8 +418,7 @@ export class PostgresSaver extends BaseCheckpointSaver {
   async put(
     config: RunnableConfig,
     checkpoint: Checkpoint,
-    metadata: CheckpointMetadata,
-    newVersions: ChannelVersions
+    metadata: CheckpointMetadata
   ): Promise<RunnableConfig> {
     if (config.configurable === undefined) {
       throw new Error(`Missing "configurable" field in "config" param`);
@@ -443,7 +444,7 @@ export class PostgresSaver extends BaseCheckpointSaver {
         thread_id,
         checkpoint_ns,
         checkpoint.channel_values,
-        newVersions
+        checkpoint.channel_versions
       );
       for (const serializedBlob of serializedBlobs) {
         await client.query(UPSERT_CHECKPOINT_BLOBS_SQL, serializedBlob);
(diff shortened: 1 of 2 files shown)


hgoona:
@benjamincburns are you saying the current PostgresSaver implementation is faulty?

benjamincburns:
I don't know whether it is faulty or not. It looks to me like either it is faulty, or the other three are. It seems unlikely that it would be the other three, especially since this one was just introduced, but I can't say for certain, which is why I asked.

jacoblee93:
Hey @benjamincburns, respecting `newVersions` is the correct thing to do. Thanks so much for flagging this!

benjamincburns:
@jacoblee93 ah, glad I asked! I'll take a crack at it and reply back here if I'm unable to make it work for some reason.

PR #578, opened 2024-10-11
fix(checkpoint-postgres): fix WHERE clause generation on list

benjamincburns:
Found while working on #541. Prior to this change, if you called the `list` method in `PostgresSaver` with an empty `configurable` object (e.g. `saver.list({ configurable: {} })`), you'd get back no results. This was due to a small error in the branching logic for building up the `WHERE` clause that is used to query for checkpoints.

Also, like in #576, I noticed that the `LIMIT` clause was being directly concatenated into the query without sanitization (a potential SQL injection vulnerability), so I went ahead and fixed that in this commit as well.


jacoblee93 (approved):
Thank you! Looks great.

jacoblee93 (approved):

PR #582, opened 2024-10-11
fix(checkpoint-sqlite): list method bug fixes

benjamincburns:
This PR fixes a few issues in the `list` method of `SqliteSaver` that were found while working on #541.

- Fixes a SQL syntax error resulting from an empty `WHERE` clause when none of the filtering options are specified
- Adds missing support for filtering on `config.configurable.checkpoint_ns`
- Sanitizes `limit` field prior to concatenating
- Adds support for the `filter` field in the `options` argument by parsing the stored metadata in a CTE

Submitting these together because they're all somewhat intertwined.

### DIFF
--- a/libs/checkpoint-sqlite/src/index.ts
+++ b/libs/checkpoint-sqlite/src/index.ts
@@ -168,6 +168,7 @@ CREATE TABLE IF NOT EXISTS writes (
     const { limit, before } = options ?? {};
     this.setup();
     const thread_id = config.configurable?.thread_id;
+    const checkpoint_ns = config.configurable?.checkpoint_ns;
 
     let sql =
       "SELECT thread_id, checkpoint_ns, checkpoint_id, parent_checkpoint_id, type, checkpoint, metadata FROM checkpoints";
@@ -176,6 +177,10 @@ CREATE TABLE IF NOT EXISTS writes (
 
     if (thread_id) {
       whereClause.push("thread_id = ?");
+    }
+
+    if (checkpoint_ns !== undefined && checkpoint_ns !== null) {
+      whereClause.push("checkpoint_ns = ?");
     }
 
     if (before?.configurable?.checkpoint_id !== undefined) {
--- a/libs/checkpoint-sqlite/src/index.ts
+++ b/libs/checkpoint-sqlite/src/index.ts
@@ -194,7 +194,8 @@ CREATE TABLE IF NOT EXISTS writes (
     sql += ` ORDER BY checkpoint_id DESC`;
 
     if (limit) {
-      sql += ` LIMIT ${limit}`;
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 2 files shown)


libs/checkpoint-sqlite/src/index.ts
```diff
@@ -165,42 +165,79 @@ CREATE TABLE IF NOT EXISTS writes (
...
 
-    const args = [thread_id, before?.configurable?.checkpoint_id].filter(
-      Boolean
-    );
+    const args = [
+      thread_id,
+      checkpoint_ns,
+      before?.configurable?.checkpoint_id,
+      ...Object.values(filter ?? {}),
+    ].filter((value) => value !== undefined && value !== null);
```
benjamincburns, line 240: Note: using `Boolean` here as before was failing due to the default namespace being a falsey value (empty string).

libs/checkpoint-sqlite/src/index.ts
```diff
@@ -165,19 +165,80 @@ CREATE TABLE IF NOT EXISTS writes (
...
+    const checkpoint_ns = config.configurable?.checkpoint_ns;
+
+    const withClause =
+      filter && Object.keys(filter).length
+        ? "WITH augmented_checkpoints AS (\n" +
+          "  SELECT \n" +
+          "    ck.thread_id,\n" +
+          "    ck.checkpoint_ns,\n" +
+          "    ck.checkpoint_id,\n" +
+          "    jsonb(CAST(metadata as TEXT)) AS json_metadata\n" +
```
jacoblee93, line 180: Why do we need this separately vs. just returning it as a field in the initial query

benjamincburns: IIRC I didn't want to modify the structure of the result set returned by the query (`CheckpointRow[]`), as ultimately I was only doing this to support a filtering operation. That said, I can't recall why I opted for a CTE here when  I could've likely just done the following:

```
    whereClause.push(
      ...Object.entries(filter ?? {})
        .filter(([_, value]) => value !== undefined)
        .map(([key]) => `jsonb(CAST(metadata as TEXT))->>'$.${key}' = ?`)
    );
```

Of course, the ideal chang would be to migrate `checkpoint.metadata` to be a `jsonb` column rather than a blob. If it's all the same to you, my preference would be to avoid the migration now, but add a note to #592 to change up the column type at the same time, as #592 necessitates a migration anyway.

benjamincburns: Also I I just realised that I'm appending the filter object's `key` field directly into the query. This could become a SQL injection vulnerability (e.g. in an implementation that defines the filter in frontend code and passes it through an API unchecked). I'll add an `Array.filter` step into this to check the key names so that can't happen.

jacoblee93: Yeah suppose that could happen - good thinking

jacoblee93:
Thank you!

PR #583, opened 2024-10-11
fix(checkpoint): add support for options.filter in MemorySaver.list

benjamincburns:
Per the title, adds support for the `options.filter` argument to `MemorySaver`.list (another one from #541).

I suppose this could be a `feat` instead of a `fix`, but I opted to call it the latter as it's a part of the core checkpoint interface that was never implemented in `MemorySaver`.


jacoblee93:
LGTM, thank you!

Issue #595, opened 2024-10-14
`MongoDbSaver` ignores `newVersions` argument on `put` - not storing deltas


benjamincburns:
Found while working on #541. Per discussion on #577, `MongoDbSaver` is ignoring the `newVersions` argument, and as a result it's storing the entire contents of `checkpoint.channel_values` on every call.

jacoblee93:
Thanks for creating this! Yes need to update these.

Issue #594, opened 2024-10-14
`SqliteSaver` ignores `newVersions` argument on `put` - not storing deltas


benjamincburns:
Found while working on #541. Per discussion on #577, `SqliteSaver` is ignoring the `newVersions` argument, and as a result it's storing the entire contents of `checkpoint.channel_values` on every call.

trollboss2572:
Good morning, we are a group of students at the University of Toronto currently investigating the issue mentioned and believe that we could help tackle this. Can we get this opened as an issue and assigned to me? I've commented on a similar task and since these tasks are related, I believe that we could handle both of these issues simultaneously. @jacoblee93

benjamincburns:
@trollboss2572 - not to discourage you from contributing, but I have a fix for this in the works, and I hope to get it into a PR by the weekend.

That said, feel free to DM me on the LangChain community slack if you have any questions about the checkpointer code.

I'm not a member of the LangChain team, so definitely listen to them over me, but if you're looking to make a contribution to this project, I'd be happy to advise you on how to build a checkpointer for a new storage backend.

If you're up to it, one idea may be to contribute a generic RDBMS checkpointer that uses a lightweight ORM instead of directly composing SQL statements. That way it can handle multiple storage backends, depending on what "driver" library you install alongside of it as a peer dependency.

It could also be useful to contribute checkpointers for popular NoSQL storage backends, like DynamoDB, Azure CosmosDB, BigTable, Firebase, Supabase, etc.

Also as a fellow contributor, it's a good idea to leave a comment saying that you plan on working on something (mostly to find out if someone already is), but for most projects like this they don't assign issues to volunteer contributors, as they quite often disappear before delivering them, and it blocks others from starting a solution of their own.

If you have a simple fix in mind for something already, your best bet is to just drop a PR. The risk though is that it may not be accepted, so you may end up wasting some time. In more complex cases, you can reduce that risk by detailing your planned fix as a comment in the issue, with a request for feedback.

Hope that helps!

nityam2002:
> It could also be useful to contribute checkpointers for popular NoSQL storage backends, like DynamoDB, Azure CosmosDB, BigTable, Firebase, etc.

Hey, would love to contribute the DynamoDB checkpointer if it's still open. Is anyone already working on it?

Issue #593, opened 2024-10-14
`MemorySaver` ignores `newVersions` argument on `put` - not storing deltas


benjamincburns:
Found while working on #541. Per discussion on #577, `MemorySaver` isn't respecting the `newVersions` argument, and as a result it's storing the entire contents of `checkpoint.channel_values` on every call.

trollboss2572:
Good morning, we are a group of students at the University of Toronto currently investigating the issue mentioned and believe that we could help tackle this. Can we get this opened as an issue and assigned to me? @nfcampos

benjamincburns:
@trollboss2572 I discussed this briefly with @nfcampos the other day, and I think there's an open question as to whether or not this is a valid issue.

The `newVersions` argument on `put` can be optionally used by checkpointers to determine which channels have updated state since the previous checkpoint. This allows storage backends to avoid storing duplicate state for channels that haven't changed since the previous checkpoint. In the case of `MemorySaver`, there's a valid argument that this isn't a worthwhile optimization.

In my opinion, whether or not this is an issue boils down to whether the LangGraph team wants `MemorySaver` to serve as a reference implementation for checkpointer authors to follow. If so, then it would make sense to implement versioned channel storage, as otherwise `MemorySaver` would be an incomplete reference.

Furthermore, if the LangGraph team decide that `MemorySaver` should stay as-is, then the tests in `@langchain/langgraph-checkpoint-validation` that validate versioned channel storage should either be removed, or more ideally, moved to an "optional" suite that only runs when the user explicitly requests tests for optimizations.

trollboss2572:
@benjamincburns Thank you for the swift response to my message.

While I was researching the codebase I came to the same conclusion. This is more of an architectural issue than an implementation issue/feature/bug, so I believe this would be above my pay grade and would be more suited to the maintainers of langgraph to see what option they would prefer, as you mentioned. 

I read your other post and saw your recommendations for implementing CheckpointSaver for a different database framework, I believe that might be the best bet if we were to implement something. Thank you for your advice on that.

Thank you for your time and for being open to questions should they arise. If there is anything that can be done by volunteers regarding the architectural updates to MemorySaver though, let me know.

benjamincburns:
@trollboss2572 no worries, happy to help. Though just a reminder, I'm a volunteer just like you, so your guess is roughly as good as mine :-)

benjamincburns:
[...]

Issue #592, opened 2024-10-14
`SqliteSaver.put` stores config with undefined `checkpoint_ns` when `config.configurable.checkpoint_ns` is undefined.


benjamincburns:
Found while working on #541. Relevant code is [here](https://github.com/langchain-ai/langgraphjs/blob/d3c8204026c351672390d36ae652b7446b258e72/libs/checkpoint-sqlite/src/index.ts#L233) (value is unchecked).

benjamincburns:
Per https://github.com/langchain-ai/langgraphjs/pull/582#discussion_r1807012164 - it would also be nice if whomever did this included a change that migrated the type of the `metadata` column to `jsonb` rather than `blob` so that the query for `list` didn't need to reproject this column to `jsonb` in its `WHERE` clause (causes a linear scan).

PR #600, opened 2024-10-15
feat(checkpoint-validation): add custom checkpointer validation tool

### LINKED ISSUE #541: Feature request: Create an acceptance test suite for checkpointers

benjamincburns:
## What's in the box?

This change adds the package `@langgraph/checkpoint-validation-tool`, which provides a portable black-box acceptance test suite for validating both internal and external checkpoint saver implementations.

This change fixes #541.

## How can I use it?

It can be ran as a CLI tool, or by embedding it into an existing Jest test project. Interfacing between the test suite and custom checkpoint saver code is handled by a user-provided object that contains a set of simple functions that are used for infrastructure and component setup/teardown. Full usage details can be found in the `README.md` introduced in this PR.

## Why is the build failing?

This package is also configured to exercise the existing checkpoint saver implementations during normal test runs (via `yarn test`). I use [TestContainers](https://testcontainers.com/guides/getting-started-with-testcontainers-for-nodejs/) where needed (MongoDB and PostgreSQL savers) for infrastructure provisioning.

A number of problems were discovered in the existing LangGraph checkpoint saver implementations while working on this PR. I've submitted additional PRs to correct some of the problems that I discovered (#578, #580, #582, #583, #584), and raised GH issues for others (#581, #589, #590, #591, #592, #593, #594, #595).

In the case of the problems that are described in GH issues, I've included implementation-specific carve-outs in the test logic to avoid test failures. These carve-outs are marked with `TODO` comments in the code.

In the case of the problems for which I've submitted PRs, I have **not** included any carve-outs in this test suite for those. As a result, the tests introduced by this PR will fail until the above mentioned PRs are accepted.

## What's the best way to evaluate this whole mess?

To solve the chicken-and-egg problem that's caused by the dependence of this PR on so many others, I've created [a merge branch](https://github.com/benjamincburns/langgraphjs/tree/compiled-checkpoint-fixes) that contains all of those fixes, along with the changes in this PR. I will endeavor to keep that branch in sync as things change.


benjamincburns:
I should also note that I haven't added any documentation for this beyond what's in the `README.md`. I'd be happy to add docs, but I wasn't sure where they'd fit best. Guidance would be appreciated.

jacoblee93:
Thanks so much for this! Will dive in today.

libs/checkpoint-validation/src/tests/postgresInitializer.ts
```diff
@@ -0,0 +1,64 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { PostgresSaver } from "@langchain/langgraph-checkpoint-postgres";
+
+// eslint-disable-next-line import/no-extraneous-dependencies
+import {
+  PostgreSqlContainer,
+  type StartedPostgreSqlContainer,
+} from "@testcontainers/postgresql";
```
jacoblee93, line 8: Ah very nice didn't know this existed

libs/checkpoint-validation/README.md
```diff
@@ -0,0 +1,103 @@
+# @langchain/langgraph-checkpoint-validation
```
jacoblee93, line 1: We should add a section to the LangGraph docs around contributing! I can do that in a separate PR after we merge this.

libs/checkpoint-validation/src/cli.ts
```diff
@@ -0,0 +1,109 @@
...
+              `Initializer imported from '${initializerImportPath}' does not conform to the expected schema. Make sure " +
+              "it is the default export, and that implements the CheckpointSaverTestInitializer interface. Error: ${e}`
+            );
+            process.exit(1);
+          }
+
+          await runCLI(
+            {
+              _: [],
+              $0: "",
```
jacoblee93, line 97: Could we avoid the need for the global by passing the initializer filename here as an extra option? Not sure what the API looks like

jacoblee93 (commented):
Some nits but overall I think this looks excellent and well thought out. Will give it a spin a bit later and approve if I'm able to run everything!

Other general nit is that we've generally used snake case for filenames stylistically - would prefer to keep it that way. I can make that change later too though.

libs/checkpoint-validation/src/tests/mongoInitializer.ts
jacoblee93, line 1: nit: rename to `mongodb_initializer.ts`

libs/checkpoint-validation/src/spec/data.ts
jacoblee93, line 1: nit: move into `testUtils.ts` or make clearer that this doesn't contain any test files

libs/checkpoint-validation/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+### `afterAll`: Tear down required infrastructure
+
+If you set up infrastructure during the `beforeAll` step, you may need to tear it down once the tests complete their execution. You can define this teardown logic in the optional `afterAll` function. Much like `beforeAll` this function will execute exactly one time, after all tests have finished executing.
+
+**IMPORTANT**: If you kill the test runner early this function may not be called. To avoid manual clean-up, give preference to test infrastructure management tools like [TestContainers](https://testcontainers.com/guides/getting-started-with-testcontainers-for-nodejs/), as these tools are designed to detect when this happens and clean up after themselves once the controlling process dies.
+
+### (Required) `createSaver`: Construct your checkpoint saver
```
jacoblee93, line 43: uber nit: have been generally calling these `checkpointer` rather than `saver` for these

jacoblee93:
Just ran locally with `MemorySaver` and it seems great :)

If you don't have time for the style things I'd feel pretty comfortable merging as is/picking them up myself.

benjamincburns:
> Just ran locally with `MemorySaver` and it seems great :)

Glad to hear it! 😅

> If you don't have time for the style things I'd feel pretty comfortable merging as is/picking them up myself.

I'll look over the feedback a bit more closely tonight or tomorrow and let you know if I'll need to put any of it off. From a quick glance through I think the only thing that _might_ take a bit of time is the bit about avoiding passing via globals in the dynamic test gen stuff (`runner.ts`).

Apart from the feedback that you've given, I also have three lingering "nice to haves" that I didn't really have time to address:

- it'd be nicer if library users didn't need to be using Jest to integrate it directly with their existing test suite
- it'd also be nicer if the CLI exposed more of the Jest functionality to the user, for configuring things like code coverage, various reporters, etc.
- It's probably overkill, but it'd give me slightly more confidence if it also ran some e2e tests that exercised the savers at a higher level, perhaps using some code from the examples, etc.

If you agree that these would be useful improvements I can raise some issues for them so they aren't forgotten.

libs/checkpoint-validation/README.md
benjamincburns: Yeah, I wasn't sure what to go with there, so I defaulted to the term used in the type definitions. Easy to switch it out, though.

FWIW, I think it'd make the code a bit more searchable/readable if you were to rename the `XyzSaver` classes (including `BaseCheckpointSaver`) to `XyzCheckpointer` next time you had to make a breaking change in the core checkpoint library.

benjamincburns:
Pushed a couple quick fixes for test logic errors that I found while using this to validate my project's custom DynamoDB checkpointer today

- dropped a `put` test that was validating an invalid assumption that I made early on 😊
  - it caused my custom checkpointer to throw while `pending_sends` (parent not found)
  - may be worth adding this back in as a failure case, as it seems a bit off that the existing checkpointers allow this to pass silently?
- updated a `getTuple` failure case test to pass an invalid `thread_id`
  - without this it was essentially duplicating the test just below it

jacoblee93:
> it'd be nicer if library users didn't need to be using Jest to integrate it directly with their existing test suite

I think it's fine for now, it's up to you but would say wait for a few people to ask

> it'd also be nicer if the CLI exposed more of the Jest functionality to the user, for configuring things like code coverage, various reporters, etc.

Same as above

> It's probably overkill, but it'd give me slightly more confidence if it also ran some e2e tests that exercised the savers at a higher level, perhaps using some code from the examples, etc.

Yeah this would be nice, perhaps something in line with:

https://github.com/langchain-ai/langgraphjs/blob/main/libs/langgraph/src/tests/pregel.postgres_saver.int.test.ts

jacoblee93:
Seems like some flakiness with `testcontainers`? Not super familiar with it but am seeing this in CI:

```
  ● @langchain/langgraph-checkpoint-mongodb › @langchain/langgraph-checkpoint-mongodb#list › {
  description: "should return no tuples when thread_id does not match pushed checkpoint(s), checkpoint_ns matches 'child', limit is 2, before child checkpoint, and filter is undefined",
  thread_id: '1ef8e8ba-72cb-6ca0-fffd-bb7bbb3b207d',
  checkpoint_ns: 'child',
  limit: 2,
  before: [Object],
  filter: undefined,
  expectedCheckpointIds: []
}

    Could not find a working container runtime strategy

      22 |
      23 |   async beforeAll() {
    > 24 |     startedContainer = await container.start();
         |                        ^
      25 |     const connectionString = `mongodb://127.0.0.1:${startedContainer.getMappedPort(
      26 |       27017
      27 |     )}/${dbName}?directConnection=true`;

      at getContainerRuntimeClient (node_modules/testcontainers/src/container-runtime/clients/client.ts:64:9)
      at MongoDBContainer.start (node_modules/testcontainers/src/generic-container/generic-container.ts:81:20)
      at MongoDBContainer.start (node_modules/@testcontainers/mongodb/src/mongodb-container.ts:15:40)
      at Object.beforeAll (libs/checkpoint-validation/src/tests/mongoInitializer.ts:24:24)
      at Object._initializer_beforeAllTimeout (libs/checkpoint-validation/src/spec/index.ts:27:5)

```

benjamincburns:
> I can repro locally, not sure if it's working for you or if it needs additional setup?

It does work for me. Do you have docker installed and running when testing locally? If so, you may be encountering a timeout when pulling the image due to the local `yarn test` applying a 30s timeout via the CLI.

In the GHA log it's failing on the mac runner with the message `Could not find a working container runtime strategy`. It looks like this is because Docker doesn't ship on the macOS images by default (per [this comment](https://github.com/actions/runner-images/issues/2150#issuecomment-737125967)). You'll notice that it did pass on the Ubuntu image, however.

I'll look into a better solution tonight. In the thread linked above, they did pre-install `colima`, so if you don't mind the mac build taking an extra minute or two to run, the solution may just be to add `brew install docker && colima start` to the GHA test run config. That'll install the docker CLI for use with the colima runtime. Otherwise we can ignore these tests in CI when running on mac, mark them as integration tests, or some combination of these options.

benjamincburns:
Colima isn't supported on M-series macs in GHA due to a lack of support for nested virtualization support (https://github.com/actions/runner-images/issues/9460#issuecomment-1981203045), so I ignored those tests in CI when running on M-series macs.

Unfortunately it seems that they also fail on Windows, likely due to docker not being installed for licensing reasons. I'll see if there's an alternative container runtime available there. If not, I'll also ignore these in CI on Windows.

libs/checkpoint-validation/src/spec/data.ts
```diff
@@ -0,0 +1,290 @@
...
+        source: "loop",
+        step: 0,
+        writes: {
+          // I think this is meant to be opaque to checkpoint savers, so I'm just stuffing the data in here to make sure it's stored and retrieved
+          someNode: parentPendingWrites,
+        },
+        parents: {
+          // I think this is meant to be opaque to checkpoint savers, so I'm just stuffing the data in here to make sure it's stored and retrieved
+          // I think this is roughly what it'd look like if it were generated by the pregel loop, though
+          checkpoint_ns: parentCheckpointId,
```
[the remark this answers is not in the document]
benjamincburns: This was actually a syntax goof on my part. I meant to write `[checkpoint_ns]: parentCheckpointId` here, as I was basing this off of the TSDoc comment on `CheckpointMetadata.parents` which reads:

```ts
export interface CheckpointMetadata {
   ...
    /**
     * The IDs of the parent checkpoints.
     * Mapping from checkpoint namespace to checkpoint ID.
     */
    parents: Record<string, string>;
}
```

libs/checkpoint-validation/src/spec/getTuple.ts
```diff
@@ -0,0 +1,265 @@
...
+  initializer: CheckpointSaverTestInitializer<T>
+) {
+  describe(`${initializer.saverName}#getTuple`, () => {
+    let saver: T;
+    let initializerConfig: RunnableConfig;
+    beforeAll(async () => {
+      const baseConfig = {
+        configurable: {},
+      };
+      initializerConfig = mergeConfigs(
```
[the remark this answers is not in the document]
benjamincburns: I think actually I need to just remove this (along with the `configure` method on `CheckpointSaverTestInitializer`).

My initial thinking was that it'd be handy to pass custom config into the checkpointer, e.g. like auth details that can be used for multi-tenant isolation. Unfortunately it doesn't seem that the core langgraph code merges the `RunnableConfig` passed on `invoke` when calling the checkpointer, so this logic doesn't hold any value in the tests.

libs/checkpoint-validation/src/cli.ts
benjamincburns: I couldn't find any nice way to do this. If Jest had proper support for ESMs (and therefore allowed top-level `await` in test files), I could've made it much cleaner.

FWIW, `vitest` does support this, and [migration from Jest](https://vitest.dev/guide/migration#migrating-from-jest) isn't usually much of a drama. Definitely not advocating for that as a result of this PR, but it might be worth keeping it in mind if you hit more of these sort of issues in the future.

libs/checkpoint-validation/src/tests/mongoInitializer.ts
benjamincburns: done, and applied this naming convention for all other files.

libs/checkpoint-validation/src/spec/data.ts
benjamincburns: done.

libs/checkpoint-validation/README.md
benjamincburns: updated all names/references/readme text/etc to reflect this naming convention.

benjamincburns:
Current HEAD (0ea8824) [passes CI](https://github.com/benjamincburns/langgraphjs/actions/runs/11433505024). AFAICT this should be ready for re-review.

> > it'd be nicer if library users didn't need to be using Jest to integrate it directly with their existing test suite
> 
> I think it's fine for now, it's up to you but would say wait for a few people to ask

I did wind up taking a small/simple step in this direction by dropping imports from `@jest/globals` and using the `types` TS compiler option. That should make it work with any test lib that injects compatible globals.

libs/checkpoint-validation/src/spec/get_tuple.ts
```diff
@@ -0,0 +1,249 @@
...
+      let latestTuple: CheckpointTuple | undefined;
+
+      beforeAll(async () => {
+        thread_id = uuid6(-3);
+        parentCheckpointId = uuid6(-3);
+        childCheckpointId = uuid6(-3);
+
+        const writesToParent = [
+          {
+            taskId: "pending_sends_task",
```
nfcampos, line 48: I think the library always uses uuids for taskId, are we using something else here on purpose to test it supports arbitrary strings?

nfcampos (approved):
This looks good, great work!

PR #618, opened 2024-10-22
fix(checkpoint-validation): fix broken initial release

benjamincburns:
Unfortunately the initial release of `@langchain/langgraph-checkpoint-validator` is broken for two reasons.

1. A bug in `@swc-node/register` attempts to resolve SWC packages from the CWD rather than the installed package's `node_modules` directory.
2. The jest config used by the CLI wasn't included in the release because I forgot to add it to the `files` array in `package.json`.

I fixed the first issue by applying the workaround mentioned here: https://github.com/swc-project/swc-node/pull/805#issuecomment-2257146304.

I fixed the second issue by moving the runtime jest config into the `src` dir and making it a TS file.

I tested this locally using `npm link` and it seems to fix both issues.

Sorry for the hiccup!

### DIFF
--- a/libs/checkpoint-validation/bin/cli.js
+++ b/libs/checkpoint-validation/bin/cli.js
@@ -1,5 +1,7 @@
 #!/usr/bin/env node
-
+import { register } from "node:module";
 import { main } from "../dist/cli.js";
+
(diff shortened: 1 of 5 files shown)


jacoblee93:
Thank you!

PR #620, opened 2024-10-22
feat(langgraph): add subgraph add node option to explicitly specify subgraphs

dqbd:
Unlike in Python, we need to specify the subgraphs beforehand when adding nodes with lambdas that invoke the subgraph. 

This is necessary to ensure that:
1. Nested interrupts are properly handled
2. Debug stream events include checkpoint metadata
3. Both the graph and the subgraph are introspectable

### DIFF
--- a/libs/langgraph/src/graph/state.ts
+++ b/libs/langgraph/src/graph/state.ts
@@ -48,6 +48,7 @@ import {
 import type { RetryPolicy } from "../pregel/utils/index.js";
 import { isConfiguredManagedValue, ManagedValueSpec } from "../managed/base.js";
 import type { LangGraphRunnableConfig } from "../pregel/runnable_types.js";
+import { isPregelLike } from "../pregel/utils/subgraph.js";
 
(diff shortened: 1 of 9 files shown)


libs/langgraph/src/graph/state.ts
```diff
@@ -314,11 +315,17 @@ export class StateGraph<
...
+    const runnable = _coerceToRunnable(action) as unknown as Runnable<S, U>;
     const nodeSpec: StateGraphNodeSpec<S, U> = {
-      runnable: _coerceToRunnable(action) as unknown as Runnable<S, U>,
+      runnable,
       retryPolicy: options?.retryPolicy,
       metadata: options?.metadata,
       input: options?.input?.spec ?? this._schemaDefinition,
+      subgraphs: isPregelLike(runnable)
+        ? // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          [runnable as any]
```
jacoblee93, line 327: Feels like this should be caught by the type guard?

PR #624, opened 2024-10-23
fix(checkpoint-validation): cli fixes

benjamincburns:
- move ts-jest to `dependencies` from `devDependencies`
- get rid of `@swc-node/register` in favor of `@swc/jest`
- handle relative path to initializer import without leading dot
- configure CLI to run with ESM support in jest, enabling top-level await (gets rid of the ugly global arg passing)
- pass jest config directly to runCLI, no need for extra jest config file

Tested on mac as global install by publishing to a local verdaccio registry


nfcampos (approved):

nfcampos (approved):

nfcampos:
Thanks!

PR #625, opened 2024-10-23
fix(checkpoint-mongodb): fix filter support, store state deltas, populate pendingWrites

### LINKED ISSUE #581: Support for options.filter is broken on MongoDBSaver.list and can't be fixed without a breaking change
For #541 I decided the best course of action for testing CheckpointSaver `list` implementations would be to write [a combinatorial test](https://github.com/benjamincburns/langgraphjs/blob/9d78f0735c6632475ed8c80cfb511af5d5b69681/libs/checkpoint-validation/src/tests/spec/list.ts#L88-L143) that spams `list` with [a bunch of different argument combinations](https://github.com/benjamincburns/langgraphjs/blob/9d78f0735c6632475ed8c80cfb511af5d5b69681/libs/checkpoint-validation/src/tests/spec/list.ts#L225-L275). For `MongoDBSaver`, every test that involved specifying values under `options.filter` failed.

The `options.filter` argument appears to be meant to constrain the result based on values in the stored `CheckpointMetadata`. The `MongoDBSaver` attempts to do this [here](https://github.com/langchain-ai/langgraphjs/blob/45e59e64175a89a7cf782015b011b7d348ef835f/libs/checkpoint-mongodb/src/index.ts#L148-L152), with the following code:

```ts
    if (filter) {
      Object.entries(filter).forEach(([key, value]) => {
        query[`metadata.${key}`] = value;
      });
    }
```

Unfortunately because of [how `checkpoint` and `metadata` are serialized in the `put` method(https://github.com/langchain-ai/langgraphjs/blob/45e59e64175a89a7cf782015b011b7d348ef835f/libs/checkpoint-mongodb/src/index.ts#L219-L228) (the output of `this.serde.dumpsTyped` is of type `[string, UInt8Array]`), `metadata` is stored as a BSON Binary field, so any time `options.filter` is passed in, `list` yields no results.

At first glance you might think that this would be fixable without a breaking change by applying the `filter` in the client (in the `for (const doc of result)` loop). That would work, but only if you also applied the `limit` constraint client-side as well. Otherwise any time a `limit` constraint is applied, you'll likely miss items that would've matched your filter.

The same problem exists on the `SqliteSaver` as well, however that's solvable without a breaking change by joining the query with [a CTE that deserializes the stored `metadata` object](https://github.com/benjamincburns/langgraphjs/blob/9d78f0735c6632475ed8c80cfb511af5d5b69681/libs/checkpoint-sqlite/src/index.ts#L175-L182) (will PR that change tomorrow). I'm not a MongoDB aficionado, but I don't think it has facilities for transforming documents in any way prior to applying filters.

The simple fix is to persist the serialized fields as objects nested into the document object, rather than byte arrays. Unfortunately, doing this without a migration will break existing collections of checkpoint history.

At present there's no mechanism baked into either the `langgraph-checkpoint` library or the `MongoDBSaver` to check for out of date document structures, or to define or kick off migrations like these, so the design of this will likely require a bit of care. It would be easy enough to write an automatic migration that runs on start, but that gets ugly in production environments when there are multiple concurrent processes. As a result, I'd recommend making it so the updated `MongoDBSaver` fails on initialization if the migration is required for an existing db. That way the migration can be applied out of band, and there's no risk that someone will blow away their checkpoint storage as a result of an `npm update`.

For now I've pruned the `filter` checks in [my `list` tests](https://github.com/benjamincburns/langgraphjs/blob/9d78f0735c6632475ed8c80cfb511af5d5b69681/libs/checkpoint-validation/src/tests/spec/list.ts#L217-L221) in order to prevent this from blocking progress on #541.

### LINKED ISSUE #589: `MongoDBSaver.list` returns `CheckpointTuples` without the `pendingWrites` field populated.
Found while working on #541. Relevant code is [here](https://github.com/langchain-ai/langgraphjs/blob/d3c8204026c351672390d36ae652b7446b258e72/libs/checkpoint-mongodb/src/index.ts#L177-L196).

### LINKED ISSUE #595: `MongoDbSaver` ignores `newVersions` argument on `put` - not storing deltas

benjamincburns:
Breaking changes:
- Changes the format of stored documents in mongo, requiring existing databases to be migrated by calling the `MongoDBSaver.migrate` method.

Non-breaking changes:
- Adds filter logic to the `list` method to support the `options.filter` argument, fixes #581
  - Adds basic document format version tagging and migration support
  - Migrates existing stored metadata to stored as BSON objects rather than binary data.
  - Includes a change to `checkpoint-sqlite` and `checkpoint` libraries to centralize an array that I'm using in multiple packages to sanitize filter inputs.
- Stores `channel_values` by version on change, respecting the `newVersions` argument to `put`, fixes #595 
- Populates `pendingWrites` and `pending_sends` correctly, fixes #589
- Updates `checkpoint-validation` to no longer skip tests that exercise the above fixes

### DIFF
--- a/libs/checkpoint/src/types.ts
+++ b/libs/checkpoint/src/types.ts
@@ -36,3 +36,26 @@ export interface CheckpointMetadata {
    */
   parents: Record<string, string>;
 }
+
+const checkpointMetadataKeys = ["source", "step", "writes", "parents"] as const;
+
+type CheckKeys<T, K extends readonly (keyof T)[]> = [K[number]] extends [
+  keyof T
+]
+  ? [keyof T] extends [K[number]]
+    ? K
+    : never
+  : never;
+
+function validateKeys<T, K extends readonly (keyof T)[]>(
+  keys: CheckKeys<T, K>
+): K {
+  return keys;
+}
+
+// Used by checkpoint list methods to sanitize the `options.filter` argument. If this line fails to compile, update
+// `checkpointMetadataKeys` to contain all the keys in `CheckpointMetadata`.
+export const validCheckpointMetadataKeys = validateKeys<
+  CheckpointMetadata,
+  typeof checkpointMetadataKeys
+>(checkpointMetadataKeys);
--- a/libs/checkpoint-mongodb/package.json
+++ b/libs/checkpoint-mongodb/package.json
@@ -43,6 +43,7 @@
     "@langchain/scripts": ">=0.1.3 <0.2.0",
     "@swc/core": "^1.3.90",
     "@swc/jest": "^0.2.29",
+    "@testcontainers/mongodb": "^10.13.2",
     "@tsconfig/recommended": "^1.0.3",
     "@types/better-sqlite3": "^7.6.9",
     "@types/uuid": "^10",
@@ -62,6 +63,7 @@
     "prettier": "^2.8.3",
     "release-it": "^17.6.0",
     "rollup": "^4.23.0",
+    "testcontainers": "^10.13.2",
     "ts-jest": "^29.1.0",
     "tsx": "^4.7.0",
     "typescript": "^4.9.5 || ^5.4.5"
--- a/libs/checkpoint-mongodb/src/index.ts
+++ b/libs/checkpoint-mongodb/src/index.ts
@@ -9,13 +9,21 @@ import {
   type PendingWrite,
   type CheckpointMetadata,
   CheckpointPendingWrite,
+  validCheckpointMetadataKeys,
 } from "@langchain/langgraph-checkpoint";
+import { applyMigrations, needsMigration } from "./migrations/index.js";
+
+export * from "./migrations/index.js";
+
+// increment this whenever the structure of the database changes in a way that would require a migration
+const CURRENT_SCHEMA_VERSION = 1;
 
 export type MongoDBSaverParams = {
   client: MongoClient;
   dbName?: string;
   checkpointCollectionName?: string;
   checkpointWritesCollectionName?: string;
+  schemaVersionCollectionName?: string;
 };
 
 /**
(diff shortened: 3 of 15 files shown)


benjamincburns:
Going to close this for now and reopen it as a fix for both #581 and #595. Both of those issues will require migrations, and it'll be faster if they're applied together.

I may also pick up #589 while I'm at it, but I'll be sure to include that as a separate commit in case you prefer to review it separately.

benjamincburns:
Ah, sorry I think I spoke too soon. I can handle #595 easily enough without a migration. I'll submit a new PR for it done that way, and if on review you prefer a migration, it'll be easy enough to add one once this PR merges.

libs/checkpoint/src/types.ts
```diff
@@ -36,3 +36,26 @@ export interface CheckpointMetadata {
    */
   parents: Record<string, string>;
 }
+
+const checkpointMetadataKeys = ["source", "step", "writes", "parents"] as const;
```
nfcampos, line 40: I think we might accept arbitrary metadata keys?

benjamincburns: I was just taking cues from `PostgresSaver` on this. Is there some mechanism for user code to modify the checkpoint metadata? It doesn't have an index signature, so I (perhaps incorrectly) assumed it was a complete type that originated internally. The typing cues are contradictory though, as `options.filter` is `Record<string, unknown>`. A type of `Partial<CheckpointMetadata>` would've been a clearer expression of intent there if it was indeed meant to be restricted.

Regardless, I added it in the RDBMS implementations because the list of filter keys is being directly concatenated into the query text (SQL injection potential), and as a result I also added it here for consistency.

Assuming arbitrary filter keys _aren't_ desired, perhaps it would be better to throw on unknown keys. Silently dropping them could be confusing behavior, otherwise.

If they _are_ desired, we should think carefully about how to sanitize the inputs for the RDBMS implementations.

benjamincburns:
> The ESLint failure is a false positive and it should be fixed by #633. It seems transient, and sometimes occurs in `main`.

Just rebased on current `main`, which should remove the ESLint violation.

christian-bromann:
Closing due to inactivity.

PR #629, opened 2024-10-24
fix(checkpoint): don't throw on MemorySaver.put if no checkpoint_ns

### LINKED ISSUE #591: `MemorySaver.put` throws when `config.configurable.checkpoint_ns` not present, rather than defaulting to empty namespace.
Found while working on #541. Relevant code is [here](https://github.com/langchain-ai/langgraphjs/blob/d3c8204026c351672390d36ae652b7446b258e72/libs/checkpoint/src/memory.ts#L284-L288). This behaviour is inconsistent with the other checkpoint savers, which default to the default namespace when one isn't supplied.

benjamincburns:
For consistency with the other checkpointers, this change makes `MemorySaver.put` use the default namespace when `config.configurable.checkpoint_ns` is undefined.

Also unignores relevant tests in `checkpoint-validation` for `MemorySaver` runs.

Fixes #591

### DIFF
--- a/libs/checkpoint-validation/src/spec/put.ts
+++ b/libs/checkpoint-validation/src/spec/put.ts
@@ -178,9 +178,6 @@ export function putTests<T extends BaseCheckpointSaver>(
     });
 
     it_skipForSomeModules(initializer.checkpointerName, {
-      // TODO: MemorySaver throws instead of defaulting to empty namespace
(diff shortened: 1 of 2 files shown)


nfcampos (approved):

nfcampos:
Thanks!

PR #634, opened 2024-10-25
fix(checkpoint-sqlite): don't return undefined namespace from put

### LINKED ISSUE #592: `SqliteSaver.put` stores config with undefined `checkpoint_ns` when `config.configurable.checkpoint_ns` is undefined.

benjamincburns:
fixes #592


jacoblee93 (dismissed):
Thank you!

jacoblee93 (approved):

PR #636, opened 2024-10-27
feat: vercel kv checkpointer

chentschel:
Vercel KV checkpointer . Contribution by `https://x.com/chentschel_`


libs/checkpoint-vercel-kv/src/index.ts
```diff
@@ -0,0 +1,389 @@
...
+            ),
+          },
+        };
+      },
+      {}
+    );
+    await this.kv.mset(values);
+  }
+
+  async flush(): Promise<void> {
```
jacoblee93, line 386: What does this do?

libs/checkpoint-vercel-kv/src/index.ts
```diff
@@ -0,0 +1,389 @@
...
+      checkpoint: JSON.parse(
+        new TextDecoder().decode(checkpointValue).replace(/\0/g, "")
+      ),
+      metadata: JSON.parse(
+        new TextDecoder().decode(metadataValue).replace(/\0/g, "")
+      ),
+      parent_checkpoint_id: checkpoint_id,
+    };
+
+    // LUA script to set checkpoint data atomically"
```
jacoblee93, line 319: nit: remove quote at the end

libs/checkpoint-vercel-kv/src/index.ts
```diff
@@ -0,0 +1,389 @@
...
+
+        const [checkpoint, metadata] = await Promise.all([
+          checkpointPromise as Checkpoint,
+          metadataPromise as CheckpointMetadata,
+        ]);
+
+        // filter by metadata
+        if (filter && Object.keys(filter).length > 0) {
+          const matches = Object.keys(filter).every(
+            // @ts-ignore-next-line
```
jacoblee93, line 243: Prefer cast

libs/checkpoint-vercel-kv/src/index.ts
```diff
@@ -0,0 +1,389 @@
...
+    const filteredKeys = keys.filter((key: string) => {
+      const [, , checkpoint_id] = key.split(":");
+      return !before || checkpoint_id < before?.configurable?.checkpoint_id;
+    });
+
+    const sortedKeys = filteredKeys.sort((a: string, b: string) =>
+      b.localeCompare(a)
+    );
+
+    const rows: (KVRow | null)[] = await this.kv.mget(...sortedKeys);
```
jacoblee93, line 223: This should get only up to `limit` keys at a time and loop?

jacoblee93 (commented):
Thanks for this! See comments

libs/checkpoint-vercel-kv/src/index.ts
chentschel: Yes. Takes the matching keys, but to support metadata filtering, we its doing so when it gets the value in filtering and count in the loop / generator. I'd be better perhaps to change the LUA script to an HSCAN and filter the metadata there.

libs/checkpoint-vercel-kv/src/index.ts
chentschel: Cleans the store. I'll remove this

benjamincburns:
Hi @chentschel,

Thanks so much for your contribution! We really appreciate the time and effort you've put into building and sharing this.

Unfortunately we've decided that we can't accept any more custom checkpointer integrations into LangGraph itself. We want to keep the library’s maintenance footprint manageable. We just don’t have the resources or expertise to support a wide variety of storage backends, and we feel that trying to do so would be a disservice to the community.

Instead of publishing potentially low-quality official integration packages that we can't maintain ourselves, we want to enable experts in the community to publish and maintain their own high-quality integrations. To help with that we'll soon be setting up a community-driven docs page where folks can share links to their own checkpointer packages. That’s not live yet, but we hope to have it up soon. When it is, I'll be sure to leave a comment here so you can add yours!

Again, we really appreciate your work, and we’re excited to see the ecosystem grow. Thanks so much for being a part of the LangGraph community! 🚀

PR #639, opened 2024-10-29
feat(langgraph): Adds RemoteGraph

jacoblee93:
Tests to follow - `getGraph()` and `getSubgraphs()` are unfortunately sync right now, so made new, bespoke async methods for now.

SDK typing also seems to be off in a few cases, will dig in further once tests are finished.

### DIFF
--- a/libs/langgraph/package.json
+++ b/libs/langgraph/package.json
@@ -32,13 +32,13 @@
   "license": "MIT",
   "dependencies": {
     "@langchain/langgraph-checkpoint": "~0.0.10",
+    "@langchain/langgraph-sdk": "~0.0.20",
     "double-ended-queue": "^2.1.0-0",
     "uuid": "^10.0.0",
     "zod": "^3.23.8"
   },
   "peerDependencies": {
-    "@langchain/core": ">=0.2.36 <0.3.0 || >=0.3.9 < 0.4.0",
-    "@langchain/langgraph-sdk": "*"
+    "@langchain/core": ">=0.2.36 <0.3.0 || >=0.3.9 < 0.4.0"
   },
   "devDependencies": {
     "@jest/globals": "^29.5.0",
@@ -47,7 +47,6 @@
     "@langchain/core": "^0.3.15",
     "@langchain/langgraph-checkpoint-postgres": "workspace:*",
     "@langchain/langgraph-checkpoint-sqlite": "workspace:*",
-    "@langchain/langgraph-sdk": "^0.0.18",
     "@langchain/openai": "^0.3.11",
     "@langchain/scripts": ">=0.1.3 <0.2.0",
     "@swc/core": "^1.3.90",
@@ -80,11 +79,6 @@
     "tsx": "^4.7.0",
     "typescript": "^4.9.5 || ^5.4.5",
     "zod-to-json-schema": "^3.22.4"
-  },
-  "peerDependenciesMeta": {
-    "@langchain/langgraph-sdk": {
-      "optional": true
-    }
   },
   "publishConfig": {
     "access": "public",
--- a/libs/langgraph/package.json
+++ b/libs/langgraph/package.json
@@ -37,7 +37,8 @@
     "zod": "^3.23.8"
   },
   "peerDependencies": {
-    "@langchain/core": ">=0.2.36 <0.3.0 || >=0.3.9 < 0.4.0"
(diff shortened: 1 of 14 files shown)


libs/langgraph/src/pregel/remote.ts
```diff
@@ -0,0 +1,446 @@
...
+          return obj.map((v) => sanitizeObj(v));
+        } else {
+          return Object.fromEntries(
+            Object.entries(obj).map(([k, v]) => [k, sanitizeObj(v)])
+          );
+        }
+      }
+
+      try {
+        JSON.stringify(obj);
```
dqbd, line 121: _question_: so the behavior is to bail out when we encounter a circular dependency and return `null` instead? Is this correct behaviour?

libs/langgraph/src/pregel/remote.ts
```diff
@@ -0,0 +1,446 @@
...
+  ): Promise<StateSnapshot> {
+    const mergedConfig = mergeConfigs(this.config, config);
+
+    const state = await this.client.threads.getState(
+      mergedConfig.configurable?.thread_id,
+      {
+        checkpoint: this._getCheckpoint(mergedConfig),
+        subgraphs: options?.subgraphs,
+      } as any
+    );
```
dqbd, line 394: This doesn't seem to be right:

```suggestion
      this._getCheckpoint(mergedConfig),
      options
```

libs/langgraph/src/pregel/remote.ts
```diff
@@ -0,0 +1,446 @@
...
+    for (const state of states) {
+      yield this._createStateSnapshot(state);
+    }
+  }
+
+  protected _getDrawableNodes(
+    nodes: Array<{
+      id: string | number;
+      name?: string;
```
dqbd, line 364: Name is not returned I think, it would be present in `data` instead

libs/langgraph/src/pregel/remote.ts
```diff
@@ -0,0 +1,446 @@
...
+      data?: Record<string, unknown>;
+      metadata?: unknown;
```
dqbd, line 366: Also not sure about the `metadata` if it's returned either

libs/langgraph/package.json
```diff
@@ -37,7 +37,8 @@
     "zod": "^3.23.8"
   },
   "peerDependencies": {
-    "@langchain/core": ">=0.2.36 <0.3.0 || >=0.3.9 < 0.4.0"
+    "@langchain/core": ">=0.2.36 <0.3.0 || >=0.3.9 < 0.4.0",
+    "@langchain/langgraph-sdk": "*"
```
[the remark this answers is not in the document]
jacoblee93: Yeah maybe. We could also make it an actual hard dep, which I think I'd prefer now that I think about it

libs/langgraph/src/pregel/remote.ts
jacoblee93: We could do the same thing as in `langsmith` if we wanted

jacoblee93: I will add a TODO

jacoblee93: Actually no I think this is correct

libs/langgraph/src/pregel/remote.ts
jacoblee93: Thanks, committing separately

libs/langgraph/src/graph/graph.ts
```diff
@@ -540,6 +544,187 @@ export class CompiledGraph<
...
+        },
+        START
+      ),
+    };
+    const endNodes: Record<string, DrawableGraphNode> = {};
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    let subgraphs: Record<string, CompiledGraph<any>> = {};
+    if (xray) {
+      subgraphs = Object.fromEntries(
+        (await gatherIterator(this.getSubgraphsAsync())).filter(
```
nfcampos, line 565: is gather iterator just Array.from?

jacoblee93: `Array.from` doesn't work with async iterators

There is this but Node 22 only:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fromAsync

PR #650, opened 2024-11-01
feat(graph): passthrough input types to invoke/stream

### LINKED ISSUE #628: Typesafe graph.invoke?
As far as I can tell, is there not a way to make the invoke method typesafe with the options argument `await graph.invoke({ key1: value })`?

## Expected Output
```ts
const StateAnnotation = Annotation.Root({
	messages: Annotation<string[]>({
		reducer: (x, y) => x.concat(y),
	}),
})

// Define a new graph
const workflow = new StateGraph(StateAnnotation)

const app = workflow.compile()

const result = await app.invoke({ keyDoesNotExist: 'hi' }) // type error here
```

dqbd:
Opting to fix output types in a different PR due to the complexity of `streamMode` 

Closes #628

### DIFF
--- a/libs/langgraph/src/graph/graph.ts
+++ b/libs/langgraph/src/graph/graph.ts
@@ -456,6 +456,7 @@ export class CompiledGraph<
 > {
   declare NodeType: N;
 
+  // TODO: deprecate these values after updating the API
   declare RunInput: State;
 
   declare RunOutput: Update;
--- a/libs/langgraph/src/graph/graph.ts
+++ b/libs/langgraph/src/graph/graph.ts
@@ -450,7 +450,9 @@ export class CompiledGraph<
   Record<N | typeof START, PregelNode<RunInput, RunOutput>>,
   Record<N | typeof START | typeof END | string, BaseChannel>,
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  ConfigurableFieldType & Record<string, any>
+  ConfigurableFieldType & Record<string, any>,
+  RunInput,
+  RunOutput
 > {
   declare NodeType: N;
 
--- a/libs/langgraph/src/graph/graph.ts
+++ b/libs/langgraph/src/graph/graph.ts
@@ -456,7 +456,6 @@ export class CompiledGraph<
 > {
   declare NodeType: N;
 
-  // TODO: deprecate these values after updating the API
   declare RunInput: State;
 
   declare RunOutput: Update;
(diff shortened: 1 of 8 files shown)


nfcampos (dismissed):

libs/langgraph/src/graph/graph.ts
```diff
@@ -441,37 +441,39 @@ export class Graph<
...
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  ConfigurableFieldType & Record<string, any>
+  ConfigurableFieldType & Record<string, any>,
+  Update,
+  State
 > {
   declare NodeType: N;
 
-  declare RunInput: RunInput;
+  declare RunInput: State;
```
dqbd, line 459: It's a little strange that `RunInput = StateType<SD>`, I think we should change it to Update?

libs/langgraph/src/prebuilt/react_agent_executor.ts
```diff
@@ -160,9 +150,7 @@ export function createReactAgent(
     return { messages: [await modelRunnable.invoke(messages, config)] };
   };
 
-  const workflow = new StateGraph<AgentState>({
-    channels: schema,
-  })
+  const workflow = new StateGraph(MessagesAnnotation)
```
dqbd, line 153: This is done primarily to get the correct input types (based off the `messagesReducer`, but open to different solutions.

libs/langgraph/src/tests/pregel.test.ts
```diff
@@ -3541,6 +3541,7 @@ export function runPregelTests(
           hello: "there",
           bye: "world",
           messages: ["hello"],
+          // @ts-expect-error This should emit a TS error
```
dqbd, line 3544: We can now actually catch excessive keys!

libs/langgraph/src/prebuilt/react_agent_executor.ts
nfcampos: I think it's good to make this change, the other syntax is deprecated

libs/langgraph/src/tests/pregel.test.ts
nfcampos: ha nice!

libs/langgraph/src/graph/graph.ts
nfcampos: Yea I agree, this looks to be the other way around, no? input should be update, output should be based on stream mode (sometimes update, sometimes state, etc)

nfcampos (dismissed):

PR #665, opened 2024-11-10
fix(checkpoint-mongodb): fix utf8 encoding support

mosluce:
Issue
The MongoDB checkpoint storage implementation currently does not properly handle UTF-8 encoded input (such as Chinese characters). When retrieving checkpoints via getTuple, message contents containing non-ASCII characters are corrupted and displayed as garbled text.
Changes

Added proper UTF-8 encoding support for MongoDB checkpoint storage
Ensures correct storage and retrieval of non-ASCII characters in checkpoint messages
Maintains data integrity for international character sets

Impact
This fix enables LangGraphJS to properly handle international text in conversations and workflows when using MongoDB as the checkpoint storage backend. Users working with non-English languages will now see their content preserved correctly throughout the graph execution process.

### DIFF
--- a/libs/checkpoint-mongodb/src/index.ts
+++ b/libs/checkpoint-mongodb/src/index.ts
@@ -87,7 +87,7 @@ export class MongoDBSaver extends BaseCheckpointSaver {
     };
     const checkpoint = (await this.serde.loadsTyped(
       doc.type,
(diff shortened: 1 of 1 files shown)


benjamincburns:
Thanks for this contribution, @mosluce!

PR #670, opened 2024-11-12
docs: Add breadcrumbs

jacoblee93:


docs/overrides/main.html
```diff
@@ -34,6 +34,10 @@
       color: #1E88E5;
     }
 
+    .md-sidebar {
+      display: none;
+    }
+
```
vbarda, line 40: need to also add this https://github.com/langchain-ai/langgraph/blob/main/docs/overrides/main.html#L41

vbarda: and we need to enable on reference page like here https://raw.githubusercontent.com/langchain-ai/langgraph/refs/heads/main/docs/docs/reference/index.md

jacoblee93: Reference pages work a bit differently in JS

PR #688, opened 2024-11-24
fix(docs): added missing dependency in tutorial by adding @langchain/langgraph

greysonthao:
While following the LangGraph tutorial, I encountered an issue where the tutorial was missing the @langchain/langgraph dependency. After adding this dependency, the errors were resolved. This PR updates the tutorial documentation to include the missing dependency installation step to help prevent others from encountering the same issue.

My username on X.com is @greyson_codes


jacoblee93:
Thanks!

PR #690, opened 2024-11-26
Implement interrupt(...) and Command({resume: ...})

nfcampos:
Fixes # (issue)

### DIFF
--- a/libs/langgraph/src/errors.ts
+++ b/libs/langgraph/src/errors.ts
@@ -18,6 +18,12 @@ export class BaseLangGraphError extends Error {
   }
 }
 
+export class GraphBubbleUp extends BaseLangGraphError {
+  get is_bubble_up() {
+    return true;
+  }
+}
+
 export class GraphRecursionError extends BaseLangGraphError {
   constructor(message?: string, fields?: BaseLangGraphErrorFields) {
     super(message, fields);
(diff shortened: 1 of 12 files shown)


libs/langgraph/src/errors.ts
```diff
@@ -18,6 +18,12 @@ export class BaseLangGraphError extends Error {
   }
 }
 
+export class GraphBubbleUp extends BaseLangGraphError {
+  get is_bubble_up() {
```
bracesproul, line 22: nit, camel case?

libs/langgraph/src/pregel/loop.ts
```diff
@@ -71,14 +74,13 @@ import { LangGraphRunnableConfig } from "./runnable_types.js";
...
 const DEFAULT_LOOP_LIMIT = 25;
-const SPECIAL_CHANNELS = [ERROR, INTERRUPT];
 
 // [namespace, streamMode, payload]
 export type StreamChunk = [string[], StreamMode, unknown];
 
 export type PregelLoopInitializeParams = {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  input?: any;
+  input?: any | Command;
```
bracesproul, line 83: Will this even give devs autocomplete since `any` is allowed here? If not, I think we should drop `Command` since `any` will cover it

bracesproul: oh wait this isnt externally facing. I think just drop then

bracesproul (dismissed):
lgtm, just had a couple of nits

libs/langgraph/src/errors.ts
nfcampos: Ah I was following the names you have elsewhere in this file?

PR #703, opened 2024-12-05
fix(interrupt): match signature of GraphInterrupt created in interrupt()

dqbd:
Fixes # (issue)


jacoblee93 (approved):
Thanks!

nfcampos (approved):

libs/langgraph/src/pregel/messages.ts
```diff
@@ -143,7 +143,7 @@ export class StreamMessagesHandler extends BaseCallbackHandler {
       (tags === undefined || !tags.includes(TAG_HIDDEN))
     ) {
       this.metadatas[runId] = [
-        (metadata.langgraph_checkpoint_ns as string).split("NS_SEP"),
+        (metadata.langgraph_checkpoint_ns as string).split("|"),
```
jacoblee93, line 146: This is available as `CHECKPOINT_NAMESPACE_SEPARATOR` btw

PR #707, opened 2024-12-05
fix(langgraph): Prevent non RunnableConfig keys from being passed

bracesproul:
Depends on https://github.com/langchain-ai/langchainjs/pull/7324 merging & releasing. Then, bump the min dep requirement


libs/langgraph/src/pregel/algo.ts
```diff
@@ -1,8 +1,8 @@
 /* eslint-disable no-param-reassign */
 import {
-  mergeConfigs,
```
nfcampos, line 3: nit: should we return this to what it was to avoid diff

PR #712, opened 2024-12-06
fix: Allow updating state as node end

bracesproul:


libs/langgraph/src/pregel/index.ts
```diff
@@ -642,8 +646,13 @@ export class Pregel<
     let checkpointConfig = patchConfigurable(config, {
       checkpoint_ns: config.configurable?.checkpoint_ns ?? "",
     });
+    let checkpointMetadata = config.metadata ?? {};
     if (saved?.config.configurable) {
       checkpointConfig = patchConfigurable(config, saved.config.configurable);
+      checkpointMetadata = {
+        ...saved.metadata,
+        ...checkpointMetadata,
+      };
```
bracesproul, line 655: is in PY, was missing in JS

libs/langgraph/src/pregel/index.ts
```diff
@@ -661,6 +670,79 @@ export class Pregel<
...
+      if (saved) {
+        // tasks for this checkpoint
+        const nextTasks = _prepareNextTasks(
+          checkpoint,
+          saved.pendingWrites || [],
+          this.nodes,
+          channels,
+          managed,
+          saved.config,
+          true, // Should I pass forExecution=true here?
```
bracesproul, line 693: question

nfcampos: yes

nfcampos (dismissed):

libs/langgraph/src/pregel/index.ts
bracesproul, line 693:
```suggestion
          true,
```

PR #713, opened 2024-12-06
fix(langgraph): Allow multiple interrupts per node

bracesproul:


libs/langgraph/src/tests/pregel.test.ts
```diff
@@ -9187,6 +9187,74 @@ export function runPregelTests(
...
+
+    const thirdResult = await gatherIterator(
+      graph.stream(
+        new Command({
+          resume: "answer 2",
+        }),
+        config
+      )
+    );
+    expect(thirdResult[thirdResult.length - 1].myKey).toEqual("DEanswer 1 answer 2");
```
bracesproul, line 9254: confirming here: it should be the last item in the array since it's `streamMode: "values"`?

In this test `thirdResult` is an array with two objects, the first is the initial state (first state update) and the second is the state after the node can return (second & last state update)

nfcampos (approved):

PR #717, opened 2024-12-06
feat(langgraph): Edgeless graph support

jacoblee93:
Also fixes a small bug in conditional edge tracing


libs/langgraph/src/graph/state.ts
```diff
@@ -488,13 +501,32 @@ export class CompiledStateGraph<
...
+        return input.update;
+      }
+      return input;
+    }
+
+    // to avoid name collision below
+    const nodeKey = key;
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    function getStateKey(key: keyof U, input: U): any {
```
nfcampos, line 518: use unknown

nfcampos (approved):

PR #718, opened 2024-12-06
fix(langgraph): Re-throw NodeInterrupt errors from ToolNode for HITL.

jaredhanson:
This changes `ToolNode` to always re-throw `NodeInterrupt` errors, rather than feeding errors back into the agent.

As documented [here](https://langchain-ai.github.io/langgraphjs/how-tos/dynamic_breakpoints/), the intent of `NodeInterrupt` is to act as a breakpoint for when a human needs to be brought into the loop.  This works as intended when thrown from a step function.  However, when wrapped in a `ToolNode`, the default behavior traps `NodeInterrupt` and feeds it back to the agent.   This is unexpected, since if a human is necessary the agent can't recover on its own.

Furthermore, setting `handleToolErrors` to `false` is a way to workaround this, but that then requires the developer to feed all other non-NodeInterrupt errors back manually (which is also unexpected).

The best default behavior seems to be feeding all errors _except_ NodeInterrupt back, and re-throwing NodeInterrupt so the application can handle it appropriately.

I discussed this with @jacoblee93 on an internal Slack channel.

Twitter: [@jaredhanson](https://x.com/jaredhanson)


jacoblee93:
Thank you!

PR #720, opened 2024-12-09
feat(langgraph): Run Sends in the current superstep

jacoblee93:


libs/langgraph/src/pregel/algo.ts
```diff
@@ -265,6 +288,7 @@ export function _applyWrites<Cc extends Record<string, BaseChannel>>(
       if (IGNORE.has(chan)) {
         // do nothing
       } else if (chan === TASKS) {
+        // TODO: remove branch in 1.0
```
nfcampos, line 291: rm this comment

libs/langgraph/src/pregel/algo.ts
```diff
@@ -440,6 +469,74 @@ export function _prepareNextTasks<
       tasks[task.id] = task;
     }
   }
+  // Consume pending Sends from this step (new version of Send)
```
nfcampos, line 472: remove this block

nfcampos (commented):
comments for ben

benjamincburns:
Closing, as we won't be moving forward w/ this feature and we have the necessary parts from this in the functional API now

PR #741, opened 2024-12-13
feat(langgraph): Adds error page for unreachable nodes

jacoblee93:
CC @vbarda @eyurtsev

### DIFF
--- a/libs/langgraph/src/graph/graph.ts
+++ b/libs/langgraph/src/graph/graph.ts
@@ -32,7 +32,11 @@ import {
   gatherIteratorSync,
   RunnableCallable,
 } from "../utils.js";
-import { InvalidUpdateError, NodeInterrupt } from "../errors.js";
+import {
+  InvalidUpdateError,
+  NodeInterrupt,
+  UnreachableNodeError,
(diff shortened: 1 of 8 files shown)


libs/langgraph/src/graph/graph.ts
```diff
@@ -463,7 +467,18 @@ export class Graph<
...
     for (const node of Object.keys(this.nodes)) {
       if (!allTargets.has(node)) {
-        throw new Error(`Node \`${node}\` is not reachable`);
+        throw new UnreachableNodeError(
+          [
+            `Node \`${node}\` is not reachable.`,
+            "",
+            "If you are returning Command objects from your node,",
+            'make sure you are passing names of potential destination nodes as an "ends" array',
+            'into ".addNode(..., { ends: ["node1", "node2"] })".',
```
bracesproul, line 476: quote formatting is weird here (some are `"`, and others are `'`). how is this passing the formatter?

jacoblee93: I have double quotes in the string so you need to use a different one

PR #788, opened 2025-01-17
fix(langgraph): add structured response format to prebuilt react agent

isahers1:


libs/langgraph/src/prebuilt/react_agent_executor.ts
```diff
@@ -30,14 +31,20 @@ import {
 import { MessagesAnnotation } from "../graph/messages_annotation.js";
 import { ToolNode } from "./tool_node.js";
 import { LangGraphRunnableConfig } from "../pregel/runnable_types.js";
+import { Annotation } from "../graph/annotation.js";
+import { Messages, messagesStateReducer } from "../graph/message.js";
 
+type StructuredResponse =
+  | z.ZodType<Record<string, unknown>>
```
jacoblee93, line 38: Don't think `z.ZodType` belongs in inputs? We are not outputting a zod schema

Let's also type as `Record<string, any>` until we get better typing

libs/langgraph/src/prebuilt/react_agent_executor.ts
```diff
@@ -223,6 +238,19 @@ export type CreateReactAgentParams<
...
+   * If provided, output will be formatted to match the given schema and returned in the 'structured_response' state key.
+   * If not provided, `structured_response` will not be present in the output state.
+   *
+   * Can be passed in as:
+   *   - Zod schema
+   *   - Dictionary object
+   *   - [prompt, schema], where schema is one of the above.
+   *        The prompt will be used together with the model that is being used to generate the structured response.
+   */
+  responseFormat?: StructuredResponse | [string, StructuredResponse];
```
jacoblee93, line 253: I know we do this everywhere in Python but would really like to not use tuples ever again in JS

jacoblee93 (commented):
Thanks, see comments

We should also do a better job of typing the structured output but I can fix that up later before merging

PR #791, opened 2025-01-20
patch(langgraph): Refactor pregel loop to use new `PregelRunner` class, ported from python LangGraph.

benjamincburns:
This work is a prerequisite to porting the imperative (functional?) API over from python LangGraph.

### DIFF
--- a/libs/langgraph/src/pregel/retry.ts
+++ b/libs/langgraph/src/pregel/retry.ts
@@ -1,11 +1,15 @@
-import { CHECKPOINT_NAMESPACE_SEPARATOR, Command } from "../constants.js";
+import {
+  CHECKPOINT_NAMESPACE_SEPARATOR,
+  Command,
+  CONFIG_KEY_RESUMING,
+} from "../constants.js";
 import {
   getSubgraphsSeenSet,
   isGraphBubbleUp,
   isParentCommand,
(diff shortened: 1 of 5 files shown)


libs/langgraph/src/pregel/retry.ts
```diff
@@ -184,18 +147,21 @@ async function _runWithRetry(
...
-        error.name ??
+        (error as Error).name ??
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        (error.constructor as any).unminifiable_name ??
-        error.constructor.name;
+        ((error as Error).constructor as any).unminifiable_name ??
+        (error as Error).constructor.name;
       console.log(
-        `Retrying task "${pregelTask.name}" after ${interval.toFixed(
+        `Retrying task "${String(pregelTask.name)}" after ${interval.toFixed(
```
nfcampos, line 155: this doesn't do anything, right? string interpolation does calls toString on things anyway

nfcampos (approved):

libs/langgraph/src/pregel/retry.ts
benjamincburns: Ah right, it's necessary because the type of `PregelExecutableTask.name` is `string | Symbol`.

PR #793, opened 2025-01-21
feat(langgraph): functional API

benjamincburns:
# feat(langgraph): Add Functional API

This PR introduces a new functional API to langgraphjs that provides a simpler, more intuitive way to create and compose tasks in a graph. The changes focus on improving developer experience while maintaining the power and flexibility of the existing graph-based system.

### 1. New Functional API Components

- Added new `task` and `entrypoint` functions for creating graph nodes
- Introduced experimental TaskOptions and EntrypointOptions interfaces
- Added support for retry policies at the task level

### 2. Configuration and Constants Updates

- Added new configuration keys for the functional API (`CONFIG_KEY_CALL`)
- Introduced new control flow constants (`NO_WRITES`, `RETURN`)
- Enhanced type safety for the SendInterface validation

### 3. Pregel Engine Improvements

- Modified task handling to support functional API calls
- Enhanced error handling and null value checking in task writes
- Updated the IGNORE set to include new control flow signals

### 4. Type System Enhancements

- Added new types for task paths and calls
- Improved type safety throughout the codebase

## Implementation Details

The implementation focuses on maintaining backward compatibility while introducing the new functional API. Key architectural decisions include:

1. Using the Pregel engine as the underlying execution system
2. Maintaining the existing channel-based communication system
3. Adding proper TypeScript types for improved developer experience
4. Implementing proper error handling and validation

## Example Usage

```typescript
import { task, entrypoint } from "@langchain/langgraph";

// Define a task with optional retry policy
const processData = task(
  "process",
  async (data) => {
    // Process the data
    return result;
  },
  { retry: { maxAttempts: 3 } }
);

// Create an entrypoint
const workflow = entrypoint({ name: "main" }, async (input) => {
  const result = await processData(input);
  return result;
});
```

## Breaking Changes

None. This is a purely additive change that maintains compatibility with existing code.

## Testing

Added new test files and cases to cover the functional API functionality:

- `func.test.ts` for testing the new API
- Additional test coverage in existing test files


libs/langgraph/src/pregel/runner.ts
```diff
@@ -113,10 +133,137 @@ export class PregelRunner {
...
+              // Task failed
+              const error =
+                // eslint-disable-next-line no-instanceof/no-instanceof
+                errorValue instanceof Error
+                  ? errorValue
+                  : new Error(String(errorValue));
+
+              rtn[idx] = Promise.reject(error);
+            } else {
+              // the only way this should happen is if the task executes multiple times and writes aren't cleared
```
nfcampos, line 210: or if it was executed without saving either error or return, but that's not expected either.

libs/langgraph/src/pregel/types.ts
```diff
@@ -232,3 +237,51 @@ export type PregelScratchpad<Resume> = {
...
+  readonly __lg_type = "call";
+
+  constructor({ func, name, input, retry, callbacks }: CallOptions) {
+    this.func = func;
+    this.name = name;
+    this.input = input;
+    this.retry = retry;
+
+    // TODO: where is this used?
+    this.callbacks = callbacks;
```
nfcampos, line 269: it should be used in prepareSingleTask I believe

libs/langgraph/src/constants.ts
```diff
@@ -58,7 +61,14 @@ export interface SendInterface {
 
 export function _isSendInterface(x: unknown): x is SendInterface {
   const operation = x as SendInterface;
-  return typeof operation.node === "string" && operation.args !== undefined;
+  return (
+    x != null &&
+    typeof x === "object" &&
+    "node" in x &&
+    "args" in x &&
```
jacoblee93, line 68: nit: just use `operation` consistently here?

jacoblee93: And actually, why was the previous guard not sufficient?

libs/langgraph/src/func.ts
```diff
@@ -0,0 +1,114 @@
...
+ * @param options.checkpointer - The checkpointer for the entrypoint
+ * @param func - The function that executes this entrypoint
+ * @returns A Pregel instance that can be run
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function entrypoint<FuncT extends (...args: any[]) => any>(
+  { name, checkpointer }: EntrypointOptions,
+  func: FuncT
+): Pregel<
+  Record<string, PregelNode>,
```
nfcampos, line 74: we can be a little stricter here, we know the exact node and channel names that will be used

libs/langgraph/src/func.ts
```diff
@@ -0,0 +1,114 @@
...
+  /**
+   * The retry policy for the task
+   */
+  retry?: RetryPolicy;
+};
+
+/**
+ * Wraps a function in a task that can be retried
+ *
+ *  !!! warning "Experimental"
```
jacoblee93, line 29: Do we actually want this warning in? Seems harsh

libs/langgraph/src/func.ts
```diff
@@ -0,0 +1,114 @@
...
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function task<FuncT extends (...args: any[]) => any>(
+  name: string,
+  func: FuncT,
+  options?: TaskOptions
+): (...args: Parameters<FuncT>) => Promisified<FuncT> {
+  return (...args: Parameters<FuncT>) => {
+    const ret = call({ func, name, retry: options?.retry }, ...args);
+    // eslint-disable-next-line no-instanceof/no-instanceof
+    if (ret instanceof Promise) {
```
nfcampos, line 45: I think this check should be in call, no?

jacoblee93: This is a bit weird? Why can't we just wrap if it doesn't return a promise?

libs/langgraph/src/pregel/types.ts
```diff
@@ -24,6 +24,11 @@ export type PregelInputType = any;
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type PregelOutputType = any;
 
+export type Promisified<T extends (...args: unknown[]) => unknown> =
```
nfcampos, line 27: I think the goal of this file is to export types we want to expose in the public api of the library. utility types should either be defined in the file they're used or in some private file

libs/langgraph/src/pregel/algo.ts
```diff
@@ -64,7 +75,7 @@ export type WritesProtocol<C = string> = {
   name: string;
   writes: PendingWrite<C>[];
   triggers: string[];
-  path?: [string, ...(string | number)[]];
+  path?: TaskPath;
```
jacoblee93, line 78: Should have refactored this ages ago, thanks

libs/langgraph/src/pregel/call.ts
```diff
@@ -0,0 +1,63 @@
...
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function getRunnableForFunc<FuncT extends (...args: any[]) => any>(
+  name: string,
+  func: FuncT
+): RunnableSequence<Parameters<FuncT>, ReturnType<FuncT>> {
+  const run = new RunnableCallable<Parameters<FuncT>, ReturnType<FuncT>>({
+    func: (input: Parameters<FuncT>) => func(...input),
+    name,
+    trace: false,
+    recurse: true,
```
nfcampos, line 24: this actually shouldn't recurse for task/entrypoint, I'll change it in py if it was accidentally set to do so

libs/langgraph/src/pregel/call.ts
```diff
@@ -0,0 +1,63 @@
+import { RunnableConfig, RunnableSequence } from "@langchain/core/runnables";
+import { AsyncLocalStorageProviderSingleton } from "@langchain/core/singletons";
+import { CONFIG_KEY_CALL, RETURN, TAG_HIDDEN } from "../constants.js";
+import { ChannelWrite, PASSTHROUGH } from "./write.js";
+import { RetryPolicy } from "./utils/index.js";
+import { Promisified } from "./types.js";
+import { RunnableCallable, type RunnableCallableArgs } from "../utils.js";
+/**
+ * Cache for storing runnable sequences to avoid recreating them
```
jacoblee93, line 9: What does this apply to?

libs/langgraph/src/pregel/call.ts
```diff
@@ -0,0 +1,63 @@
...
+    retry?: RetryPolicy;
+  };
+
+export function call<FuncT extends (...args: unknown[]) => unknown>(
+  { func, name, retry }: CallWrapperOptions<FuncT>,
+  ...args: Parameters<FuncT>
+): Promisified<FuncT> {
+  const config =
+    AsyncLocalStorageProviderSingleton.getRunnableConfig() as RunnableConfig;
+  // TODO: type the CONFIG_KEY_CALL function
```
jacoblee93, line 55: Just make sure types for this don't appear in autocomplete for user

libs/langgraph/src/pregel/call.ts
benjamincburns: I think it was, but I could've misread - good eye, either way

libs/langgraph/src/pregel/call.ts
nfcampos: @benjamincburns didn't we decide not to do this?

libs/langgraph/src/func.ts
nfcampos: Yea this is not the copy I'd want, I think better to say something like

!!! warning "Beta"
     This is a new API which might evolve over the next few weeks.
     We plan to mark it as stable by X day [next 3 weeks]

libs/langgraph/src/pregel/call.ts
nfcampos: yea opened pr in py

libs/langgraph/src/pregel/runner.ts
```diff
@@ -113,10 +133,137 @@ export class PregelRunner {
...
+
+      // eslint-disable-next-line no-instanceof/no-instanceof
+      if (result !== undefined) {
+        if (result.length === 1) {
+          return result[0];
+        }
+        return Promise.all(result);
+      }
+
+      return Promise.resolve(result);
```
nfcampos, line 266: what's this for?

libs/langgraph/src/func.ts
benjamincburns: I lifted it directly from the python lib. The `@experimental` and `@beta` tsdoc tags should ideally convey the same meaning, although per [the spec](https://tsdoc.org/pages/tags/beta/#suggested-meaning) they may be removed by tooling on non-beta releases. I haven't checked the docs build yet to see what actually happens, and whether we can configure this, so I left it in for now as a sort of reminder to myself.

libs/langgraph/src/pregel/algo.ts
benjamincburns: np - there's always _something_ that needs polishing!

libs/langgraph/src/pregel/call.ts
benjamincburns: Yeah, this is for internal devX. Gist is that we're cruisin' for a bruisin' with this passing scheme, as everything is implicit `any` here, which makes things hard to test & maintain. Also makes it less discoverable for external contributors. I won't go too crazy with it in this PR, but my plan is to clean it up some so these keys are typed properly at least for the ones I'm touching here.

libs/langgraph/src/pregel/runner.ts
benjamincburns: Good catch - I was basically telling the type checker to bugger off because `writer` actually returns `Promise<unknown>[] | undefined`.

I'm erring on the side of matching the structure of the python lib, however were that not the case I'd take this as a strong sign that `writer` is taking on too many responsibilities and needs a refactor. I may well extract out a bunch of things to private methods and/or utility functions, just to make the main/high-level code paths more readable -- cyclomatic complexity is quite high in a lot of this code.

libs/langgraph/src/constants.ts
benjamincburns: > why was the previous guard not sufficient

This was the old one:

```typescript
export function _isSendInterface(x: unknown): x is SendInterface {
  const operation = x as SendInterface;
  return typeof operation.node === "string" && operation.args !== undefined;
}
```

`_prepareNextTask` was passing `undefined`, causing this guard to throw a `ReferenceError` on access of `operation.node`.

Side note: I'd have expected that the `typeof x === "object"` check to rule out null/undefined in later clauses of the statement. For some reason (likely `strictNullChecks` or similar in `compilerOptions`?) it doesn't, and the type checker gets angry on the `node in x` clause if I omit the `x != null`.

benjamincburns: will clean up the nit, though.

libs/langgraph/src/pregel/call.ts
benjamincburns: I apparently missed the comment.

libs/langgraph/src/func.ts
benjamincburns: We know the node names, but unfortunately we don't know the channel names. TypeScript's type system erases function parameter names and only captures their type and order. The best we can do is expose it as a generic argument to `entrypoint` and allow users to define it themselves. :-(

benjamincburns: Hmm, I might've been misunderstanding how channels apply here. Just debugging the streaming stuff now and that is shining some light on it, though.

nfcampos: the channels here are END and START I believe

libs/langgraph/src/pregel/runner.ts
nfcampos: cool, and agreed

libs/langgraph/src/tests/func.test.ts
```diff
@@ -274,110 +274,62 @@ export function runFuncTests(
...
+            { name: "graph", checkpointer },
+            async (input: string, config: LangGraphRunnableConfig) => {
+              config.writer?.(`hello ${input}`);
+              await new Promise((resolve) => {
+                setTimeout(resolve, 100);
+              });
+              config.writer?.(`hello again, ${input}`);
+              return `goodbye, ${input}`;
+            }
+          );
```
benjamincburns, line 288: @nfcampos @eyurtsev - not worth reviewing this entire PR yet, but just confirming that this behavior is what we're expecting for sans entrypoint generators.

I suppose I should also write a test demonstrating subgraph streaming 🤔

libs/langgraph/src/constants.ts
```diff
@@ -33,18 +47,28 @@ export const TASK_NAMESPACE = "6ba7b831-9dad-11d1-80b4-00c04fd430c8";
...
-  CONFIG_KEY_STREAM,
+  CONFIG_KEY_CALL,
+  CONFIG_KEY_RESUME_VALUE,
+  CONFIG_KEY_SCRATCHPAD,
+  CONFIG_KEY_PREVIOUS_STATE,
   CONFIG_KEY_CHECKPOINT_MAP,
-  INPUT,
+  CONFIG_KEY_CHECKPOINT_NS,
+  CONFIG_KEY_CHECKPOINT_ID,
 ];
```
benjamincburns, line 72: aligned order w/ python code so it's easier to check

libs/langgraph/src/func.ts
benjamincburns: @nfcampos @eyurtsev is the current copy in the python lib what we're happy to go with here?

for reference:

```
    !!! warning "Beta"
        The Functional API is currently in beta and is subject to change.
```

libs/langgraph/src/pregel/call.ts
benjamincburns: saving this one for later - needs a bigger refactor than I wanted to add to this already huge PR.

libs/langgraph/src/tests/utils.ts
```diff
@@ -406,3 +414,97 @@ export class _AnyIdAIMessageChunk extends AIMessageChunk {
...
+}
+
+export async function dumpDebugStream<
+  Nn extends StrRecord<string, PregelNode>,
+  Cc extends StrRecord<string, BaseChannel | ManagedValueSpec>,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  ConfigurableFieldType extends Record<string, any> = StrRecord<string, any>,
+  InputType = PregelInputType,
+  OutputType = PregelOutputType
+>(
```
benjamincburns, line 433: these are actually unused now, but I left them in, in case they're handy.

libs/langgraph/src/pregel/utils/config.ts
```diff
@@ -92,3 +93,29 @@ export function ensureLangGraphConfig(
...
+    AsyncLocalStorageProviderSingleton.getRunnableConfig();
+  return config?.store;
+}
+
+/**
+ * A helper utility function that returns the {@link LangGraphRunnableConfig#writer} if "custom" stream mode is enabled, otherwise undefined
+ *
+ * @returns a reference to the {@link LangGraphRunnableConfig#writer} if "custom" stream mode is enabled, otherwise undefined
+ */
+export function getWriter(): ((chunk: unknown) => void) | undefined {
```
jacoblee93, line 113: Export this from the main entrypoint

jacoblee93 (approved):
Overall looks good to me, will leave for @nfcampos for final review

libs/langgraph/src/func/index.ts
```diff
@@ -0,0 +1,396 @@
...
+ * Define a LangGraph workflow using the `entrypoint` function.
+ *
+ * !!! warning "Beta"
+ *     The Functional API is currently in beta and is subject to change.
+ *
+ * @beta
+ *
+ * ### Function signature
+ *
+ * The decorated function must accept at most **two parameters**. The first parameter
```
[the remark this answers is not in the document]
benjamincburns: teeeechnically it still follows the decorator pattern (just without the nice syntax that python offers), but I agree - should refer to it as "wrapped."

libs/langgraph/src/func/index.ts
```diff
@@ -0,0 +1,396 @@
...
+ *     thread_id: "some_thread"
+ *   }
+ * };
+ *
+ * await myWorkflow.invoke(3, config); // 0 (previous was undefined)
+ * await myWorkflow.invoke(1, config); // 6 (previous was 3 * 2 from the previous invocation)
+ * ```
+ */
+export const entrypoint = function entrypoint<InputT, OutputT>(
+  options: EntrypointOptions,
```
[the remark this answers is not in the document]
benjamincburns: yeah, would make it consistent with `task` good spot.

libs/langgraph/src/func/types.ts
```diff
@@ -0,0 +1,85 @@
+import { LangGraphRunnableConfig } from "../pregel/runnable_types.js";
+
+/**
+ * Allows the entrypoint function to return a value to the caller, as well as a separate state value to persist to the checkpoint
+ */
+export type EntrypointFinal<ValueT, SaveT> = {
```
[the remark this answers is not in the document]
benjamincburns: It's the type that's returned by `entrypoint.final`, so I think it needs to be exported?

libs/langgraph/src/pregel/index.ts
```diff
@@ -1239,6 +1240,7 @@ export class Pregel<
      * from the loop to the main stream and therefore back to the user as soon as
      * they are available.
      */
+
```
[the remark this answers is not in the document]
benjamincburns: odd, figured `prettier` would've fixed that 🤔

libs/langgraph/src/pregel/utils/config.ts
benjamincburns: ah, good catch!

libs/langgraph/src/func/index.ts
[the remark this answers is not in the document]
benjamincburns: actually, I think I'll make both overloaded so you can do

```
entrypoint("name", () => ...)
entrypoint({ name }, () => ...)
task("name", () => ...)
task({ name }, () => ...)
```

PR #798, opened 2025-01-23
fix(messages): reuse the first valid message ID for subsequent chunks

dqbd:
Fixes # (issue)

### DIFF
--- a/libs/langgraph/src/pregel/messages.ts
+++ b/libs/langgraph/src/pregel/messages.ts
@@ -1,4 +1,3 @@
-import { v4 } from "uuid";
 import {
   BaseCallbackHandler,
   HandleLLMNewTokenCallbackFields,
@@ -44,6 +43,8 @@ export class StreamMessagesHandler extends BaseCallbackHandler {
 
   emittedChatModelRunIds: Record<string, boolean> = {};
 
+  stableMessageIdMap: Record<string, string> = {};
+
   lc_prefer_streaming = true;
 
   constructor(streamFn: (streamChunk: StreamChunk) => void) {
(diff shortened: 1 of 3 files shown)


libs/langgraph/src/tests/pregel.test.ts
```diff
@@ -8564,24 +8564,6 @@ graph TD;
           tags: ["c_two_chat_model"],
         },
       ],
-      [
```
nfcampos, line 8567: were these being emitted by mistake?

nfcampos (approved):

libs/langgraph/src/tests/pregel.test.ts
dqbd: I think so? @jacoblee93 can clarify, but it doesn't make sense to emit the message twice

nfcampos: cool yea def don't want to emit things twice, can you point me to the line that was causing this?

dqbd: The issue happens on when we emit the actual chunk instead of a raw message: https://github.com/langchain-ai/langgraphjs/pull/798/files?w=1#diff-fdc54140c541e7ddfea663ff6751691ca2971772df39a299c904947b27873560R127

nfcampos (dismissed):

libs/langgraph/src/pregel/messages.ts
```diff
@@ -44,28 +43,42 @@ export class StreamMessagesHandler extends BaseCallbackHandler {
...
     }
-    if (message.id === undefined) {
-      const id = v4();
+
+    // For instance in ChatAnthropic, the first chunk has an message ID
+    // but the subsequent chunks do not. To avoid clients seeing two messages
+    // we rename the message ID if it's being auto-set to `run-${runId}`
+    // (see https://github.com/langchain-ai/langchainjs/pull/6646).
+    let messageId = message.id;
+    if (messageId == null || messageId === `run-${runId}`) {
```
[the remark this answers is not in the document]
dqbd: `==` does check for both `undefined` and `null`, the new condition would check for empty string as well.

Issue #684, opened 2024-11-20
RemoteGraph's .stream() "messages" streamMode is not working


Jronk:
For example, following this how-to guide: [How to stream LLM tokens from your graph](https://langchain-ai.github.io/langgraphjs/how-tos/stream-tokens/), And using [How to interact with the deployment using RemoteGraph](https://langchain-ai.github.io/langgraphjs/how-tos/use-remote-graph/), does not work and the messages are not properly streaming.

TL;DR, it _appears_ the bug/issue happens [here](https://github.com/langchain-ai/langgraphjs/blob/main/libs/langgraph/src/pregel/remote.ts#L471-L477), where the otherwise valid messages to be streamed are getting short-circuited and the generator yields nothing instead of the messages it should. The values for `chunk.event.split(CHECKPOINT_NAMESPACE_SEPARATOR)[0])` look like: `messages/partial` or `messages/complete` but since `updatedStreamModes` looks like `['messages', 'updates']` it will never match...

![Screen Shot 2024-11-19 at 12 24 32 PM](https://github.com/user-attachments/assets/255e8ac3-70e5-444d-bb7a-f4869508d678)

For a proof of concept, building on the above two article's examples, here is an example simple graph and tests to show how local works, but remote does not:

```typescript
import { StateGraph, END } from "@langchain/langgraph";
import { ToolNode } from "@langchain/langgraph/prebuilt";
import { StateAnnotation } from "./simple-messages-state.js";
import { ChatOpenAI } from "@langchain/openai";

import { tool } from "@langchain/core/tools";
import { z } from "zod";

import { AIMessage } from "@langchain/core/messages";

const model = new ChatOpenAI({
  model: "gpt-4o-mini",
  temperature: 0,
  streaming: true
});

const searchTool = tool((_) => {
  // This is a placeholder for the actual implementation
  return "Cold, with a low of 3℃";
}, {
  name: "search",
  description:
    "Use to surf the web, fetch current information, check the weather, and retrieve other information.",
  schema: z.object({
    query: z.string().describe("The query to use in your search."),
  }),
});

const tools = [searchTool];

const toolNode = new ToolNode(tools);

const boundModel = model.bindTools(tools);

const routeMessage = (state: typeof StateAnnotation.State) => {
  const { messages } = state;
  const lastMessage = messages[messages.length - 1] as AIMessage;
  // If no tools are called, we can finish (respond to the user)
  if (!lastMessage?.tool_calls?.length) {
    return END;
[...]

PR #799, opened 2025-01-24
fix(remote): use "messages-tuple" when requesting "messages"

dqbd:
~Marking as draft, because we might want to deserialise raw JSON objects to actual messages~ Turns out the Python's RemoteGraph does not convert the message tuples back to BaseMessage anyway

https://github.com/langchain-ai/langgraphjs/issues/684

### DIFF
--- a/libs/langgraph/src/pregel/remote.ts
+++ b/libs/langgraph/src/pregel/remote.ts
@@ -105,16 +105,12 @@ const getStreamModes = (
     (typeof streamMode === "string" ||
       (Array.isArray(streamMode) && streamMode.length > 0))
   ) {
-    if (typeof streamMode === "string") {
-      updatedStreamModes.push(streamMode);
-    } else {
-      reqSingle = false;
-      updatedStreamModes.push(...streamMode);
-    }
+    reqSingle = typeof streamMode === "string";
+    const mapped = Array.isArray(streamMode) ? streamMode : [streamMode];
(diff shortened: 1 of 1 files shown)


libs/langgraph/src/pregel/remote.ts
```diff
@@ -105,16 +105,12 @@ const getStreamModes = (
...
   ) {
-    if (typeof streamMode === "string") {
-      updatedStreamModes.push(streamMode);
-    } else {
-      reqSingle = false;
-      updatedStreamModes.push(...streamMode);
-    }
+    reqSingle = typeof streamMode === "string";
+    const mapped = Array.isArray(streamMode) ? streamMode : [streamMode];
+    updatedStreamModes.push(...mapped);
```
jacoblee93, line 110: This is equivalent to what was there before right?

jacoblee93 (approved):
Does this change anything for the end-user in terms of format? Or was `streamMode: messages` just not doing the right thing?

libs/langgraph/src/pregel/remote.ts
dqbd: Yep!

dqbd:
I think the previous implementation didn't work. The expected behaviour should be that calling `stream` on `RemotePregel` with `messages` streamMode should behave the same like calling `stream` on `Pregel`.

PR #801, opened 2025-01-24
docs: functional API

ccurme:
Will coordinate with other docs changes to avoid merge conflicts.

PR #810, opened 2025-01-28
fix(langgraph): Add support for returnDirect in createReactAgent, typing fixes

jacoblee93:
Fixes nits encountered during functional API docs porting

- Circular LangSmith serialization issue when passing AsyncBatchStore into tasks
- Export `messagesStateReducer` as `addMessages` and relax typing
- Add support for tools with `returnDirect: true` for `createReactAgent`

Should also follow up with a new `@langchain/langgraph-checkpoint` version and bump requirement in `@langchain/langgraph`

CC @benjamincburns 

Other half of #801


benjamincburns:
Merging as part of #793

PR #838, opened 2025-02-02
feat(checkpoint-postgres): Add support for providing a custom schema during initialization

nick-w-nick:
This PR adds support for specifying a custom schema in the `checkpoint-postgres` module during initialization. Instead of relying on the default `public` schema used by Postgres or modifying the connection string’s search path (which can be unreliable at best), SQL queries now explicitly reference the provided schema, ensuring compatibility across all PostgreSQL versions and configurations, while still allowing the original module to stay backward compatible with previous versions since `public` was already the default schema.

```ts
// you can optionally pass a configuration object as the second parameter
const checkpointer = PostgresSaver.fromConnString("postgresql://...", {
  schema: "schema_name" // defaults to "public"
});
```

### Key Changes
- **Custom Schema Support**
  - Added an extendable `options` parameter to specify a schema when creating a `PostgresSaver` instance via the `fromConnString` method
  - Consolidated SQL queries and migration scripts into reusable functions that accept a schema name
- **Documentation & Testing**
  - Improved JSDoc comments for `fromConnString` with usage example
  - Enhanced integration tests to validate behavior with both the default and a custom schema

### DIFF
--- a/libs/checkpoint-postgres/src/index.ts
+++ b/libs/checkpoint-postgres/src/index.ts
@@ -13,7 +13,7 @@ import {
 import pg from "pg";
 
 import { getMigrations } from "./migrations.js";
-import { type SQL_STATEMENTS, getSQLStatements, getTables } from "./sql.js";
+import { type SQL_STATEMENTS, getSQLStatements, getTablesWithSchema } from "./sql.js";
 
 const { Pool } = pg;
 
@@ -93,7 +93,7 @@ export class PostgresSaver extends BaseCheckpointSaver {
    */
   async setup(): Promise<void> {
     const client = await this.pool.connect();
-    const SCHEMA_TABLES = getTables(this.schema);
+    const SCHEMA_TABLES = getTablesWithSchema(this.schema);
     try {
       await client.query(`CREATE SCHEMA IF NOT EXISTS ${this.schema}`);
       let version = -1;
--- a/libs/checkpoint-postgres/src/index.ts
+++ b/libs/checkpoint-postgres/src/index.ts
@@ -12,13 +12,10 @@ import {
 } from "@langchain/langgraph-checkpoint";
 import pg from "pg";
 
-import { MIGRATIONS } from "./migrations.js";
+import { getMigrations } from "./migrations.js";
 import {
-  INSERT_CHECKPOINT_WRITES_SQL,
-  SELECT_SQL,
-  UPSERT_CHECKPOINT_BLOBS_SQL,
-  UPSERT_CHECKPOINT_WRITES_SQL,
-  UPSERT_CHECKPOINTS_SQL,
+  type SQL_STATEMENTS,
+  getSQLStatements,
 } from "./sql.js";
 
 const { Pool } = pg;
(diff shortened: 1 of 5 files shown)


benjamincburns:
Thanks for the contribution! I'll have a closer look tomorrow, but from a quick glance this looks really nice!

benjamincburns:
Don't worry about the docs deploy - that's expected to fail on external PRs for the moment. 😅

libs/checkpoint-postgres/src/index.ts
```diff
@@ -61,17 +71,49 @@ const { Pool } = pg;
...
+   * @example
+   * const connString = "postgresql://user:password@localhost:5432/db";
+   * const checkpointer = PostgresSaver.fromConnString(connString, {
+   *  schema: "custom_schema" // defaults to "public"
+   * });
+   * await checkpointer.setup();
+   */
+  static fromConnString(
+    connString: string,
+    options: PostgresSaverOptions = {}
```
benjamincburns, line 113: This is kind of a nit, but on first read through I thought that perhaps you were mistakenly causing default options to not be applied by specifying this default value.

It might be more readable/explicit if you have a module-scoped `_defaultOptions`, kept the `options` arg here and in the ctor optional, and then conditionally applied that at the start of the method.

benjamincburns (changes requested):
Just one small change requested, please & thanks.

libs/checkpoint-postgres/src/index.ts
nick-w-nick: @benjamincburns I've pushed a commit to update the default options behavior, including moving it to a module-scoped `_defaultOptions` variable.

Let me know if this works for you.

Thanks!

libs/checkpoint-postgres/src/index.ts
```diff
@@ -61,17 +80,44 @@ const { Pool } = pg;
...
+   * @param {PostgresSaverOptions} [options] - Optional configuration object.
+   * @returns {PostgresSaver} A new instance of PostgresSaver.
+   *
+   * @example
+   * const connString = "postgresql://user:password@localhost:5432/db";
+   * const checkpointer = PostgresSaver.fromConnString(connString, {
+   *  schema: "custom_schema" // defaults to "public"
+   * });
+   * await checkpointer.setup();
+   */
```
benjamincburns, line 114: Thanks for adding the reference docs, btw!

libs/checkpoint-postgres/src/index.ts
```diff
@@ -12,15 +12,30 @@ import {
...
+};
+
+const _ensureCompleteOptions = (
+  options?: Partial<PostgresSaverOptions>
+): PostgresSaverOptions => {
+  return {
+    ..._defaultOptions,
+    ...options,
+  };
+};
```
benjamincburns, line 37: Thanks for this. In the future you can also just in-line this, as it's a really common pattern for applying defaults. Having it be a well-named function call does express intent more clearly, however.

benjamincburns (approved):
Thanks again for the contribution!

PR #846, opened 2025-02-06
fix(langgraph): Dedupe input (right-side) messages in messageStateReducer

jacoblee93:
There are some cases where duplicate messages (with the same id) can accumulate in a result in before they are returned by a node and passed into a reducer

This change dedupes in new messages

FYI @benjamincburns @nfcampos


vanb (approved):
❤️

PR #858, opened 2025-02-11
langgraph: use 'prompt' param for model input preprocessing in create_react_agent

vbarda:


libs/langgraph/src/prebuilt/react_agent_executor.ts
```diff
@@ -84,68 +84,65 @@ function _convertMessageModifierToStateModifier(
   );
 }
 
-function _getStateModifierRunnable(
-  stateModifier?: StateModifier
-): RunnableInterface {
-  let stateModifierRunnable: RunnableInterface;
+const PROMPT_RUNNABLE_NAME = "state_modifier";
```
vbarda, line 87:
```suggestion
const PROMPT_RUNNABLE_NAME = "prompt";
```

PR #859, opened 2025-02-11
langgraph: allow passing model.bindTools() to createReactAgent

vbarda:


libs/langgraph/src/prebuilt/react_agent_executor.ts
```diff
@@ -142,6 +143,103 @@ function _getPrompt(
...
+        $schema: string;
+      };
+    };
+  };
+
+  type AnthropicTool = {
+    name: string;
+    description: string;
+    input_schema: Record<string, unknown>;
+  };
```
benjamincburns, line 181: these feel out of place here - can we import these from langchain/core?

libs/langgraph/src/prebuilt/react_agent_executor.ts
```diff
@@ -142,6 +143,103 @@ function _getPrompt(
...
+
+  type BoundTool = OpenAITool | AnthropicTool;
```
benjamincburns, line 183: ideally this would also come from core, but I can see this living here as it's not exported and very specific to this use case

libs/langgraph/src/prebuilt/react_agent_executor.ts
benjamincburns, line 181: are there not types for these that we can pull in from langchain/core? these feel a bit awkward living here - not the end of the world, though

vbarda: they don't exist there since they are library-specific

vbarda: actually figured out a way to simplify

libs/langgraph/src/prebuilt/react_agent_executor.ts
```diff
@@ -142,6 +146,80 @@ function _getPrompt(
...
+    else {
+      continue;
+    }
+
+    boundToolNames.add(boundToolName);
+  }
+
+  const missingTools = [...toolNames].filter((x) => !boundToolNames.has(x));
+  if (missingTools.length > 0) {
+    throw new Error(`Missing tools '${missingTools}' in the model.bindTools()`);
```
benjamincburns, line 195: Probably worth explaining why they weren't added here

benjamincburns (approved):
Looks good to me - just one nit comment

PR #874, opened 2025-02-19
chore(docs): improve API reference docs for Pregel and associated types

benjamincburns:


libs/langgraph/src/pregel/index.ts
```diff
@@ -106,30 +109,77 @@ function isString(value: unknown): value is string {
...
+ * Channels are the communication pathways between nodes in a Pregel graph.
+ * They enable message passing and state updates between different parts of the graph.
+ */
 export class Channel {
+  /**
+   * Creates a PregelNode that subscribes to a single channel.
+   * This is used to define how nodes receive input from channels.
+   *
+   * @example
+   * ```typescript
```
nfcampos, line 125: a little odd to repeat the same code block in class docs and method docs?

libs/langgraph/src/pregel/index.ts
```diff
@@ -982,7 +1270,25 @@ export class Pregel<
...
+   * - output keys
+   * - remaining config
+   * - interrupt before nodes
+   * - interrupt after nodes
+   * - checkpointer
+   * - store
+   * - whether stream mode is single
+   * @internal
+   */
+  _defaults(config: PregelOptions<Nodes, Channels>): [
```
nfcampos, line 1291: Not sure why we're documenting this (also not sure why this even exists in JS, in py it exists because we need to share it between sync and async methods)

libs/langgraph/src/pregel/index.ts
```diff
@@ -1097,11 +1404,20 @@ export class Pregel<
...
+   * This is an internal method used to set up the graph's communication channels
+   * and managed state before execution.
+   *
+   * @param config - Configuration for preparing specs
+   * @param options - Additional options
+   * @param options.skipManaged - Whether to skip initialization of managed values
+   * @returns Object containing channel specs and managed value mapping
+   * @internal
+   */
   protected async prepareSpecs(
```
nfcampos, line 1418: This should not be a method on the class, much less documented

nfcampos (dismissed):

libs/langgraph/src/pregel/index.ts
```diff
@@ -392,6 +643,16 @@ export class Pregel<
     yield* this.getSubgraphs(namespace, recurse);
   }
 
+  /**
```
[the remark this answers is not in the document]
benjamincburns: For internal use. Makes it easier for external contributors, too.

PR #877, opened 2025-02-19
feat(langgraph): send subgraph state to parent when using Command.PARENT

benjamincburns:
Ports langchain-ai/langgraph#3134

### DIFF
--- a/libs/langgraph/src/pregel/io.ts
+++ b/libs/langgraph/src/pregel/io.ts
@@ -192,13 +192,19 @@ export function* mapOutputValues<C extends PropertyKey>(
 
 /**
  * Map pending writes (a sequence of tuples (channel, value)) to output chunk.
+ * @internal
+ *
+ * @param outputChannels - The channels to output.
+ * @param tasks - The tasks to output.
+ * @param cached - Whether the output is cached.
+ *
+ * @returns A generator that yields the output chunk (if any).
  */
 export function* mapOutputUpdates<N extends PropertyKey, C extends PropertyKey>(
   outputChannels: C | Array<C>,
   tasks: readonly [PregelExecutableTask<N, C>, PendingWrite<C>[]][],
   cached?: boolean
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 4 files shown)


libs/langgraph/src/pregel/loop.ts
```diff
@@ -989,4 +991,21 @@ export class PregelLoop {
...
+          : this.outputKeys
+      );
+
+      // eslint-disable-next-line no-param-reassign
+      error.command.update = {
+        ...state,
+        ...(error.command.update ?? {}),
+      };
+    }
+  }
```
benjamincburns, line 1010: The direct port from python would've been something like the following:

```
  protected _syncStateOnParentCommand(error: unknown) {
    if (isParentCommand(error)) {
      const state = Object.entries(
        readChannels(
          this.channels,
          typeof this.outputKeys === "string"
            ? [this.outputKeys]
            : this.outputKeys
        )
      );
      
      const update = [
        ...state,
        ...error.command._updateAsTuples(),
      ]

      // eslint-disable-next-line no-param-reassign
      error.command.update = update;
    }
  }
```

When I do this, two issues occur:

1. The streaming check in my test only sees updates for `node_1`, `subgraph_node_1`, and `node_3`.
2. The test that I added in #875 fails with the following error: `InvalidUpdateError: Invalid update for channel "foo" with values [null,"a"]: LastValue can only receive one value per step.`

libs/langgraph/src/pregel/io.ts
```diff
@@ -209,63 +215,83 @@ export function* mapOutputUpdates<N extends PropertyKey, C extends PropertyKey>(
...
-  let updated: [N, Record<string, any>][];
+
+  let updated: [N, Record<string, unknown>][];
+
   if (
     outputTasks.some(([task]) =>
       task.writes.some(([chan, _]) => chan === RETURN)
     )
   ) {
+    // TODO: probably should assert that RETURN is the only "non-special" channel (starts with "__")
```
benjamincburns, line 226: left this in on purpose - current implementation is paralleled with python, but it'd be good to add more validation logic in `_commit` in general.

PR #894, opened 2025-02-24
fix(messages): avoid assigning same ID for tool messages

dqbd:
Tool messages usually don't have a message ID, thus assigning a message ID inferred from run ID will cause parallel tool responses to be dropped, as they will be merged.

### DIFF
--- a/libs/langgraph/src/pregel/messages.ts
+++ b/libs/langgraph/src/pregel/messages.ts
@@ -7,6 +7,7 @@ import {
   AIMessageChunk,
   BaseMessage,
   isBaseMessage,
+  isToolMessage,
 } from "@langchain/core/messages";
 import { Serialized } from "@langchain/core/load/serializable";
 import {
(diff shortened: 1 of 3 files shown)


vbarda (dismissed):

libs/langgraph/src/pregel/messages.ts
```diff
@@ -67,7 +67,13 @@ export class StreamMessagesHandler extends BaseCallbackHandler {
...
     let messageId = message.id;
     if (messageId == null || messageId === `run-${runId}`) {
-      messageId = this.stableMessageIdMap[runId] ?? messageId ?? `run-${runId}`;
+      messageId = this.stableMessageIdMap[runId] ?? messageId;
+
+      // Avoid assigning message ID for tool messages
+      // as doing so will assign same ID for all parallel tool responses.
+      // Which b/c the callback runs earlier than the reducer will cause
+      // other tools responses to be dropped.
+      if (message.getType() !== "tool") messageId ??= `run-${runId}`;
```
benjamincburns, line 76: nit: either `?:` or use braces - probably not a major issue with our formatting/linting setup, but sometimes people don't read closely when adding a line to a conditional block and miss that there's no curly braces.

PR #904, opened 2025-02-26
improve migration for PostgresSaver and error handling

### LINKED ISSUE #903: Error  PostgresSaver.setup(): the relation "public.checkpoint_migrations" does not exists
The issue is that the setup method is doing a select query on non-existing table, before the table can be created

mattia-consiglio:
Ensure that table `checkpoint_migrations` is created before the `SELECT` query.
This will prevent the error of non-existing table

Also fixed some minor code style errors

Fixes #903

### DIFF
--- a/libs/checkpoint-postgres/src/index.ts
+++ b/libs/checkpoint-postgres/src/index.ts
@@ -252,7 +252,7 @@ export class PostgresSaver extends BaseCheckpointSaver {
       pending_sends: [],
     };
     if ("channel_values" in serialized) {
-      serialized.channel_values = undefined;
+      delete serialized.channel_values;
     }
     return serialized;
   }
--- a/libs/checkpoint-postgres/src/index.ts
+++ b/libs/checkpoint-postgres/src/index.ts
@@ -17,6 +17,7 @@ import {
   type SQL_STATEMENTS,
   getSQLStatements,
   getTablesWithSchema,
+  tableExistsSQL,
 } from "./sql.js";
 
 interface PostgresSaverOptions {
@@ -31,8 +32,8 @@ const _ensureCompleteOptions = (
   options?: Partial<PostgresSaverOptions>
 ): PostgresSaverOptions => {
   return {
-    ..._defaultOptions,
     ...options,
+    schema: options?.schema ?? _defaultOptions.schema,
   };
 };
 
@@ -132,11 +133,17 @@ export class PostgresSaver extends BaseCheckpointSaver {
     const SCHEMA_TABLES = getTablesWithSchema(this.options.schema);
     try {
       await client.query(`CREATE SCHEMA IF NOT EXISTS ${this.options.schema}`);
-      let version = 0;
+      let version = -1;
       const MIGRATIONS = getMigrations(this.options.schema);
-
-      // Run first migration to ensure checkpoint_migrations table exists
-      await client.query(MIGRATIONS[0]);
+      // Check if checkpoint_migrations table exists
+      const checkpointMigrationsExists = await client.query(
+        tableExistsSQL(this.options.schema, SCHEMA_TABLES.checkpoint_migrations)
+      );
+      if (!checkpointMigrationsExists.rows[0].exists) {
+        // Run first migration to ensure checkpoint_migrations table exists
+        await client.query(MIGRATIONS[0]);
+        version += 1;
+      }
 
       try {
         const result = await client.query(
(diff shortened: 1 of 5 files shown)


benjamincburns:
Hi @mattia-consiglio,

Thanks for submitting this! Before we proceed, would you mind adding a test that demonstrates the need for this change? I would've expected existing tests to fail if the setup logic was broken, especially the ones in the `checkpoint-validation` package, as we run those tests over uninitialized databases.

libs/checkpoint-postgres/src/index.ts
```diff
@@ -132,19 +132,27 @@ export class PostgresSaver extends BaseCheckpointSaver {
...
       try {
         const result = await client.query(
           `SELECT v FROM ${SCHEMA_TABLES.checkpoint_migrations} ORDER BY v DESC LIMIT 1`
         );
         if (result.rows.length > 0) {
           version = result.rows[0].v;
         }
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      } catch (error: any) {
+      } catch (error: unknown) {
```
benjamincburns, line 148: The original intention behind this logic was to ensure that we'd only attempt to execute migrations if they needed to be executed, as you'd expect this to be a No-op otherwise. The query in question should be benign, but you can imagine how someone looking at a query log from this might be a bit surprised that it attempts to execute DDL operations when none are necessary.

I suspect that what's actually happening in #903 is that the error checking in this catch block isn't working as intended, probably because the string match is failing in certain cases. We should probably just check for `/relation "[^"]+" does not exist/` or similar.

mattia-consiglio:
Hi @benjamincburns,

I've added a test to demonstrate that the database initialization wasn't functioning correctly. Specifically, I included a `SELECT` query to verify the existence of the `checkpoint_migrations` table before executing the first migration.

The issue was with the following code:

```typescript
const result = await client.query(
     `SELECT v FROM ${SCHEMA_TABLES.checkpoint_migrations} ORDER BY v DESC LIMIT 1`
);
```

This code was entering the `catch` block because it attempted to query the `checkpoint_migrations` table before it was created.

benjamincburns:
> The issue was with the following code:
> 
> ```ts
> const result = await client.query(
>      `SELECT v FROM ${SCHEMA_TABLES.checkpoint_migrations} ORDER BY v DESC LIMIT 1`
> );
> ```
> 
> This code was entering the `catch` block because it attempted to query the `checkpoint_migrations` table before it was created.

Hi @mattia-consiglio - as I said in my comment [here](https://github.com/langchain-ai/langgraphjs/pull/904/files/73d4b27127ae437800227bd6f8d00132fd5b6e32#r1976772520) branching to the `catch` block here was an intended part of the original design.

The issue as I see it is that the error check that sets the `version` to `-1` (indicating that the database is brand new and that migration should therefore proceed) was broken in some cases upon introduction of the schema name feature. As I said in that comment, to avoid running DDL statements every time that `setup` is called, I'd prefer a solution that checks the postgres error code rather than relying on a string match for the error message. Would you be willing to rework this PR to address this issue that way?

benjamincburns (changes requested):
Setting review status so other maintainers don't accidentally merge this. Please see comments elsewhere in PR for requested changes.

mattia-consiglio:
> Would you be willing to rework this PR to address this issue that way?

I'll try to use the `catch` approach using the error code in this week

mattia-consiglio:
Hi @benjamincburns,

I've done some changes:
- I removed the duplicate checks of the table existence, and check the error code instead of the message.
- I changed the test to work accordingly to the new code.
- I also changed the `tsconfig.json` module to `NodeNext` since I had TypeScript reporting that the module should be `NodeNext` if the `moduleResolution` is `nodenext`

libs/checkpoint-postgres/src/index.ts
```diff
@@ -250,7 +252,7 @@ export class PostgresSaver extends BaseCheckpointSaver {
       pending_sends: [],
     };
     if ("channel_values" in serialized) {
-      delete serialized.channel_values;
```
benjamincburns, line 253: I believe that we actually do want `delete` here - otherwise the key is still in the object with a value of `undefined` and it gets serialized as `null`.

benjamincburns (approved):
Thanks for the updates - this looks good now apart from one minor tweak to the `.gitignore` which I'll handle myself.

PR #906, opened 2025-02-28
fix(docs): Add section to demonstrate saving a graph diagram locally

nick-w-nick:
Inspired by the below conversation in the LangChain Community Slack, I have added an example to the Quickstart documentation page for how to save the mermaid representation of the graph to a local PNG file.

image

I also noticed a minor typo in the how-to breakpoints doc which I fixed while I was in the area.

Thanks!


benjamincburns:
Thanks, @nick-w-nick!

PR #930, opened 2025-03-03
WIP DRAFT - feat(langgraph): add `exec` function 

benjamincburns:
This is very draft, and naming is very placeholder, TBD.

Goals:

- Super tightly constrained types for streaming events
  - Make it so consumers know exactly what runtime checks they _actually_ need and don't feel compelled to implement runtime checks for unioned types that are never emitted
- Gives us an option to get ahead of the "combinatorial explosion" that comes from having multiple ways to invoke the graph
  - Single function for configuring the graph, single function for invocation that works regardless of whether the user wants to `await` a `Promise` or consume a stream/generator/event emitter.
  - Importantly, these are separate functions, which opens the possibility to pass multiple graph input args in the future on invocation

TODO:

- [ ] Way more test cases
- [ ] Subgraphs
- [ ] Further Update type refinements
  - [ ] Can we get the type system to be aware of what each node _actually_ returns so the type of each node key in the update event matches that rather than the update type for the entire schema?
  - [ ] Can we get the type system to detect subgraphs so we can type `updates` event keys correctly?
- [ ] More intentional design for resume (rn it's passed in place of the input as before - do we want to keep doing that?)
- [ ] Potentially emit events for interrupt, error, etc?
  - Right now the promise rejects, and I don't think we have an explicit `error` event
- [ ] Potentially buffer events so they can be inspected post `await`


benjamincburns:
Closing - this is stale and requires a different approach, anyway.

PR #942, opened 2025-03-04
fix(docs): Replace "Python" with "JavaScript" in swarm README

nick-w-nick:
Now that `langgraph-swarm` has been ported to LangGraphJS, I thought I'd take a quick look at it and noticed a minor typo likely due to a find and replace missing it due to the casing.

Thought I'd fix it up real quick.

Thanks!


benjamincburns:
Good looking out, @nick-w-nick!

PR #981, opened 2025-03-11
chore(langgraph): add tests to validate AbortSignal is passed into nodes

benjamincburns:
Also validates that

- The signal received by nodes
  - fires in the event of a timeout
  - fires in the event of an external trigger (user code calling `abort` on the associated `AbortController`)
  - propagates correctly from parent graphs into subgraphs
    - when subgraph is called from within a node w/o explicitly passing `config`
    - when subgraph is a node

### DIFF
--- a/libs/langgraph/src/tests/pregel.test.ts
+++ b/libs/langgraph/src/tests/pregel.test.ts
@@ -9708,47 +9708,6 @@ graph TD;
     expect(thirdState.tasks).toHaveLength(0);
   });
 
-  it("should cancel when signal is aborted", async () => {
-    let oneCount = 0;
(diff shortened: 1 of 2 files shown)


libs/langgraph/src/tests/pregel.test.ts
```diff
@@ -9708,47 +9708,6 @@ graph TD;
     expect(thirdState.tasks).toHaveLength(0);
   });
 
-  it("should cancel when signal is aborted", async () => {
```
benjamincburns, line 9711: Moved to new file because this one is massive and I'm actively splitting it up in another branch.

PR #994, opened 2025-03-14
Fix handoff tool name for multiple whitespaces

ramansah:
Fix: Handoff Tool function name to replace all whitespace characters
Before: Bob The Builder -> bob_the builder
After: Bob The Builder -> bob_the_builder


benjamincburns:
Thanks for this, @ramansah!

By any chance would you be willing to add a test that demonstrates the need for this fix? That will help us to avoid regressions around this in the future.

Thanks!

benjamincburns:
Thanks for the contribution, @ramansah.

PR #998, opened 2025-03-17
perf(checkpoint-sqlite): reuse `getTuple` prepared statements

littledivy:
Prepare SQL statement once and reuse in `getTuple`. This avoids parsing and query compilation overhead.


benjamincburns (approved):
Nice!

PR #1003, opened 2025-03-18
feat(prebuilt-cua): Prebuilt computer use agent

bracesproul:
todo for launch:
- [ ] update readme to not use message classes
- [ ] ensure docs are 1:1 port with py
- [ ] release 0.0.1
- [ ] figure out serde bug?


libs/langgraph-cua/src/nodes/call-model.ts
```diff
@@ -0,0 +1,76 @@
...
+      },
+    ])
+    .bind({
+      truncation: "auto",
+      previous_response_id: previousResponseId,
+    });
+
+  let response: AIMessageChunk;
+  if (state.computerCallOutput) {
+    // TODO: How to pass back computer call outputs?
```
bracesproul, line 62: @dqbd

PR #1007, opened 2025-03-18
feat(pregel): add ability to bulk update as input

dqbd:
- Add option to submit an update as input, creates a `source: input` checkpoint that will contain `__start__` task, which can be populated with a second update
- Also adds a test for expected usage + map reduce patterns

### DIFF
--- a/libs/langgraph/src/pregel/index.ts
+++ b/libs/langgraph/src/pregel/index.ts
@@ -891,6 +891,8 @@ export class Pregel<
    * from a list of updates, especially if a checkpoint
    * is created as a result of multiple tasks.
    *
+   * @internal The API might change in the future.
+   *
    * @param startConfig - Configuration for the update
    * @param updates - The list of updates to apply to graph state
    * @returns Updated configuration
--- a/libs/langgraph/src/pregel/index.ts
+++ b/libs/langgraph/src/pregel/index.ts
@@ -1129,7 +1129,6 @@ export class Pregel<
         }
 
         // apply to checkpoint
-        // TODO: Why does keyof StrRecord allow number and symbol?
         _applyWrites(
           checkpoint,
           channels,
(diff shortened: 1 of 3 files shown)


libs/langgraph/src/pregel/index.ts
```diff
@@ -1107,6 +1107,70 @@ export class Pregel<
           saved ? saved.metadata : undefined
         );
       }
+
+      if (asNode === "__input__") {
```
benjamincburns, line 1111: good idea to use the values in `constants.ts` for this:

```suggestion
      if (asNode === INPUT) {
```

might as well get the other magic strings elsewhere in this method as well.

benjamincburns (commented):
I don't think I'm able to review this effectively for correctness without doing a bunch of comparative testing, so I'm not blocking this - but also I'm not able to approve it, either. I do have a few questions/concerns, however.

- Why doesn't `bulkUpdateState` just proxy to `updateState`?
  - I'm assuming it's for performance reasons? If so, how big is the speed up vs calling `updateState` in a loop?
- If the process fails midway through, I assume there's no rollback?
  - If so, is there any way to be be more proactive with input validation?
- This method is >500 LOC with short-circuit returns, exceptions thrown, etc. It would be much nicer if the logic could be broken down into more manageable chunks so it's easier to grok (and test). I know that a lot of code in LangGraph is bulky like this, but it'll just keep getting less and less readable unless we make a change
  - How much of this logic is duplicated from other places? Can we extract those pieces and reuse, instead?
- Are any users asking for this feature, or is this something we're doing to enable other features? If the latter, perhaps we should consider marking this `@internal` until users request it.
  - Not opposed to it being public if users want it, it'd just be nice to retain some flexibility, if not.

If we do want it to be public, I think it needs a lot more testing, especially for negative test cases. Given how crucial checkpointing is, I think it save us a bunch of headaches if we targeted 100% branch coverage for the logic in `bulkUpdateState`.

libs/langgraph/src/pregel/index.ts
dqbd: The main reason why `bulkUpdateState` exists is that `updateState` only supports accepting a single update for checkpoint, which won't work for map-reduce use cases. It's currently used only internally, so I'm fine with marking this with `@internal` and/or hiding this from public API.

Haven't attempted to break down the method in the previous PR (https://github.com/langchain-ai/langgraphjs/pull/969), instead tried to make as few changes as possible. Will try to do so in a different PR.

PR #1022, opened 2025-03-24
fix(langgraph): compose & propagate AbortSignal correctly, trigger on more early termination cases

### LINKED ISSUE #319: Abort signal doesn't work
When I try using an abort signal with my graph, it doesn't work; It calls the graph but then never het's past the __start__ node. 
🥲 I try adding the signal to the config after compiling my graph; I'm not sure if that makes a difference here.

benjamincburns:
The `AbortSignal` passed to each node via the `signal` field on the node's `config` argument now aborts when:

- The `AbortSignal` passed into `Pregel.invoke`, `Pregel.stream`, or `Pregel.streamEvents` via `config.signal` is aborted.
- The currently executing graph super-step times out.
- Any other node executing concurrently within the same superstep throws an `Error`.
  - Note: "bubble-up" errors (e.g. `NodeInterrupt` and `CommandParent`) will _not_ trigger an abort.
- The `cancel` method is called on the stream returned by `Pregel.stream` or `Pregel.streamEvents`.

Importantly, `AbortSignal` are also now composed intelligently throughout the graph execution, allowing for a node in a parent graph to call a subgraph with an `AbortSignal` that originates in the node, and the signal received by the subgraph nodes will abort when `abort()` is called on the associated `AbortController`, as well as in any of the cases mentioned above, including when the parent graph's external `AbortSignal` is aborted.

fixes #319

### DIFF
--- a/libs/langgraph/src/pregel/index.ts
+++ b/libs/langgraph/src/pregel/index.ts
@@ -87,6 +87,7 @@ import {
 } from "./algo.js";
 import {
   _coerceToDict,
+  combineAbortSignals,
   getNewChannelVersions,
   patchCheckpointMap,
   RetryPolicy,
@@ -109,7 +110,10 @@ import {
 import { LangGraphRunnableConfig } from "./runnable_types.js";
 import { StreamMessagesHandler } from "./messages.js";
 import { PregelRunner } from "./runner.js";
-import { IterableReadableWritableStream } from "./stream.js";
+import {
+  IterableReadableStreamWithAbortSignal,
+  IterableReadableWritableStream,
+} from "./stream.js";
 
 type WriteValue = Runnable | RunnableFunc<unknown, unknown> | unknown;
 type StreamEventsOptions = Parameters<Runnable["streamEvents"]>[2];
@@ -1647,13 +1651,25 @@ export class Pregel<
     // There is currently no way in _streamIterator to determine whether this was
     // set by by ensureConfig or manually by the user, so we specify the bound value here
     // and override if it is passed as an explicit param in `options`.
+    const abortController = new AbortController();
+
     const config = {
       recursionLimit: this.config?.recursionLimit,
       ...options,
+      signal: options?.signal
+        ? combineAbortSignals(options.signal, abortController.signal)
+        : abortController.signal,
     };
-    return super.stream(input, config);
+
+    return new IterableReadableStreamWithAbortSignal(
+      await super.stream(input, config),
+      abortController
+    );
   }
 
+  /**
+   * @inheritdoc
+   */
   override streamEvents(
     input: InputType | Command | null,
     options: Partial<PregelOptions<Nodes, Channels, ConfigurableFieldType>> & {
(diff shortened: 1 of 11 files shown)


libs/langgraph/src/pregel/index.ts
```diff
@@ -1768,7 +1828,7 @@ export class Pregel<
       checkpointer,
       store,
       streamModeSingle,
-    ] = this._defaults(restConfig);
+    ] = defaults;
```
benjamincburns, line 1831: Whoops, need to revert this.

libs/langgraph/src/pregel/retry.ts
```diff
@@ -80,6 +82,12 @@ export async function _runWithRetry<
   if (configurable) {
     config = patchConfigurable(config, configurable);
   }
+
+  config = {
+    ...config,
+    signal,
+  };
```
benjamincburns, line 89: Dunno why I didn't just do `config.signal = signal;` 🤦

libs/langgraph/src/pregel/runner.ts
```diff
@@ -77,31 +87,38 @@ export class PregelRunner {
...
       if (isGraphInterrupt(error)) {
         graphBubbleUp = error;
       } else if (isGraphBubbleUp(error) && !isGraphInterrupt(graphBubbleUp)) {
         graphBubbleUp = error;
-      } else if (error) {
+      } else if (error && (nodeErrors.size === 0 || !signalAborted)) {
+        // trigger task cancellation due to error
+        exceptionSignalController.abort();
         nodeErrors.add(error);
       }
```
benjamincburns, line 123: The goal here is to capture the exception that causes us to abort. In theory it's possible for multiple nodes to throw, so this also handles the edge case of capturing concurrent exceptions thrown before the node saw an abort.

Also if an external abort triggers a node to throw an `AbortError` or similar, we want to propagate that back out to the caller, so we always grab the first non-bubble-up error here, even if it occurred after the signal was aborted. Otherwise in the event that this happens as a result of some other error being thrown, we don't bother keeping the `AbortError` or whatever, as it's not causal and therefore unimportant to the caller.

libs/langgraph/src/pregel/runner.ts
```diff
@@ -131,6 +148,79 @@ export class PregelRunner {
...
+      : // Otherwise, just keep using the external abort signal, or initialize it if it hasn't been
+        // assigned yet
+        previousSignals.externalAbortSignal ?? signal;
+
+    const errorAbortSignal = previousSignals.errorAbortSignal
+      ? combineAbortSignals(
+          previousSignals.errorAbortSignal!,
+          exceptionSignalController.signal
+        )
+      : exceptionSignalController.signal;
```
benjamincburns, line 198: Probably also worth a comment - chaining here rather than always using a fresh one handles the case where a subgraph is called in a parallel branch to some other node in the parent graph. Honestly I'm not even sure that we support that, but if we do, we propagate the `AbortSignal` correctly in that case now.

libs/langgraph/src/pregel/runner.ts
```diff
@@ -333,10 +409,15 @@ export class PregelRunner {
...
+    const abortPromise = new Promise<never>((_resolve, reject) => {
       listener = () => reject(new Error("Abort"));
-      signal?.addEventListener("abort", listener);
-    }).finally(() => signal?.removeEventListener("abort", listener));
+      signals?.timeoutAbortSignal?.addEventListener("abort", listener);
+      signals?.externalAbortSignal?.addEventListener("abort", listener);
+    }).finally(() => {
+      signals?.timeoutAbortSignal?.removeEventListener("abort", listener);
+      signals?.externalAbortSignal?.removeEventListener("abort", listener);
+    });
```
benjamincburns, line 420: could've just done `{ once: true }` as the 3rd arg here, but I wanted to avoid a double call to `reject`.

benjamincburns: Ah, although I could compose them and then ditch the `finally` block - I'll do that instead.

libs/langgraph/src/pregel/stream.ts
```diff
@@ -4,6 +4,50 @@ import { StreamMode } from "./types.js";
 // [namespace, streamMode, payload]
 export type StreamChunk = [string[], StreamMode, unknown];
 
+export class IterableReadableStreamWithAbortSignal<
```
benjamincburns, line 7: Note to self: typedoc this.

PR #1029, opened 2025-03-25
feat(pregel): add zod plugin

dqbd:
Uses the proposed mechanism for Zod plugins: https://github.com/colinhacks/zod/pull/3445

Example usage:

```
const schema = z.object({
  list: z.array(z.string).default(() => ["default"])
    .langgraph.reducer(
      (a, b) => [...a, ...(Array.isArray(b) ? b : [b])],
      z.union([z.array(z.string()), z.string()])
    )
    .langgraph.metadata({ ... })
})

```


libs/langgraph/src/graph/zod/plugin.ts
```diff
@@ -0,0 +1,73 @@
...
+          metadata(jsonSchemaExtra) {
+            extendMeta(zodThis, (meta) => ({
+              ...meta,
+              jsonSchemaExtra,
+            }));
+
+            return zodThis;
+          },
+          reducer<Input>(
+            transform: (a: Output, arg: Input) => Output,
```
dqbd, line 48: Tried sticking the transform function to the options object, but for some reason Typescript will freak out and break all Zod types.

PR #1035, opened 2025-03-25
fix(langgraph): callbacks not passed in streamEvents

dqbd:


jacoblee93 (approved):
Thanks! Great catch

Alternatively should we just merge callbacks in `ensureLangGraphConfig`?

PR #1050, opened 2025-03-27
langgraph: include full message history when generating structured response in createReactAgent

vbarda:


benjamincburns (approved):
Thanks for the quick turnaround!

PR #1052, opened 2025-03-27
feat(cua): add hyperbrowser provider for cua agent

NikhilShahi:
This PR adds Hyperbrowser as an optional provider for using with LangGraph and CUA. It uses [Hyperbrowser](https://docs.hyperbrowser.ai/) browser session and playwright to allow CUA to interact with the browser page. Scrapybara is kept as the default provider.

### DIFF
--- a/libs/langgraph-cua/src/index.ts
+++ b/libs/langgraph-cua/src/index.ts
@@ -15,6 +15,7 @@ import {
   CUAAnnotation,
   CUAConfigurable,
   CUAUpdate,
+  Provider,
 } from "./types.js";
 import { getToolOutputs, isComputerCallToolMessage } from "./utils.js";
 
@@ -64,11 +65,31 @@ interface CreateCuaParams<
   StateModifier extends AnnotationRoot<any> = typeof CUAAnnotation
 > {
   /**
+   * The provider to use for the browser instance.
+   * @default "scrapybara"
+   */
+  provider?: Provider;
+
+  /**
    * The API key to use for Scrapybara.
    * This can be provided in the configuration, or set as an environment variable (SCRAPYBARA_API_KEY).
    * @default process.env.SCRAPYBARA_API_KEY
    */
   scrapybaraApiKey?: string;
+
+  /**
+   * The API key to use for Hyperbrowser.
+   * This can be provided in the configuration, or set as an environment variable (HYPERBROWSER_API_KEY).
+   * @default process.env.HYPERBROWSER_API_KEY
+   */
+  hyperbrowserApiKey?: string;
+
+  /**
+   * Parameters to use for configuring the Hyperbrowser session, such as proxy usage, screen dimensions, etc.
+   * For more information on the available parameters, see the [Hyperbrowser API documentation](https://docs.hyperbrowser.ai/sessions/overview/session-parameters).
+   * @default undefined
+   */
+  sessionParams?: Record<string, unknown>;
 
   /**
    * The number of hours to keep the virtual machine running before it times out.
@@ -145,7 +166,10 @@ export function createCua<
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   StateModifier extends AnnotationRoot<any> = typeof CUAAnnotation
 >({
+  provider = "scrapybara",
   scrapybaraApiKey,
+  hyperbrowserApiKey,
+  sessionParams,
   timeoutHours = 1.0,
   zdrEnabled = false,
   recursionLimit = 100,
@@ -195,7 +219,10 @@ export function createCua<
   // Configure the graph with the provided parameters
   const configuredGraph = cuaGraph.withConfig({
     configurable: {
+      provider,
       scrapybaraApiKey,
+      hyperbrowserApiKey,
+      sessionParams,
       timeoutHours,
       zdrEnabled,
       authStateId,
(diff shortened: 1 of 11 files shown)


benjamincburns:
Hi @NikhilShahi - thanks for the contribution!

It's great to offer freedom of choice for underlying components like this. However my concern with this specific change would be that it forces the user to install dependencies for _all_ providers rather than installing only the packages that they need to support their desired configuration. This sort of extensibility may be something that @bracesproul has already considered, so I'm afraid that I'll need to defer to him on this.

In the meantime, if you can turn this PR into one that allows the user to bring any arbitrary provider without further modification to the code, that would be much easier to accept.

libs/langgraph-cua/src/nodes/take-browser-action.ts
```diff
@@ -0,0 +1,240 @@
...
+      `\n\nFailed to execute computer call: ${actionType}\n\n`,
+      error
+    );
+    console.error(`Computer call details: ${output}`);
+    computerCallToolMsg = {
+      type: "tool",
+      tool_call_id: call_id,
+      content: `data:image/jpeg;base64,${DUMMY_SCREENSHOT}`,
+      additional_kwargs: { type: "computer_call_output", status: "incomplete" },
+    };
```
bracesproul, line 232: I don't think we want to return a computer call if the execution fails. This would be a change in behavior

libs/langgraph-cua/src/nodes/take-browser-action.ts
```diff
@@ -0,0 +1,240 @@
...
+const translateKey = (key: string): string => {
+  const lowerKey = key.toLowerCase();
+  return lowerKey in CUA_KEY_TO_PLAYWRIGHT_KEY
+    ? CUA_KEY_TO_PLAYWRIGHT_KEY[
+        lowerKey as keyof typeof CUA_KEY_TO_PLAYWRIGHT_KEY
+      ]
+    : key;
+};
+
+export async function takeHyperbrowserAction(
```
bracesproul, line 52: I think it would be cleaner to have a single `takeComputerAction` function, which then has a switch case for each action type like before, and the switch case calls a function which it passes the action, and selected provider. this way we can abstract the specific action code to these utils

bracesproul: If you need additional setup for the provider before iterating over the actions to execute it, I would move that to a util as well. this should make the take computer action func cleaner, and easier to maintain

libs/langgraph-cua/src/nodes/take-browser-action.ts
```diff
@@ -0,0 +1,240 @@
...
+        }
+        for (const key of [...mappedKeys].reverse()) {
+          await page.keyboard.up(key);
+        }
+        break;
+      }
+
+      case "type": {
+        const { text } = action;
+        // console.log(`Action: type text '${text}'`);
```
bracesproul, line 154: all these console logs should be dropped

libs/langgraph-cua/src/types.ts
```diff
@@ -47,9 +50,26 @@ export const CUAAnnotation = Annotation.Root({
     reducer: (_state, update) => update,
     default: () => undefined,
   }),
+  /**
+   * The state of the browser instance.
+   */
+  browserState: Annotation<
+    { browser: Browser | undefined; currentPage: Page | undefined } | undefined
```
bracesproul, line 57: If `Browser` and `Page` are not JSON searlizable objects, they can't be stored in state.

libs/langgraph-cua/src/index.ts
```diff
@@ -63,13 +64,33 @@ interface CreateCuaParams<
...
+   * @default process.env.HYPERBROWSER_API_KEY
+   */
+  hyperbrowserApiKey?: string;
+
+  /**
+   * Parameters to use for configuring the Hyperbrowser session, such as proxy usage, screen dimensions, etc.
+   * For more information on the available parameters, see the [Hyperbrowser API documentation](https://docs.hyperbrowser.ai/sessions/overview/session-parameters).
+   * @default undefined
+   */
+  sessionParams?: Record<string, unknown>;
```
bracesproul, line 92: Can we type this?

libs/langgraph-cua/src/types.ts
```diff
@@ -128,9 +173,14 @@ export function getConfigurationWithDefaults(
...
 ): typeof CUAConfigurable.State {
   return {
+    provider: config.configurable?.provider ?? "scrapybara",
     scrapybaraApiKey:
       config.configurable?.scrapybaraApiKey ||
       getEnvironmentVariable("SCRAPYBARA_API_KEY"),
+    hyperbrowserApiKey:
+      config.configurable?.hyperbrowserApiKey ||
+      getEnvironmentVariable("HYPERBROWSER_API_KEY"),
+    sessionParams: config.configurable?.sessionParams ?? {},
```
bracesproul, line 183: default to undefined if not set
```suggestion
    sessionParams: config.configurable?.sessionParams ?? undefined,
```

bracesproul (changes requested):
some feedback, very good start. thanks for putting this PR up!

NikhilShahi:
Hey @bracesproul , thanks for the feedback. I think I have implemented all the changes requested, let me know if there is anything else I should change.

dqbd:
Hello! Closing this PR for now, since we're not focusing on adding different CUA providers for now (and we're re-thinking CUA to be handled as a middleware).

PR #1103, opened 2025-04-14
fix: Never wrap async batch store with async batch store

bracesproul:


benjamincburns (dismissed):
LGTM, but please add a simple unit test for it prior to merging.

Maybe just make `store` protected and then extend `AsyncBatchedStore` in the test to add a `getStore` method or something like that? Could also just do `// @ts-expect-error`, but it'd be nice if the compiler complained if we ever break the interface.

PR #1115, opened 2025-04-17
chore(langgraph): use vitest for unit tests

littledivy:
Migrate unit tests to vitest and add `yarn test:browser` using Vitest's browser mode.

Extracted from #999


benjamincburns:
After rebasing there were yarn lockfile issues and then build errors. I managed to get it building, but then based on the test failures I was seeing it seemed that ALS was broken. Upon inspection, it looked like `async_hooks` was only being imported in the CJS build, so I added a CJS-specific `src/setup/async_local_storage.cts` file, tweaked some `tsconfig` settings, and reverted the change in the `.ts` (ESM-only). This got things building, but the tests are still failing, and I'm not sure why.

Didn't expect this to require quite so much effort to merge and I don't want to keep sinking time into it w/o a better understanding of what's left, so I've put this on the back burner for now.

benjamincburns:
Realized I'd made a mistake when splitting `src/setup/async_local_storage.ts` into ESM/CJS files but after fixing that issue the tests still fail as before.

PR #1129, opened 2025-04-24
docs: Adds additional Google Analytics tag to docs

jacoblee93:
CC @danielrlambert3

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -440,7 +440,9 @@ extra:
           data: 0
           note: >-
             Thanks for your feedback! Please help us improve this page by adding to the discussion below.
-
+  shared_analytics:
+    provider: google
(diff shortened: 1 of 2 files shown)


benjamincburns:
Did you test this @jacoblee93? I don't think that the Material for MkDocs will work as expected with two GA providers, as the code creates properties on the global `Window` object (e.g. `gTag`). Supporting this might take a bit more effort than simply enabling it in the config.

Also bear in mind that the mkdocs has zero input validation for plugin config, so it building/running locally isn't really any confirmation that it's working as intended. You'd have to validate with Tag Assistant or similar.

jacoblee93:
Danny said it's good, ran locally and saw two requests that looked right being sent

benjamincburns (approved):

PR #1167, opened 2025-05-09
fix(docs): typos

omahs:
fix(docs) typos


dqbd:
Thank you @omahs! LGTM

PR #1169, opened 2025-05-09
Implement updateState for entrypoint/task

nfcampos:
Fixes # (issue)

### DIFF
--- a/libs/langgraph/src/pregel/index.ts
+++ b/libs/langgraph/src/pregel/index.ts
@@ -1318,7 +1318,10 @@ export class Pregel<
       if (updates.length === 1) {
         // eslint-disable-next-line prefer-const
         let { values, asNode } = updates[0];
-        if (asNode === undefined && nonNullVersion === undefined) {
+        if (asNode === undefined && Object.keys(this.nodes).length === 1) {
+          // if only one node, use it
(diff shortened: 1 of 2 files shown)


libs/langgraph/src/pregel/index.ts
```diff
@@ -1318,7 +1318,10 @@ export class Pregel<
       if (updates.length === 1) {
         // eslint-disable-next-line prefer-const
         let { values, asNode } = updates[0];
-        if (asNode === undefined && nonNullVersion === undefined) {
+        if (asNode === undefined && Object.keys(this.nodes).length === 1) {
```
hinthornw, line 1321: ubernit but personal pref is one asNode === undefined` and then conditions (Object.keys(this.nodes).length === 1, else if nonNullVersion === undefined) else

hinthornw (approved):

libs/langgraph/src/pregel/index.ts
benjamincburns: I tend to agree, but I think for this particular code it's probably better to keep it inlined - the nesting complexity is already really high and I find it hard to keep track of conditions that applied in some block 100+ lines up when I'm reading a bunch of conditional logic like this

PR #1174, opened 2025-05-12
feat(langgraph): add option to inline agent name in createReactAgent

benjamincburns:


vbarda (approved):

thinhnguyennt7 (approved):

benjamincburns:
Docs build appears to be broken due to new cloud docs pages that have relative links that don't work on the JS side. Will wait until that's fixed up to merge.

PR #1188, opened 2025-05-16
supervisor: add forward message tool

vbarda:


libs/langgraph-supervisor/src/tests/supervisor.test.ts
```diff
@@ -306,3 +307,171 @@ describe("Test supervisor basic workflow", () => {
...
+      agents: [echoAgent],
+      llm: supervisorModel,
+      tools: [forwardingTool],
+      prompt: "You are a team supervisor managing an echo agent.",
+    });
+
+    const app = workflow.compile();
+    expect(app).toBeDefined();
+
+    const result = await app.invoke({
```
vbarda, line 383: need to fix the test here

PR #1203, opened 2025-05-21
feat: `MessagesZodSchema`

hntrl:
Adds a zod-based graph state utility that mimics `MessagesAnnotation`

### DIFF
--- a/libs/langgraph/src/graph/messages_annotation.ts
+++ b/libs/langgraph/src/graph/messages_annotation.ts
@@ -70,6 +70,7 @@ export const MessagesAnnotation = Annotation.Root({
  * ```ts
  * import { BaseMessage } from "@langchain/core/messages";
  * import { Annotation, StateGraph, messagesStateReducer } from "@langchain/langgraph";
+ * import "@langchain/langgraph/zod";
  *
  * const schema = z.object({
  *   messages: z
--- a/libs/langgraph/src/graph/messages_annotation.ts
+++ b/libs/langgraph/src/graph/messages_annotation.ts
@@ -68,25 +68,23 @@ export const MessagesAnnotation = Annotation.Root({
  *
  * @example
  * ```ts
- * import { BaseMessage } from "@langchain/core/messages";
- * import { Annotation, StateGraph, messagesStateReducer } from "@langchain/langgraph";
+ * import { z } from "zod";
+ * import type { BaseMessage, BaseMessageLike } from "@langchain/core/messages";
+ * import { StateGraph, messagesStateReducer } from "@langchain/langgraph";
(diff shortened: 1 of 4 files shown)


libs/langgraph/src/graph/messages_annotation.ts
```diff
@@ -45,3 +47,69 @@ export const MessagesAnnotation = Annotation.Root({
...
+ * const schema = z.object({
+ *   messages: z
+ *     .array(z.string())
+ *     .default(() => [])
+ *     .langgraph.reducer(
+ *        messagesStateReducer,
+ *        z.union([z.string(), z.array(z.string())])
+ *     ),
+ * });
+ * export const StateAnnotation = Annotation.Root({
```
dqbd, line 83: The example is incorrect, would just use the same schema as the code

libs/langgraph/src/graph/messages_annotation.ts
```diff
@@ -45,3 +47,69 @@ export const MessagesAnnotation = Annotation.Root({
...
+ *  .addNode(...)
+ *  ...
+ * ```
+ */
+export const MessagesZodState = z.object({
+  messages: withLangGraph(z.custom<BaseMessage[]>(), {
+    reducer: {
+      schema: z.union([
+        z.custom<BaseMessageLike>(),
+        z.array(z.custom<BaseMessageLike>()),
```
dqbd, line 109: Should we do `z.custom<Messages>` directly instead, given the presence of `z.custom`?

libs/langgraph/src/graph/messages_annotation.ts
```diff
@@ -45,3 +47,69 @@ export const MessagesAnnotation = Annotation.Root({
...
+ * const schema = MessagesZodState.and(
+ *   z.object({ count: z.number() }),
+ * );
+ *
+ * const graph = new StateGraph(schema)
+ *  .addNode(...)
+ *  ...
+ * ```
+ */
+export const MessagesZodState = z.object({
```
dqbd, line 104: Worth adding tests similar to ones found in `prebuilt.test.ts`?

libs/langgraph/src/graph/messages_annotation.ts
hntrl: I'm pulling the example from [here](https://langchain-ai.github.io/langgraphjs/how-tos/define-state/#using-zod). I had to switch over to using `withLangGraph` to add reducer meta instead of using the prototype plugin since it was causing some odd behavior in testing if I had a top level import . I wanted to keep this example since `.langgraph.reducer` since I didn't see `withLangGraph` having dedicated docs like the zod plugin does.

dqbd (approved):

PR #1238, opened 2025-05-29
fix(langgraph): propagate recursionLimit and signal in RemoteGraph

### LINKED ISSUE #1237: Bug: RemoteGraph strips config properties like `recursionLimit`
## Behavior
When calling `remoteGraph.invoke(input, { recursionLimit: 30 })`, the recursion limit should be set to 30 and any recursion error should reference this limit.
Currently the `recursionLimit` is ignored and the default (25) is being used instead.

## Reproduction
```typescript
import { RemoteGraph } from "@langchain/langgraph/remote";

const client = new Client({ apiUrl: "http://localhost:8123" });
const graph = new RemoteGraph({
  graphId: "infinite-graph", 
  client: client,
});

// This should use recursionLimit: 30, but uses default instead
await graph.invoke({}, { recursionLimit: 30 });
```

## Root Cause
The `RemoteGraph._sanitizeConfig` method only returns a subset of config properties:

```typescript
export class RemoteGraph {
   // ...

  override async *_streamIterator(
      input: PregelInputType,
      options?: Partial<PregelOptions<Nn, Cc, ConfigurableFieldType>>
    ): AsyncGenerator<PregelOutputType> {
    const mergedConfig = mergeConfigs(this.config, options);
    // config is stripped here:
    const sanitizedConfig = this._sanitizeConfig(mergedConfig);
    // ...
    for await (const chunk of this.client.runs.stream(
      sanitizedConfig.configurable.thread_id as string,
      this.graphId,
      {
        config: sanitizedConfig,
        // ...
      }
    )) 
    // ...
  }

  protected _sanitizeConfig(config: RunnableConfig) {
    // ... sanitization logic ...
    
    return {
      tags: sanitizedConfig.tags ?? [],
      metadata: sanitizedConfig.metadata ?? {},
      configurable: newConfigurable,
  
      /* 
      missing:
        recursion_limit: sanitizedConfig.recursionLimit,
      */
    };
  }
}
```

---
I will submit a pull request with the fix to include the missing config properties in the `_sanitizeConfig` method return value.

xavierroma:
Fixes #1237

### DIFF
--- a/libs/langgraph/src/pregel/remote.ts
+++ b/libs/langgraph/src/pregel/remote.ts
@@ -262,6 +262,7 @@ export class RemoteGraph<
       tags: sanitizedConfig.tags ?? [],
       metadata: sanitizedConfig.metadata ?? {},
       configurable: newConfigurable,
+      recursion_limit: sanitizedConfig.recursionLimit,
     };
   }
 
@@ -435,6 +436,7 @@ export class RemoteGraph<
         interruptAfter: interruptAfter as string[],
         streamSubgraphs,
         ifNotExists: "create",
+        signal: mergedConfig.signal,
       }
     )) {
       let mode;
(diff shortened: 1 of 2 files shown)


xavierroma:
Hi @dqbd, friendly ping on this PR. 
Let me know if there’s anything I can clarify or tweak.  
Thanks for your time!

dqbd:
Thank you so much @xavierroma!

PR #1239, opened 2025-05-29
feat(langgraph): introduce strict typing for `.stream()` methods

dqbd:
Supersedes https://github.com/langchain-ai/langgraphjs/pull/512

### DIFF
--- a/libs/langgraph/src/tests/pregel.test-d.ts
+++ b/libs/langgraph/src/tests/pregel.test-d.ts
@@ -5,6 +5,13 @@ import { StateGraph } from "../graph/state.js";
 import { Annotation } from "../graph/annotation.js";
 import { gatherIterator } from "../utils.js";
 import { StreamMode } from "../pregel/types.js";
+import { task, entrypoint } from "../func/index.js";
+import { initializeAsyncLocalStorageSingleton } from "../setup/async_local_storage.js";
+
+beforeAll(() => {
+  // Will occur naturally if user imports from main `@langchain/langgraph` endpoint.
+  initializeAsyncLocalStorageSingleton();
+});
 
 it("state graph", async () => {
   const state = Annotation.Root({
@@ -28,6 +35,8 @@ it("state graph", async () => {
     .compile();
 
   const input = { foo: "bar" };
+
+  expectTypeOf(await graph.invoke(input)).toExtend<{ foo: string[] }>();
 
   expectTypeOf(await gatherIterator(graph.stream(input))).toExtend<
     Record<"one" | "two" | "three", { foo?: string[] | string }>[]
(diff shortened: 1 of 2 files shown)


libs/langgraph/src/pregel/index.ts
```diff
@@ -1645,10 +1644,31 @@ export class Pregel<
    * @param options - Configuration options for streaming
    * @returns An async iterable stream of graph state updates
    */
-  override async stream(
+  // @ts-expect-error Return type of `stream()` differs from `invoke()`, which is expected.
```
dqbd, line 1647: Unsure if there are downsides to this, but feels like a good shortcut to make, since I don't want to wait for a @langchain/core update to have a different generic parameter just for `stream`. Open to suggestions.

libs/langgraph/src/tests/pregel.test-d.ts
```diff
@@ -0,0 +1,146 @@
...
+      two: () => ({ foo: "two" }),
+      three: () => ({ foo: "three" }),
+    })
+    .addEdge("__start__", "one")
+    .compile();
+
+  const input = { foo: "bar" };
+
+  expectTypeOf(await gatherIterator(graph.stream(input))).toExtend<
+    { foo?: string[] | string | undefined }[]
```
nfcampos, line 33: this one should be I think `{ ['one' | 'two' | 'three']: { foo?: string[] | string | undefined }}[]` ?

Diluka:
This is too complicated. I can only `as unknown as MyStreamType`

```ts
  formatSteam(stream: Awaited<ReturnType<G['stream']>>) {
    return from(stream).pipe(
      map((it) => it as unknown as MyStreamType),
```

dqbd:
Hello @Diluka! Could you share what `MyStreamType` in your case is?

Diluka:
> Hello @Diluka! Could you share what `MyStreamType` in your case is?

```ts
interface MessageMetadata {
  tags: string[];
  name?: string; 
  thread_id: string;
  langgraph_step: number;
  langgraph_node: string;
  langgraph_triggers: string[];
  langgraph_path: string[];
  langgraph_checkpoint_ns: string;
  __pregel_task_id: string;
  checkpoint_ns: string;
  ls_provider: string;
  ls_model_name: string;
  ls_model_type: string;
  ls_temperature?: number; 
  ls_max_tokens?: number; 
  ls_stop?: undefined; 
}

interface Interrupt {
  value: InterruptOptions;
  when: string; 
  resumable: boolean;
  ns: string[];
}

type StreamUpdate = {
  agent?: { messages: BaseMessageFields[] };
  supervisor?: { messages: BaseMessageFields[] };
  __interrupt__?: Interrupt;
};

type MyStreamType =
  | ['messages', [...BaseMessageFields[], MessageMetadata]]
  | ['updates', StreamUpdate];
```

PR #1242, opened 2025-05-30
Langgraph postgres store

abogacki:
# Implement PostgreSQL Store with Vector Search Capabilities

## Overview
This PR implements a PostgreSQL-based store for LangGraph.js inspired by https://github.com/langchain-ai/langgraphjs/pull/887 and [Python implementation](https://github.com/langchain-ai/langgraph/tree/main/libs/checkpoint-postgres/langgraph/store/postgres). The implementation includes comprehensive vector similarity search, hybrid search (combining vector and text search), and TTL management for automatic data expiration.

## Key Features
- Modular architecture with specialized components for different responsibilities
- Vector similarity search with support for multiple distance metrics (cosine, L2, inner product)
- Hybrid search combining vector similarity and full-text search
- Advanced filtering with rich query operators ($eq, $gt, $lt, $in, etc.)
- Automatic TTL management with configurable expiration settings
- Support for PostgreSQL's pgvector extension for efficient vector operations

## Implementation Details
- Split the monolithic implementation into specialized modules:
  - `database-core.js`: Core connection and transaction management
  - `database-setup.js`: Schema initialization and migration
  - `crud-operations.js`: Basic CRUD operations
  - `search-operations.js`: Advanced search capabilities
  - `vector-operations.js`: Vector embedding and similarity calculations
  - `ttl-manager.js`: Expiration and cleanup management
  - `query-builder.js`: SQL query construction
- Added comprehensive TypeScript types for better developer experience
- Implemented efficient text extraction from JSON documents using configurable field paths
- Added extensive test coverage for all features


int-test-deps-docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: "3.8"
 
 services:
   postgres:
-    image: postgres:latest
+    image: pgvector/pgvector:pg16
```
abogacki, line 5: I figured that this won't have any effect on other integration tests relying on this docker-compose file, and it will enable store-postgres to be tested against vector search. I've also added `docker-compose.test.yml` file in `/libs/store-postgres`

libs/store-postgres/src/index.ts
```diff
@@ -0,0 +1,392 @@
+import pg from "pg";
+import {
```
abogacki, line 2: these types are currently included in `langgraph-checkpoint` library, so I've added it as a dependency. I was kind of torn apart between copying them, or creating even separate module for `@langchain/langgraph-store` (similarily to `@langchain/langgraph-checkpoint`). Lmk which approach would you like me to go with (or include the changes yourself 🙂)

brettshollenberger: Yo nice job @abogacki! I'm not a maintainer (but invested in seeing this feature ship). In the Python repo, this lives under the `checkpoint-postgres` lib as `store/postgres`. I would think we should keep parity by putting this under the `checkpoint-postgres` submodule as well, which will keep you from needing to add this as a dependency.

abogacki: @brettshollenberger thanks for your input! I've noticed that the strucutre in the Python project is a little bit different then the one in the js implementation. 
I've introduced a separate library as @benjamincburns mentioned in the review of previous Postgres Store implementation proposal that it would be better to create it as a separate lib: https://github.com/langchain-ai/langgraphjs/pull/887#pullrequestreview-2634770279.
Imho best option would be to create implementation-agnostic library (`@langchain/langgraph-store`) that would only provide abstractions for other implementations to use 🙂. This would create cohesion with checkpoint implementation in JS. But I'm open for more suggestions :)

abogacki:
Is there a chance this will be reviewed? Is there anyone I could mention? Perhaps @dqbd

dqbd:
Hello! Thank you for the PR, will be reviewing thoroughly shortly. At first glance I think we want to have this implementation in the same package as the checkpointer

libs/store-postgres/src/index.ts
dqbd: I think for maintenance sake it'll be better to colocate the Postgres code alongside the checkpointer while keeping the entrypoint separately.

abogacki:
Awesome, thank you for applying changes, lmk if there's anything I should work on/adjust :)

libs/checkpoint-postgres/src/store/modules/database-setup.ts
```diff
@@ -0,0 +1,175 @@
...
+import { VectorIndexType, DistanceMetric } from "./types.js";
+
+/**
+ * Handles database schema creation, table setup, and index creation.
+ */
+export class DatabaseSetup {
+  constructor(private core: DatabaseCore) {}
+
+  async initialize(): Promise<void> {
+    await this.core.withClient(async (client) => {
```
dqbd: Worth adding a migration mechanism similar to the postgres checkpointer?

libs/checkpoint-postgres/src/store/modules/search-operations.ts
```diff
@@ -0,0 +1,486 @@
...
+      let sqlQuery = `
+        SELECT 
+          namespace_path, 
+          key, 
+          value, 
+          created_at, 
+          updated_at,
+          CASE 
+            WHEN $2::text IS NOT NULL THEN 
+              ts_rank(to_tsvector('english', value::text), plainto_tsquery('english', $2::text))
```
dqbd: Worth making this configurable?

libs/checkpoint-postgres/src/store/index.ts
```diff
@@ -0,0 +1,576 @@
...
+    this.core.pool.end().catch((error) => {
+      console.error("Error closing PostgreSQL pool:", error);
+    });
+    this.isClosed = true;
+  }
+
+  /**
+   * Close all database connections.
+   */
+  async end(): Promise<void> {
```
dqbd: Shouldn't we consolidate `end()` with `stop()`?

dqbd: I see, we need to make `start()` and `stop()` potentially async as well.

abogacki: Hmm, technically the `BasteStore` interface requires these methods to be synchronous. I'll apply the changes, but from what I can see, to maintain cohesion, I will:
- make `start`/`stop` asynchronous in `BaseStore` 
- I'll make `PregelLoop.initialize()` and `PregelLoop.tick()` methods await the `store.start()`
- make all other concrete implementations use async `start`/`stop`

abogacki:
@dqbd why has this been closed?

dqbd:
Ah decided to cut 0.4 without the PR for now, doing a fast follow-up. Can you reopen the PR if possible? @abogacki

abogacki:
I cannot reopen this one, but will create a new PR :)

PR #1275, opened 2025-06-11
patch: accept zod 4 types & centralize meta store

hntrl:
Adds support for zod 4 using the interoperable utilities being exported from langchain-core in https://github.com/langchain-ai/langchainjs/pull/8316 (CI will be failing here until that lands).

Also refactors the way we store schema metadata to surface a zod-compatible [schema registry](https://zod.dev/metadata)

Example usage:
```ts
import { z } from "zod/v4";
import { StateGraph } from "@langchain/langgraph";
import { registry } from "@langchain/langgraph/zod";

const graphState = z.object({
  items: z.array(z.string()).register(registry, {
    reducer: {
      schema: z.string(),
      fn: (a, b) => a.concat([b])
    }
  });

new StateGraph(graphState)
  ...
```


libs/langgraph/src/graph/zod/meta.ts
```diff
@@ -0,0 +1,290 @@
...
+  >
+    ? Schema extends InteropZodType<infer V>
+      ? ReducerSchema extends InteropZodType<infer U>
+        ? BaseChannel<V, U>
+        : never
+      : never
+    : TShape[key] extends InteropZodType<infer V, infer U>
+    ? BaseChannel<V, U>
+    : never;
+};
```
hntrl, line 41: `ZodToStateDefinition` was always assuming that the zod type was expressive of both the in/out types of the channel, but when providing a reducer (i.e. using `.langgraph.reducer` or `withLangGraph`), the reflected input type has no way of making it's way through to the channel type.

`withLangGraph` now pins the reducer schema itself to the object using `ReducedZodChannel` which gives typescript a hatch in evaluating what the actual in/out of the channel will be. The intersection type is important since it lets us use reducer-augmented schemas wherever its needed while still retaining type information.

libs/langgraph/src/graph/zod/meta.ts
hntrl, line 1: This new structure takes place of the old top-level controller orchestrated meta registry we had before in `state.ts`. The motivation for making this class based is to make it portable and usable for validation-lib specific meta stores using property inheritance (like in `zod-registry.ts` for zod/v4 meta registries)

PR #1287, opened 2025-06-16
feat: allow user defined callbacks to extend graph callbacks

lc-arjun:
Fixes # (issue)

### DIFF
--- a/libs/langgraph/src/pregel/index.ts
+++ b/libs/langgraph/src/pregel/index.ts
@@ -1808,7 +1808,7 @@ export class Pregel<
       recursionLimit: this.config?.recursionLimit,
       // Similar to `stream`, we need to pass the `config.callbacks` here,
       // otherwise the user-provided callback will get lost in `ensureLangGraphConfig`.
-      callbacks: this.config?.callbacks,
+      callbacks: [...(this.config?.callbacks ?? []), ...(options?.callbacks ?? [])],
       ...options,
       signal: options?.signal
         ? combineAbortSignals(options.signal, abortController.signal)


libs/langgraph/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@langchain/langgraph",
-  "version": "0.3.3",
```
jacoblee93: We release this via CLI script so no need to bump here

libs/langgraph/src/pregel/index.ts
```diff
@@ -1804,12 +1805,41 @@ export class Pregel<
...
     const config = {
       recursionLimit: this.config?.recursionLimit,
+      ...options,
       // Similar to `stream`, we need to pass the `config.callbacks` here,
       // otherwise the user-provided callback will get lost in `ensureLangGraphConfig`.
-      callbacks: this.config?.callbacks,
-      ...options,
+
+      // extend the callbacks with the ones from the config
+      callbacks: combineCallbacks(this.config?.callbacks, options?.callbacks),
```
jacoblee93, line 1815: Should we be adding this in other places too?

libs/langgraph/package.json
lc-arjun: oh nice i was wondering how releases work for this

libs/langgraph/src/pregel/index.ts
lc-arjun: other places being recursion limit? im not sure cc @dqbd

lc-arjun: I updated to perfer `options.recursionLimit`. abort signal is already being combined

lc-arjun: for posterity: other places means other functions like `stream`

PR #1288, opened 2025-06-17
fix(langgraph): zod 4 registry reducer schema

dqbd:
`.registry(...)` does not support modifying the `input` Zod internal, thus the input types for graphs are incorrectly inferred to output type. Goal is to augment `ZodType` with an overloaded `.registry(LanggraphZodMetaRegistry)` method, that casts the return type to proper schema.

### DIFF
--- a/libs/langgraph/src/graph/zod/zod-registry.ts
+++ b/libs/langgraph/src/graph/zod/zod-registry.ts
@@ -1,3 +1,4 @@
+import type * as core from "zod/v4/core";
 import { getInteropZodDefaultGetter } from "@langchain/core/utils/types";
 import { $ZodType, $ZodRegistry, $replace } from "zod/v4/core";
 import { SchemaMeta, SchemaMetaRegistry, schemaMetaRegistry } from "./meta.js";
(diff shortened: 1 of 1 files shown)


nfcampos (dismissed):

libs/langgraph/src/graph/zod/zod-registry.ts
```diff
@@ -46,4 +49,30 @@ export class LanggraphZodMetaRegistry<
...
+      TOutput = core.output<this>,
+      TInput = core.input<this>,
+      TInternals extends core.$ZodTypeInternals<
+        TOutput,
+        TInput
+      > = core.$ZodTypeInternals<TOutput, TInput>
+    >(
+      registry: R,
+      meta: SchemaMeta<TOutput, TInput>
+    ): ZodType<TOutput, TInput, TInternals>;
```
hntrl: should be something like this (to match withLangGraph):
```suggestion
    ): ReducedZodChannel<core.$ZodType<Output, Input>, TInput>;
```

hntrl: Isn't this restrictive on the types accepted by `meta.reducer.schema`? To me this reads as the reducer schema needs the same input/ output types (which I don't think is what we want)

hntrl: (it isnt)

hntrl (approved):
thanks for looking at this!

Issue #1195, opened 2025-05-19
Add `checkpoint_during`


dqbd:
- https://github.com/langchain-ai/langgraph/pull/4169

PR #1290, opened 2025-06-17
feat(langgraph): implement checkpoint_during

dqbd:
Port of https://github.com/langchain-ai/langgraph/pull/4169

Closes Add `checkpoint_during` from LangGraph Python #1195

### DIFF
--- a/libs/langgraph/src/pregel/loop.ts
+++ b/libs/langgraph/src/pregel/loop.ts
@@ -47,6 +47,8 @@ import {
   PUSH,
   CONFIG_KEY_SCRATCHPAD,
   CONFIG_KEY_CHECKPOINT_NS,
+  CHECKPOINT_NAMESPACE_END,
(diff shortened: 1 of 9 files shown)


libs/langgraph/src/pregel/loop.ts
```diff
@@ -824,6 +834,22 @@ export class PregelLoop {
...
   async finishAndHandleError(error?: Error) {
+    // persist current checkpoint and writes
+    if (
+      !this.checkpointDuring &&
+      // if it's a top graph
+      (!this.isNested ||
+        // or a nested graph with error or interrupt
+        error !== undefined ||
+        // or a nested graph with checkpointer: true
+        this.checkpointNamespace.some((part) =>
```
nfcampos: I think this should be the opposite condition no?

dqbd: Addresed in https://github.com/langchain-ai/langgraphjs/pull/1301

PR #1315, opened 2025-06-26
fix(langgraph): change return value of getWriter

bkim14:
writer exists in runConfig.
 fix getWriter to correctly pass writer.


dqbd:
Thank you @bkim14!

PR #1375, opened 2025-07-09
feat(sdk): make history fetching configurable in useStream

dqbd:
useStream now accepts `fetchStateHistory`. This can be used to limit the number of checkpoints returned by the API or switch to `/getState` endpoint if `fetchStateHistory: false`.

### DIFF
--- a/libs/sdk/src/react/stream.tsx
+++ b/libs/sdk/src/react/stream.tsx
@@ -200,6 +200,18 @@ function getBranchSequence<StateType extends Record<string, unknown>>(
 ) {
   const childrenMap: Record<string, ThreadState<StateType>[]> = {};
 
+  // Short circuit if there's only a singular one state
+  // TODO: I think we can make this more generalizable for all `fetchStateHistory` values.
+  if (history.length <= 1) {
+    return {
+      rootSequence: {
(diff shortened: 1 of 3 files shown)


libs/sdk/src/react/stream.tsx
```diff
@@ -314,14 +326,23 @@ function getBranchView<StateType extends Record<string, unknown>>(
...
+  options?: { limit?: boolean | number }
 ) {
-  return client.threads.getHistory<StateType>(threadId, { limit: 1000 });
+  if (options?.limit === false) {
+    return client.threads
+      .getState<StateType>(threadId)
+      .then((state) => [state]);
+  }
+
+  const limit = typeof options?.limit === "number" ? options.limit : 1000;
```
hinthornw, line 338: should we also decrease the default? 1k is probably more than what people would want (?)

hinthornw (approved):

libs/sdk/src/react/stream.tsx
dqbd: Would like to leave it as-is for now, since this may break users who rely on large limit (ie. a graph with a lot of nodes executed in between forks)

PR #1381, opened 2025-07-11
feat: Exactly once for run ID resuming

hinthornw:
Mirror https://github.com/langchain-ai/langgraph/blob/579c7831b915d7e596d612333da4b94be7000c72/libs/langgraph/langgraph/pregel/loop.py#L585

### DIFF
--- a/libs/langgraph/package.json
+++ b/libs/langgraph/package.json
@@ -22,6 +22,7 @@
     "lint:fix": "yarn lint:eslint --fix && yarn lint:dpdm",
     "prepack": "yarn build",
     "test": "vitest run",
+    "test:single": "vitest run --mode single",
     "test:browser": "vitest run --mode browser",
     "test:watch": "vitest watch",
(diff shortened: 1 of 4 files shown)


libs/langgraph/package.json
```diff
@@ -22,6 +22,7 @@
     "lint:fix": "yarn lint:eslint --fix && yarn lint:dpdm",
     "prepack": "yarn build",
     "test": "vitest run",
+    "test:single": "vitest run --mode single",
```
hntrl:
```suggestion
```

hntrl: You should just be able to do `yarn test <file>` in place of test:single

hntrl (dismissed):
Could you run `yarn changeset` from the root of the repo and follow the prompts? Should just be a patch bump for langgraph. We have a new utility for creating releases and it makes it easier if the PR author includes the change.

Otherwise LGTM

PR #1384, opened 2025-07-14
fix(langgraph): dedupe signal listeners

hntrl:
Currently, our approach to combining event listeners has a tendency to emit an annoying 'potential memory leak' error on some versions of node. This is because we add an event listener to the abort signal (whether passed in from options or from the default AbortController that gets created) on every runner tick, which as graphs get over a certain complexity we create more than 10 event listeners on a single signal which triggers the warning.

This fixes that by using a hierarchical `MultiAbortSignal` class in place of `combineAbortSignals` which dedupes listeners by ensuring that any signals passed in the constructor arent already tracked by a downstream abort signal.


dqbd:
Closed in favour of #1404

PR #1395, opened 2025-07-14
feat(sdk): add `isStudioUser` for custom auth

dqbd:


docs/docs/how-tos/auth/custom_auth.md
```diff
@@ -132,3 +132,30 @@ Assuming you are using JWT token authentication, you could access your deploymen
     ```bash
     curl -H "Authorization: Bearer ${your-token}" http://localhost:2024/threads
     ```
+
+### Authorizing a Studio user
+
+By default, if you add custom authorization on your resources, this will also apply to interactions made from the Studio. If you want, you can handle logged-in Studio users in a special way with [isStudioUser()](../../reference/functions/sdk_auth.isStudioUser.html).
```
hinthornw, line 138: Should we include the min sdk version?

dqbd: Changelog is fine I think?

PR #1402, opened 2025-07-15
feat(langgraph): add UpdateType type utility for Zod, improve Zod 4 and Zod 4 mini support

dqbd:
Also add tests for Zod 3, 4 and 4 mini.

### DIFF
--- a/libs/langgraph/src/graph/messages_annotation.ts
+++ b/libs/langgraph/src/graph/messages_annotation.ts
@@ -1,10 +1,10 @@
 /* __LC_ALLOW_ENTRYPOINT_SIDE_EFFECTS__ */
 
 import { BaseMessage } from "@langchain/core/messages";
-import { z } from "zod";
+import { z } from "zod/v3";
 import { Annotation } from "./annotation.js";
 import { Messages, messagesStateReducer } from "./message.js";
-import { withLangGraph } from "./zod/meta.js";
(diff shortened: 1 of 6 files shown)


libs/langgraph/src/graph/messages_annotation.ts
```diff
@@ -48,6 +48,25 @@ export const MessagesAnnotation = Annotation.Root({
...
+ * ```ts
+ * import { z } from "zod/v4-mini";
+ * import { MessagesZodState, StateGraph } from "@langchain/langgraph";
+ *
+ * const AgentState = z.object({
+ *   messages: z.custom<BaseMessage[]>().register(registry, MessagesZodMeta),
+ * });
+ * ```
+ */
+export const MessagesZodMeta: SchemaMeta<BaseMessage[], Messages> = {
```
dqbd, line 64: We need to nudge the type here to avoid being tied to runtime Zod.

libs/langgraph/src/graph/messages_annotation.ts
```diff
@@ -88,14 +107,5 @@ export const MessagesAnnotation = Annotation.Root({
  * ```
  */
 export const MessagesZodState = z.object({
```
dqbd, line 109: At the end of the day I think we want to discourage using this with Zod 4 going forward

PR #1407, opened 2025-07-16
fix(langgraph): preModelHook `llmInputMessages` should not keep concatenating messages

### LINKED ISSUE #1316: 【BUG】Agent node cannot intercept messages normally after combining trimMessage
trimMessage works fine, but for some reason, there is a problem when passing it to the next agent node.

As shown in the figure, the final result returned by trimMessage is "Tell me another one", which is correct.

Image

But the message sent to llm in the end sent all human messages.

Image

Image

Here is my code
```
  private readonly trimMessages = trimMessages({
    maxTokens: 50,
    tokenCounter: new ChatOpenAI({ model: 'gpt-4o' }),
    strategy: 'last',
    includeSystem: true,
    startOn: 'human',
  });

```
```
  createReactAgent({
        llm: this.getModel(model),
        tools,
        prompt,
        name,
        checkpointer,
        responseFormat,
        preModelHook: async (state) => {
          const messages = await this.trimMessages.invoke(state.messages);
          return { llmInputMessages: messages };
        },
      });
```

dqbd:
Closes #1316

### DIFF
--- a/.changeset/orange-beans-learn.md
+++ b/.changeset/orange-beans-learn.md
@@ -2,4 +2,4 @@
 "@langchain/langgraph": patch
 ---
 
-fix(langgraph): preModelHook `llmInputMessages` should not keep incrementing messages
+fix(langgraph): preModelHook `llmInputMessages` should not keep concatenating messages
--- a/libs/langgraph/src/tests/prebuilt.test.ts
+++ b/libs/langgraph/src/tests/prebuilt.test.ts
@@ -1096,7 +1096,7 @@ describe("createReactAgent with ToolNode", () => {
 });
 
 describe("createReactAgent with hooks", () => {
-  it.only("preModelHook", async () => {
+  it("preModelHook", async () => {
     const llm = new FakeToolCallingChatModel({
       responses: [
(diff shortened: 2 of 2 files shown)


libs/langgraph/src/prebuilt/react_agent_executor.ts
```diff
@@ -457,7 +457,7 @@ type WithStateGraphNodes<K extends string, Graph> = Graph extends StateGraph<
 
 const PreHookAnnotation = Annotation.Root({
   llmInputMessages: Annotation<BaseMessage[], Messages>({
-    reducer: messagesStateReducer,
+    reducer: (_, update) => messagesStateReducer([], update),
```
Diluka, line 460: Why don‘t you return it directly, but use that function again? Does it have any special function?

dqbd: We want to make sure that (1) we're sending the same messages state-wise (assigning IDs if they are not present) and (2) we still accept `Messages` as update instead of just `BaseMessage[]`

PR #1416, opened 2025-07-17
feat(api,sdk): add typing and support for "checkpoints" and "tasks" stream mode

dqbd:


libs/sdk/src/react/stream.tsx
```diff
@@ -545,6 +547,20 @@ export interface UseStreamOptions<
    */
   onDebugEvent?: (data: DebugStreamEvent["data"]) => void;
 
+  /**
+   * Callback that is called when a checkpoints event is received.
+   */
+  onCheckpointsEvent?: (
```
nfcampos: should it be onCheckpointEvent/onTaskEvent without the s for consistency?

PR #1426, opened 2025-07-21
Feat/ElasticSearch Introduce ElasticSearch checkpointer

KDKHD:
Introduces a new checkpointer that uses ElasticSearch as the backing database.


dqbd:
Hello! Thank you for the PR, will currently hold off from reviewing since we're updating the checkpointer interface a bit for 0.4

KDKHD:
We have decided to build the Elastic Search checkpointer inside of [Kibana first](https://github.com/elastic/kibana/blob/main/x-pack/platform/packages/shared/kbn-langgraph-checkpoint-saver/server/elastic-search-checkpoint-saver/index.ts). Once we have tested it there, and it is more mature, we will consider contributing it to this repository.

A new PR will be opened when we get to that stage.

PR #1445, opened 2025-07-24
feat(langgraph): cleanup of interrupt interface

dqbd:


libs/langgraph/src/interrupt.ts
```diff
@@ -107,14 +107,11 @@ export function interrupt<I = unknown, R = any>(value: I): R {
   );
   throw new GraphInterrupt([
     {
-      value,
-      when: "during",
-      resumable: true,
-      ns,
-      get interrupt_id() {
+      get id() {
```
nfcampos: how is a getter persisted in checkpointer?

PR #1452, opened 2025-07-26
fix(sdk): handle truncated history when root checkpoint missing

ibbybuilds:
## Problem

When using `fetchStateHistory: { limit: N }`, two critical issues occurred:

1. **Empty Branch Tree**: The `getBranchSequence` function would return an empty branch tree, causing downstream consumers to receive an empty messages array. This happened because:
   - Limited history requests don't include the very first checkpoint (whose parent is `null`)
   - The algorithm looks for children of the synthetic root `"$"` but finds none
   - Result: empty `rootSequence.items` → empty `history` → empty `messages` array

2. **Incorrect Message Metadata**: Messages beyond the loaded history window received incorrect metadata, allowing editing/regeneration from wrong checkpoints. This happened because:
   - `findLastIndex` would find messages in the oldest loaded state
   - Each LangGraph state contains the entire conversation history
   - Messages created before the loaded window appeared to be "created" in the oldest loaded state
   - Result: editing old messages would branch from incorrect checkpoints

### Fix 1: Branch Tree Fallback
Added a fallback mechanism that detects "orphan" checkpoints - those whose parent isn't present in the truncated slice - and attaches them directly to the root.

### Fix 2: Accurate Message Creation Detection  
Replaced the flawed `findLastIndex` approach with `findMessageCreationState` function that:
- Iterates through loaded history chronologically
- Identifies the first state where a message appears that wasn't in the previous state
- Returns `undefined` for messages that existed before the loaded window
- Prevents editing/regeneration when parent checkpoint is unknown

### `src/react/stream.tsx`
- ✅ Exported `getBranchSequence` for testing
- ✅ Added orphan-to-root fallback with clear documentation
- ✅ Moved constants to fix linting issues
- ✅ Implemented `findMessageCreationState` algorithm
- ✅ Updated `messageMetadata` to use accurate creation state detection
- ✅ Removed unused `findLastIndex` function

### `src/tests/branchSequence.test.ts`  
- ✅ New unit test verifying truncated histories produce non-empty trees

## Impact

- **Before**: `fetchStateHistory: { limit: 5 }` → empty messages array + incorrect edit behavior
- **After**: `fetchStateHistory: { limit: 5 }` → returns expected message list + safe editing only for messages with known parent checkpoints
- **No breaking changes** - existing functionality unchanged
- **Patch release** recommended

## Testing

- [x] Build passes
- [x] All existing tests pass  
- [x] New test covers the branch tree regression
- [x] Lint clean
- [x] Manual verification in production app
- [x] Edit/regenerate buttons properly hidden for messages without reliable parent checkpoints

Fixes empty message arrays and prevents incorrect branching when using `fetchStateHistory` limits.

Twitter/X handle: `ibbyybuilds`

### DIFF
--- a/libs/sdk/src/react/stream.tsx
+++ b/libs/sdk/src/react/stream.tsx
@@ -138,13 +138,6 @@ function unique<T>(array: T[]) {
   return [...new Set(array)] as T[];
 }
 
-function findLastIndex<T>(array: T[], predicate: (item: T) => boolean) {
-  for (let i = array.length - 1; i >= 0; i -= 1) {
-    if (predicate(array[i])) return i;
-  }
-  return -1;
-}
-
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 interface Node<StateType = any> {
   type: "node";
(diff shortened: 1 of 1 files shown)


dqbd:
Thank you! Superseded by #1557

PR #1480, opened 2025-08-01
fix(langgraph): handle empty messages

yt20chill:
Gracefully handle empty messages when deciding conditional mapping to "tools" or END

If messages is empty, message is assigned to `undefined`, and "tool_calls" in message will throw.


dqbd:
Thank you!

PR #1551, opened 2025-08-20
fix(langgraph): task result from stream mode debug / tasks should match format from getStateHistory / getState

dqbd:


libs/langgraph/src/pregel/debug.ts
```diff
@@ -104,6 +104,33 @@ export function* mapDebugTasks<N extends PropertyKey, C extends PropertyKey>(
...
+function mapTaskResultWrites(writes: PendingWrite<unknown>[]) {
+  const result: Record<string, unknown> = {};
+
+  for (const [channel, value] of writes) {
+    const strChannel = String(channel);
+
+    if (strChannel in result) {
+      const channelWrites = isMultipleChannelWrite(result[strChannel])
+        ? result[strChannel].$writes
+        : [result[strChannel]];
```
lc-arjun, line 123: nit: instead of just checking `isMultipleChannelWrite` why not have the helper instead just return channel writes?

lc-arjun (dismissed):
potentially worth getting another reviewer

libs/langgraph/src/pregel/debug.ts
dqbd: I think we want to be able to distinguish multiple channel writes vs. an array as a write

PR #1553, opened 2025-08-20
refactor(api): Add abstraction layer for storage operations

pjrule:
## Summary
This refactors `storage` to support a pluggable operations layer. This is intended to eventually enable alternate storage backends.

##  How
(I'm using the term "repository" in the ports and adapters/hexagonal architecture sense—happy to tweak this nomenclature as desired.)

* Converts static methods in the CRUD/data repository classes in `storage/ops.mts` to instance methods that consume a connection object.
* Extracts interfaces for these updated repository classes, plus related type definitions, into a separate type definition file (`storage/types.mts`). These definitions are exported under `@langgraph/langgraph-api/storage`.
* Adds a wrapper interface (`Ops`) around all data repositories. This interface will need to be implemented by any alternate storage backend.
* Updates API implementations to consume the updated repository classes. The global singleton `FileSystemPersistence`  object is removed in favor of initializing persistence in `startServer()`. An object implementing `ops` is injected into each request using Hodo's [context storage middleware](https://hono.dev/docs/middleware/builtin/context-storage).
* Exports `startServer` (now parameterized with `storage`); this is almost certainly necessary for writing an alternate entrypoint that uses an alternate storage backend.
    * Moves version fetching in `api/meta.mts` to middleware to avoid an error related to root-level `async` usage.

This is intended as an almost entirely pure refactoring, so no new tests are added here (yet).


libs/langgraph-api/src/api/meta.mts
```diff
@@ -13,22 +13,26 @@ const packageJsonPath = path.resolve(
...
-} catch {
-  console.warn("Could not determine version of langgraph-api");
-}
 
-// Get the installed version of @langchain/langgraph
-try {
-  const langgraphPkg = await import("@langchain/langgraph/package.json");
-  if (langgraphPkg?.default?.version) {
-    langgraph_js_version = langgraphPkg.default.version;
+api.use(async (_, next) => {
```
jdrogers940: Why do we want/need to do this on every request? Seems like a lot of overhead.

libs/langgraph-api/src/cli/spawn.mts
```diff
@@ -31,6 +32,9 @@ export async function spawnServer(
         };
       };
     };
+    storage?: {
```
jdrogers940: Are we actually passing this through anywhere?

libs/langgraph-api/src/storage/types.mts
```diff
@@ -0,0 +1,445 @@
...
+    config: RunnableConfig,
+    values:
+      | Record<string, unknown>[]
+      | Record<string, unknown>
+      | null
+      | undefined,
+    asNode: string | undefined,
+    auth: AuthContext | undefined
+  ): Promise<{ checkpoint: Record<string, unknown> | undefined }>;
+  bulk(
```
jdrogers940, line 346: nit: Noticing some spacing things. Can we make sure we've got a linter here too?

libs/langgraph-api/src/server.mts
```diff
@@ -87,11 +93,24 @@ export async function startServer(options: z.infer<typeof StartServerSchema>) {
...
+      runs: {},
+      threads: {},
+      assistants: {},
+      assistant_versions: [],
+      retry_counter: {},
+    }));
+    initCalls.push(opsConn.initialize(options.cwd));
+  }
+  const callbacks = await Promise.all(initCalls);
+  const ops =
```
jdrogers940: Should we just group this into the logic above:

```
ops = storage?.ops;
if (ops == null) {
  opsConn = ...;
  initCalls.push(...);
  ops = new FileSystemOps(...);
}
```

Will we need initialization for other ops types as well? Maybe can be refactored to support down the road.

libs/langgraph-api/src/server.mts
```diff
@@ -132,7 +151,13 @@ export async function startServer(options: z.infer<typeof StartServerSchema>) {
     return info;
   });
 
-  const app = new Hono();
+  const app = new Hono<StorageEnv>();
+
+  app.use(contextStorage());
+  app.use(async (c, next) => {
```
jdrogers940, line 160: I guess we'll figure out, but any idea of how expensive this is to set on each call?

pjrule: Shouldn't be expensive, as `ops` is only actually initialized once.

libs/langgraph-api/src/server.mts
pjrule: ah yeah, that's a bit cleaner. fixed

libs/langgraph-api/src/storage/types.mts
pjrule: ooh yeah, looks like we're not running a linter in this package. Most of the other ones are using ESLint + dpdm. I can open a separate PR to set it up and fix any violations...

libs/langgraph-api/src/api/meta.mts
pjrule: fixed—back to loading version info once now! we just can't load it at the root level because root-level async was causing some export issues...

libs/langgraph-api/src/storage/types.mts
pjrule: (btw, I did fix this particular spacing issue)

libs/langgraph-api/src/cli/spawn.mts
pjrule: ha, good catch—this was vestigial. I ultimately realized `startServer()` is a better place to pass this. removed it here!

libs/langgraph-api/src/server.mts
```diff
@@ -132,7 +154,13 @@ export async function startServer(options: z.infer<typeof StartServerSchema>) {
...
+    c.set("ops", ops);
```
dqbd: _nit_: can we use different name for ops stored within context variables (ie: `LANGGRAPH_OPS`?), since they are visible in custom HTTP routes

libs/langgraph-api/src/queue.mts
```diff
@@ -12,18 +12,23 @@ import { getGraph } from "./graph/load.mjs";
 const MAX_RETRY_ATTEMPTS = 3;
 
 const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
-export const queue = async () => {
+export const queue = async (ops: Ops) => {
```
dqbd, line 15: We probably want to abstract queue as well

dqbd (dismissed):

libs/langgraph-api/src/queue.mts
pjrule: Agreed—expect a followup PR for that :)

libs/langgraph-api/src/server.mts
pjrule: ah yes, good point—updated

PR #1558, opened 2025-08-21
feat(api): add thread join stream

dqbd:


libs/langgraph-api/src/storage/ops.mts
```diff
@@ -1805,8 +1811,74 @@ export class Runs {
...
+
+            const streamTopic = message.topic.substring(
+              `thread:${threadId}:stream:`.length
+            );
+
+            yield { id, event: streamTopic, data: message.data };
+          } catch (error) {
+            if (error instanceof AbortError) break;
+
+            // check if there are any runs that are pending or running
```
hinthornw, line 1862: I think here we'd want to actually just keep listening until the client disconnects. Otherwise, the client would have to keep polling anyway

PR #1572, opened 2025-08-25
feat(sdk): add thread streaming endpoint

isahers1:


hntrl (approved):

libs/sdk/src/client.ts
```diff
@@ -992,6 +992,28 @@ export class ThreadsClient<
       }
     );
   }
+
+  async *getStream(
```
dqbd: We probably want to name this `joinStream` as in Python SDK?
https://github.com/langchain-ai/langgraph/pull/6009/files?w=1

libs/sdk/src/client.ts
```diff
@@ -992,6 +992,28 @@ export class ThreadsClient<
...
+    threadId: string,
+    lastEventId?: string
```
dqbd: _nit_: make this accept an object instead for future proofing

```suggestion
    options?: { lastEventId?: string }
```

PR #1576, opened 2025-08-26
chore(prebuilt): deprecate createReactAgent

christian-bromann:
We have ported `createReactAgent` to the `langchain` package and can deprecate the version we have in Langgraph.


README.md
```diff
@@ -16,52 +16,6 @@ npm install @langchain/langgraph @langchain/core
 
 To learn more about how to use LangGraph, check out [the docs](https://langchain-ai.github.io/langgraphjs/). We show a simple example below of how to create a ReAct agent.
```
sydney-runkle, line 18: Can't just remove, let's redo to use create react agent from langchain + suggest migration

libs/langgraph/src/prebuilt/react_agent_executor.ts
```diff
@@ -654,6 +654,8 @@ export type CreateReactAgentParams<
  * }
  * // Returns the messages in the state at each step of execution
  * ```
+ *
```
sydney-runkle: What are the implications of this for ts developers? Can they see this in their IDE? Does this raise a warning?

christian-bromann: Yes:
Screenshot 2025-08-27 at 10 20 18 AM

dqbd (approved):

PR #1577, opened 2025-08-27
feat: Add initialThreadLoaded

hieusmiths:
https://forum.langchain.com/t/usestream-how-to-detect-first-history-loaded/1132/1

We need a flag to indicate that a thread is loaded.

We have an `isThreadLoading` flag, but it starts as false and is a dynamic flag.Whether the thread is currently being loaded.

For a loaded state, we need a separate flag. So I created a flag `initialThreadLoaded`

Example: A chat application component might have `initialThreadLoaded` to track if it has fetched the first set of messages. This is useful for:

- Showing a loading spinner only on the first load.
- Preventing redundant initial data fetches.
- Showing a "welcome" message or onboarding UI after the data is available.

### DIFF
--- a/libs/sdk/src/react/stream.tsx
+++ b/libs/sdk/src/react/stream.tsx
@@ -697,6 +697,11 @@ export interface UseStream<
   isThreadLoading: boolean;
 
   /**
+   * Has the initial set of threads and their data been successfully loaded.
+   */
+  initialThreadLoaded: boolean;
+
+  /**
    * Stops the stream.
    */
   stop: () => void;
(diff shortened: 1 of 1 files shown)


dqbd:
Hello! I think this might be a bug in `isThreadLoading` then, since it should not start with `false`.

hieusmiths:
> Hello! I think this might be a bug in `isThreadLoading` then, since it should not start with `false`.

Thank for response, Base on the behavior if start with true so behavior will be.
`isLoading`
- True: whenever the first fetch for a query is in-flight, no query attempt was finished yet
- False: First fetch for a query was finished

And I see in lib we are constraint it with `history.data != null` It mean `isLoading` also make sure that query attempt must be successfully. 

CMIIW

dqbd:
Thank you! Superseded by #1581 and released in 0.0.112, let us know if the issue is fixed!

PR #1583, opened 2025-08-28
feat(build): update to new build system

christian-bromann:
Some questions we should resolve before publishing the package from LangChain:

- we need to migrate to pnpm first or add support for yarn in `@langchain/build`
- `@langchain/langgraph-checkpoint-validation` has `runner.ts` which currently fails the build because: we can't have ESM only builds for specific import paths only (it can't be compiled into ESM due to root level `await`)


libs/checkpoint/package.json
```diff
@@ -6,21 +6,20 @@
...
+  "main": "./dist/index.js",
+  "types": "./dist/index.d.ts",
   "repository": {
     "type": "git",
     "url": "git@github.com:langchain-ai/langgraphjs.git"
   },
   "scripts": {
     "build": "yarn turbo:command build:internal --filter=@langchain/langgraph-checkpoint",
-    "build:internal": "yarn clean && yarn lc_build --create-entrypoints --pre --tree-shaking",
+    "build:internal": "yarn workspace @langchain/build compile @langchain/langgraph-checkpoint",
```
dqbd, line 17: I think ideally we we don't need to specify package name here

environment_tests/test-exports-cf/package.json
```diff
@@ -8,7 +8,7 @@
     "@cloudflare/workers-types": "^4.20230321.0"
   },
   "dependencies": {
-    "@langchain/core": ">=0.2.31 <0.4.0",
+    "@langchain/core": ">=0.2.31 <0.4.0 || ^1.0.0",
```
hntrl: this dep range wont grab alpha

libs/checkpoint-mongodb/package.json
```diff
@@ -6,21 +6,20 @@
...
   "scripts": {
     "build": "yarn turbo:command build:internal --filter=@langchain/langgraph-checkpoint-mongodb",
-    "build:internal": "yarn clean && yarn lc_build --create-entrypoints --pre --tree-shaking",
+    "build:internal": "yarn workspace @langchain/build compile @langchain/langgraph-checkpoint-mongodb",
     "clean": "rm -rf dist/ dist-cjs/ .turbo/",
     "lint:eslint": "NODE_OPTIONS=--max-old-space-size=4096 eslint --cache --ext .ts,.js src/",
     "lint:dpdm": "dpdm --exit-code circular:1 --no-warning --no-tree src/*.ts src/**/*.ts",
     "lint": "yarn lint:eslint && yarn lint:dpdm",
     "lint:fix": "yarn lint:eslint --fix && yarn lint:dpdm",
-    "prepack": "yarn build",
```
dqbd, line 23: Removing `prepack` will not build the package when npm release is called

libs/checkpoint/package.json
christian-bromann: You can call `yarn workspace @langchain/build compile` to compile all packages but in order to just compile a specific workspace you have to provide a filter. Alternatively I think:
```suggestion
    "build:internal": "yarn workspace @langchain/build compile $npm_package_name",
```

should work?

environment_tests/test-exports-cf/package.json
christian-bromann: Should we just
```suggestion
    "@langchain/core": "1.0.0-alpha.1",
```

for now?

libs/checkpoint-mongodb/package.json
christian-bromann: I had a weird infinite loop of build script calls with this in place. I think one of the rolldown plugin calls `npm pack` for package analysis.

PR #1601, opened 2025-09-02
Agent Middleware

christian-bromann:
Opening for review purposes only.

### DIFF
--- a/libs/langgraph/src/agent/agent.ts
+++ b/libs/langgraph/src/agent/agent.ts
@@ -1,0 +1,203 @@
+import { LanguageModelLike } from "@langchain/core/language_models/base";
+import { type RunnableToolLike } from "@langchain/core/runnables";
+import type {
+  DynamicTool,
+  StructuredToolInterface,
+} from "@langchain/core/tools";
+import { ToolNode } from "../prebuilt";
+import { Annotation, StateGraph } from "../graph";
+import { END, MessagesAnnotation, START } from "../web";
+import {
+  BaseMessage,
+  isAIMessage,
+  SystemMessage,
+} from "@langchain/core/messages";
+import { unknown } from "zod/v4";
+import {
+  BaseChatModel,
+  ToolChoice,
+} from "@langchain/core/language_models/chat_models";
+import { EphemeralValue } from "../channels/ephemeral_value";
+
+type Tool = StructuredToolInterface | DynamicTool | RunnableToolLike;
+
+export const AgentState = Annotation.Root({
+  messages: MessagesAnnotation["spec"]["messages"],
+  jumpTo: new EphemeralValue<"tools" | "model" | "__end__">(),
+  response: Annotation<Record<string, unknown> | undefined>(),
+});
+
+export type AgentState = typeof AgentState;
+
+const isToolReturnDirect = (tool: Tool) =>
+  "returnDirect" in tool && tool.returnDirect;
+
+export interface AgentMiddleware {
+  name: string;
+  beforeModel?: (
+    state: AgentState["State"]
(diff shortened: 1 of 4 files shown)


libs/langgraph/src/agent/agent.ts
```diff
@@ -0,0 +1,203 @@
...
+      // TODO: add support for responseFormat
+      const output = await model.invoke(messages);
+      return { messages: output, response: undefined };
+    },
+
+    tools: toolNode,
+  });
+
+  const firstNode = Object.keys(toolNodeBefore).at(0) ?? "modelRequest";
+  const lastNode = Object.keys(toolNodeAfter).at(0) ?? "modelRequest";
```
christian-bromann, line 148: Given `afterModel` hooks run in reversed order, shouldn't this be
```suggestion
  const lastNode = Object.keys(toolNodeAfter).at(-1) ?? "modelRequest";
```

libs/langgraph/src/agent/middleware/humanInTheLoop.ts
```diff
@@ -0,0 +1,120 @@
...
+import {
+  HumanInterruptConfig,
+  ActionRequest,
+  HumanInterrupt,
+  HumanResponse,
+} from "../../prebuilt/interrupt";
+
+export type ToolInterruptConfig = Record<string, HumanInterruptConfig>;
+
+export class HumanInTheLoopMiddleware implements AgentMiddleware {
```
christian-bromann, line 13: Do you have an example here? Let's say we have a GitHub triage agent that we trigger and requests approval of the triage result before posting a comment on the issue. How would that be implemented from end-to-end?

libs/langgraph/src/agent/agent.ts
```diff
@@ -0,0 +1,203 @@
...
+    (acc, { name, beforeModel }) => {
+      if (beforeModel) acc[`${name}.beforeModel`] = beforeModel;
+      return acc;
+    },
+    {}
+  );
+
+  const toolNodeAfter = middleware.reduce<MiddlewareMap>(
+    (acc, { name, afterModel }) => {
+      if (afterModel) acc[`${name}.afterModel`] = afterModel;
```
christian-bromann, line 100: What happens if something throws within `beforeModel`, `afterModel` or `modifyModelRequest`?

libs/langgraph/src/agent/agent.ts
```diff
@@ -0,0 +1,203 @@
...
+  jumpTo: new EphemeralValue<"tools" | "model" | "__end__">(),
+  response: Annotation<Record<string, unknown> | undefined>(),
+});
+
+export type AgentState = typeof AgentState;
+
+const isToolReturnDirect = (tool: Tool) =>
+  "returnDirect" in tool && tool.returnDirect;
+
+export interface AgentMiddleware {
```
christian-bromann, line 35: Doesn't this type require to have `state` and `context` generics so these are propagated to the hooks?

christian-bromann: also should we also include `beforeTool` and `afterTool`? Not that we specced this out before but seems to be a logical enhancements, maybe for later?

libs/langgraph/src/agent/agent.ts
dqbd: This should be correct, since toolNodeAfter callbacks are inserted in the same order as toolNodeBefore

libs/langgraph/src/agent/agent.ts
dqbd: Same as with an exception occurring in a node: the graph is paused till the checkpoint and can be resumed from there

dqbd:
Closing in favour of implementation found in `langchain`

PR #1612, opened 2025-09-08
feat(langgraph): allow explicit type declaration for interrupt and custom stream events

dqbd:


libs/langgraph/src/pregel/runnable_types.ts
```diff
@@ -24,23 +24,43 @@ export type RunnableLike<
...
+  ContextType extends Record<string, any> = Record<string, any>,
+  InterruptType = unknown,
+  WriterType = unknown
 > extends RunnableConfig<ContextType> {
   context?: ContextType;
 
   store?: BaseStore;
 
-  writer?: (chunk: unknown) => void;
+  writer?: IsEqual<WriterType, unknown> extends true
```
christian-bromann, line 39: Are there cases where `writer` is not defined? Wonder if we can remove the optionality.

libs/langgraph/src/pregel/runnable_types.ts
```diff
@@ -24,23 +24,43 @@ export type RunnableLike<
...
+    ? (chunk: unknown) => void
+    : WriterType;
+
+  interrupt?: IsEqual<InterruptType, unknown> extends true
```
christian-bromann, line 43: Is it possible to infer the existence of `interrupt` based on whether a checkpointer parameter was provided?

libs/langgraph/src/pregel/runnable_types.ts
```diff
@@ -24,23 +24,43 @@ export type RunnableLike<
...
+    ? (value: unknown) => unknown
+    : InterruptType;
 }
 
-export interface Runtime<ContextType = Record<string, unknown>> {
+export interface Runtime<
```
christian-bromann, line 48: How is `Runtime` different from `LangGraphRunnableConfig`? Both seem to have very similar types. Maybe we can do:

```ts
export interface Runtime<
  ContextType = Record<string, unknown>,
  InterruptType = unknown,
  WriterType = unknown
> extends LangGraphRunnableConfig<ContextType, InterruptType, WriterType> {
  signal?: AbortSignal;
}
```

libs/langgraph/src/pregel/runnable_types.ts
dqbd: Writer is currently defined only when `custom` stream mode is requested

libs/langgraph/src/pregel/runnable_types.ts
dqbd: Since I'd like to support creating the builder separately, that would entail making `checkpointer` a constructor parameter.

An alternative I think can be always providing the function in runtime but throwing an error in runtime if no checkpointer is passed

libs/langgraph/src/pregel/runnable_types.ts
dqbd: Runtime is a pruned version of LangGraphRunnableConfig, so ideally it should be in reverse.

libs/langgraph/src/pregel/runnable_types.ts
christian-bromann: I think we throw an error already that is meaningful enough.

libs/langgraph/src/pregel/runnable_types.ts
christian-bromann: Yeah, that's not possible to type. In this case, I would advocate to remove the optionality afterall and throw a meaningful error at runtime, because the optionality doesn't really provide any value IMHO.

PR #1616, opened 2025-09-08
feat(langgraph): narrow Runtime types to remove optionality from `writer`, `interrupt` and `signal`

dqbd:


libs/langgraph/src/pregel/index.ts
```diff
@@ -2025,24 +2025,25 @@ export class Pregel<
...
-      config.interrupt = interrupt;
-    }
+      stream.push([ns ?? [], "custom", chunk]);
+    };
+
+    config.interrupt =
+      this.checkpointer != null
+        ? interrupt
+        : () => {
+            throw new GraphValueError("No checkpointer set");
```
christian-bromann, line 2045: Worth adding a link to this?
```suggestion
            throw new GraphValueError("No checkpointer set! Make sure you provide a memory store to the graph, see https://docs.langchain.com/oss/javascript/langgraph/persistence#memory-store");
```

dqbd: Would probably use custom troubleshooting code, will add in a separate PR + docs

PR #1635, opened 2025-09-10
[Feature] Add onerror cb for submit

hieusmiths:
Current Situation:
We are currently using the onError callback within useStream to handle all errors. However, this approach seems too generic and is primarily suitable for logging purposes rather than robust error handling.

Problem:
The generic error handling is not sufficient for user-facing scenarios where specific error conditions require particular user actions, such as retrying an operation.

Sample Use Case:

- A user submits a request to initiate a stream.
- The stream operation fails.
- The application needs to display a specific error message to the user, along with an option to retry the operation.

### DIFF
--- a/libs/sdk/src/react/stream.tsx
+++ b/libs/sdk/src/react/stream.tsx
@@ -471,6 +471,7 @@ export function useStream<
           return undefined;
         },
         onError(error) {
+          submitOptions?.onError?.(error);
           options.onError?.(error, callbackMeta);
         },
       }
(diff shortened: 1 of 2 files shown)


dqbd:
Hello! I think it might make more sense to handle the exceptions in a `try / catch` instead, will check the types!

hieusmiths:
> Hello! I think it might make more sense to handle the exceptions in a `try / catch` instead, will check the types!

Yes, using try-catch on a submit Promise would be good. Unfortunately, `client.runs.stream` is designed for callbacks, and all its behavior is already within try-catch blocks without re-throwing, so we can't implement that here.

hieusmiths:
> > Hello! I think it might make more sense to handle the exceptions in a `try / catch` instead, will check the types!
> 
> Yes, using try-catch on a submit Promise would be good. Unfortunately, `client.runs.stream` is designed for callbacks, and all its behavior is already within try-catch blocks without re-throwing, so we can't implement that here.

Hi @dqbd Could you help this one?

PR #1645, opened 2025-09-10
chore(api): add description field for assistants

pjrule:
This adds support for setting/updating the optional `description` field for assistants (matching the Python API implementation). 

(Depends on type updates in #1644)

### DIFF
--- a/libs/langgraph-api/package.json
+++ b/libs/langgraph-api/package.json
@@ -86,7 +86,7 @@
     "@langchain/core": "^0.3.59 || ^1.0.0-alpha",
     "@langchain/langgraph": "^0.2.57 || ^0.3.0 || ^0.4.0 || ^1.0.0-alpha",
     "@langchain/langgraph-checkpoint": "~0.0.16 || ^0.1.0",
-    "@langchain/langgraph-sdk": "^0.1.2",
+    "@langchain/langgraph-sdk": "~0.0.70 || ^0.1.0",
     "typescript": "^5.5.4"
   },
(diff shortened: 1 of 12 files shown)


libs/langgraph-api/package.json
```diff
@@ -86,7 +86,7 @@
     "@langchain/core": "^0.3.59 || ^1.0.0-alpha",
     "@langchain/langgraph": "^0.2.57 || ^0.3.0 || ^0.4.0 || ^1.0.0-alpha",
     "@langchain/langgraph-checkpoint": "~0.0.16 || ^0.1.0",
-    "@langchain/langgraph-sdk": "~0.0.70 || ^0.1.0",
+    "@langchain/langgraph-sdk": "^0.1.2",
```
dqbd: Is this necessary?

pjrule: I believe it's necessary to consume the type updates in the SDK (and therefore necessary to guarantee that `yarn typecheck` passes). currently getting this error with an older SDK version:

```
src/storage/ops.mts:339:9 - error TS2353: Object literal may only specify known properties, and 'description' does not exist in type 'AssistantCreate'.

339         description: options.description,
            ~~~~~~~~~~~

src/storage/ops.mts:410:9 - error TS2353: Object literal may only specify known properties, and 'description' does not exist in type 'AssistantUpdate'.

410         description: options?.description,
            ~~~~~~~~~~~


Found 2 errors in the same file, starting at: src/storage/ops.mts:339
```

PR #1656, opened 2025-09-12
fix(sdk): handle RemoveMessage when received in messages-tuple stream

### LINKED ISSUE #1652: bug: useStream should support streaming RemoveMessage
### Privileged issue

- [x] I am a LangGraph.js maintainer, or was asked directly by a LangGraph.js maintainer to create an issue here.

### Issue Content

RemoveMessage is currently ignored by useStream. We should remove the message when RemoveMessage is received

dqbd:
Closes #1652

### DIFF
--- a/.changeset/itchy-lights-float.md
+++ b/.changeset/itchy-lights-float.md
@@ -1,0 +1,5 @@
+---
+"@langchain/langgraph-sdk": patch
+---
+
+Add support for streaming of RemoveMessage in useStream


timcharper:
This change is supposed to be released, now, right? I'm still getting this error with `langchain 0.3.35` and `@langchain/langgraph 0.4.9`, which, from what I can tell, was released 6 days ago

PR #1659, opened 2025-09-12
fix(langgraph): handle `toSerialized()` if present to handle V1 changes

dqbd:


libs/checkpoint/src/serde/jsonplus.ts
```diff
@@ -123,14 +123,24 @@ function _default(obj: any): any {
...
+
 export class JsonPlusSerializer implements SerializerProtocol {
   protected _dumps(obj: any): Uint8Array {
     const encoder = new TextEncoder();
-    return encoder.encode(
-      stringify(obj, (_: string, value: any) => {
-        return _default(value);
-      })
-    );
+    return encoder.encode(stringify(obj, _defaultReplacer));
```
hntrl, line 143: nit: can we just make this a condition inside of `_default`?

libs/checkpoint/src/serde/jsonplus.ts
```diff
@@ -123,14 +123,24 @@ function _default(obj: any): any {
...
 }
 
+function _defaultReplacer(this: any, key: string, value: any) {
+  const original = this[key] as unknown;
+  if (
+    typeof original === "object" &&
+    original != null &&
+    "toSerialized" in original &&
+    typeof original.toSerialized === "function"
+  ) {
```
hntrl, line 133: would it be helpful to export a a `isSerializable` fn from core?

dqbd: Maybe after we release `@langchain/core`

libs/checkpoint/src/serde/jsonplus.ts
dqbd: No, since we need the pre-serialized object

PR #1696, opened 2025-09-23
chore(langgraph): rename internal type declarations

dqbd:


libs/langgraph/src/graph/graph.ts
```diff
@@ -598,13 +591,13 @@ export class CompiledGraph<
   CommandType,
   StreamCustomType
 > {
-  declare NodeType: N;
+  declare "~NodeType": N;
 
   declare "~NodeReturnType": NodeReturnType;
 
-  declare RunInput: State;
+  declare "~RunInput": Update;
```
dqbd, line 598: This was previously incorrect.

PR #1701, opened 2025-09-25
fix(langgraph-checkpoint-redis): allow alpha version of 1.0.0

baptistejamin:
Allows using langgraph-checkpoint-redis with alpha 1.0.0


dqbd:
Thank you!

PR #1719, opened 2025-10-04
fix(docs): remove incorrect comment

timcharper:
The comment on the "Add function that invokes subgraph" code example is incorrect; bar is not shared with the parent subgraph state.


dqbd:
Thank you!

PR #1726, opened 2025-10-16
chore: update oss API spec to match python

isahers1:


libs/langgraph-api/src/api/threads.mts
```diff
@@ -68,7 +68,21 @@ api.post(
...
+      c.res.headers.set(
+        "X-Pagination-Total",
+        ((payload.offset ?? 0) + total + 1).toString()
+      );
+    } else {
+      c.res.headers.set(
+        "X-Pagination-Total",
+        ((payload.offset ?? 0) + total).toString()
+      );
+    }
```
hntrl, line 77: nit:
```suggestion
    const nextOffset = (payload.offset ?? 0) + total
    if (total === payload.limit) {
      c.res.headers.set(
        "X-Pagination-Next",
        nextOffset.toString()
      );
      c.res.headers.set(
        "X-Pagination-Total",
        (nextOffset + 1).toString()
      );
    } else {
      c.res.headers.set(
        "X-Pagination-Total",
        nextOffset.toString()
      );
    }
```

hntrl (dismissed):

PR #1766, opened 2025-10-30
feat(prebuilt): port LangChain tool node changes back to LangGraph

christian-bromann:
During our v1 push we've added several changes to the `ToolNode` that we now want to port back. The `langchain` implementation of `ToolNode` for `createAgent` will remain in the `langchain` package as it contains additional changes related to middleware that aren't relevant to LangGraph.

### 🛡️ Enhanced Error Handling
- **Function-based error handling**: `handleToolErrors` now supports custom error handling functions
  ```ts
  new ToolNode(tools, {
    handleToolErrors: (error, toolCall) => {
      // Custom error handling logic
      return new ToolMessage({ ... });
    }
  });
  ```
- **ToolInvocationError**: New dedicated error class for tool invocation failures with detailed context
- **Improved error messages**: Better formatting and context in error responses

### ⏹️ Abort Signal Support
- **Signal propagation**: Added `signal` option to `ToolNodeOptions` for cancellation support
- **Signal merging**: Automatically combines node-level and config-level abort signals
- **Graceful cancellation**: Proper handling of aborted operations

### 🔧 Enhanced Type Safety
- **Generic type parameters**: `ToolNode<StateSchema, ContextSchema>` for better type inference
- **Improved type checking**: Uses `BaseMessage.isInstance()` and `AIMessage.isInstance()` for runtime type safety
- **Better annotation support**: Enhanced compatibility with LangGraph annotation system

### Updated Interface
```ts
export interface ToolNodeOptions {
  name?: string;
  tags?: string[];
  signal?: AbortSignal;  // 🆕 NEW
  handleToolErrors?:     // 🔄 ENHANCED
    | boolean
    | ((error: unknown, toolCall: ToolCall) => ToolMessage | undefined);
}
```

### Enhanced Constructor
```ts
constructor(
  tools: (StructuredToolInterface | DynamicTool | RunnableToolLike)[],
  public options?: ToolNodeOptions  // 🔄 Now public
)
```

### DIFF
--- a/.changeset/curly-chairs-kiss.md
+++ b/.changeset/curly-chairs-kiss.md
@@ -1,0 +1,5 @@
+---
+"@langchain/langgraph": minor
+---
+
+Add `handleToolErrors` callback and `wrapToolCall` method for ToolNode. 


libs/langgraph-core/src/prebuilt/tool_node.ts
```diff
@@ -1,26 +1,173 @@
...
+      `Error invoking tool '${toolCall.name}' with kwargs ${toolArgs} with error: ${error.stack}\n Please fix the error and try again.`
+    );
+
+    this.toolCall = toolCall;
+    this.toolError = error;
+  }
+}
+
+// Create a mergeAbortSignals function since the import might not work
+function mergeAbortSignals(
```
dqbd, line 125: Why would the import not work here?

libs/langgraph-core/src/prebuilt/tool_node.ts
```diff
@@ -150,80 +297,246 @@ const isSendInput = (input: unknown): input is { lg_tool_call: ToolCall } =>
...
+          tool_call_id: toolCall.id!,
+          status: "success",
+        });
+      } catch (e: unknown) {
+        /**
+         * Handle errors from tool execution (not from wrapToolCall)
+         * If tool invocation fails due to input parsing error, throw a {@link ToolInvocationError}
+         */
+        if (e instanceof ToolInputParsingException) {
+          throw new ToolInvocationError(e, toolCall);
```
dqbd, line 466: _question_: Couldn't we just rethrow `ToolInputParsingException`? Do we need to introduce a new error class?

libs/langgraph-core/src/prebuilt/tool_node.ts
```diff
@@ -150,80 +297,246 @@ const isSendInput = (input: unknown): input is { lg_tool_call: ToolCall } =>
...
+          content: typeof output === "string" ? output : JSON.stringify(output),
+          tool_call_id: toolCall.id!,
+          status: "success",
+        });
+      } catch (e: unknown) {
+        /**
+         * Handle errors from tool execution (not from wrapToolCall)
+         * If tool invocation fails due to input parsing error, throw a {@link ToolInvocationError}
+         */
+        if (e instanceof ToolInputParsingException) {
```
dqbd, line 465: Is there a way how to determine `ToolInputParsingException` without `instanceof`? This will break if there are multiple `@langchain/core` versions imported in runtime

christian-bromann:
I think it isn't worth continuing here since we don't see many people in need for it. Will go ahead and close.

PR #1776, opened 2025-11-12
feat(sdk): add vue, svelte, angular implementation of useStream hook

dqbd:


libs/sdk-svelte/src/index.ts
```diff
@@ -0,0 +1,704 @@
...
+        threadState: $branchContext.threadHead,
+        error: $streamError,
+      });
+    }
+  );
+
+  const interrupts = derived(
+    [streamValues, streamError, branchContext, isLoading],
+    ([$streamValues, $streamError, $branchContext, $isLoading]) => {
+      const vals = $streamValues ?? get(historyValues);
```
pawel-twardziak, line 611: shouldn't `historyValues` be defined as a dependency?

pawel-twardziak: hmmm `branchContext` is already there and it might be not an issue

pawel-twardziak:
@christian-bromann  I'm having a hard time trying to do a CR here on the platform :disappointed: it is so sluggish... 

Svelte/Angular custom stream don't react to threadId changes

- libs/sdk-svelte/src/stream.custom.ts:38
- libs/sdk-angular/src/stream.custom.ts:51

patch:

[Changes-On-c49af8.patch](https://github.com/user-attachments/files/25502716/Changes-On-c49af8.patch)

libs/sdk-angular/src/tests/components/SwitchThread.ts
```diff
@@ -0,0 +1,83 @@
...
+          ],
+        },
+      };
+    }
+    return generate();
+  },
+};
+
+@Component({
+  standalone: true,
```
pawel-twardziak: it's standalone by default in v20

christian-bromann: makes sense, removing

christian-bromann:
Closing in favor of #2001

PR #1783, opened 2025-11-19
Add `values` parameter to thread search

andrewnguonly:
### Summary
Thread search supports a `values` parameter, which is an arbitrary map to filter the `values` field.

API reference: https://langchain-ai.github.io/langgraph/cloud/reference/api/api_ref.html?_gl=1*p8hdx4*_gcl_au*NTI4MjM5MjguMTc2MjQ2NDgzMQ..*_ga*MTY2MTYxNDUxMi4xNzYzMDYyMDYw*_ga_47WX3HKKY2*czE3NjM1ODkxOTEkbzE3JGcxJHQxNzYzNTkwNTM1JGo1NSRsMCRoMA..#tag/threads/post/threads/search

Reference implementation in Python SDK: https://github.com/langchain-ai/langgraph/blob/main/libs/sdk-py/langgraph_sdk/client.py#L1415

### DIFF
--- a/.changeset/unlucky-experts-suffer.md
+++ b/.changeset/unlucky-experts-suffer.md
@@ -1,0 +1,5 @@
+---
+"@langchain/langgraph-sdk": patch
+---
+
+Add `values` parameter to thread search


dqbd (approved):

amitchaudharyc:
@dqbd It would have been great if we could get only first N messages instead of the whole list of messages.

PR #1785, opened 2025-11-21
feat: Add search by `name` functionality to in-memory JS agent server

andrewnguonly:
### Summary
We're adding functionality to search for assistants by name. Example:
```
POST /assistants/search
{
    ...
    "name": "foo",
    ...
}
```
Functions like `WHERE name ILIKE '%foo%'`.

The `AssistantRepo` interface needs to be updated first, since it's a dependency of the JS server implementation.

### DIFF
--- a/.changeset/chilly-forks-sort.md
+++ b/.changeset/chilly-forks-sort.md
@@ -1,5 +1,6 @@
 ---
 "@langchain/langgraph-api": patch
+"@langchain/langgraph-sdk": patch
 ---
 
 Adding functionality to search assistants by name in the in-memory server implementation.


.changeset/chilly-forks-sort.md
```diff
@@ -0,0 +1,5 @@
+---
+"@langchain/langgraph-api": patch
```
dqbd, line 2: Either create a new changeset for SDK or amend `chilly-forks-sort.md` with `"@langchain/langgraph-sdk": patch` to make sure we're releasing both of the packages at the same time

PR #1788, opened 2025-11-22
chore: fix AsyncBatchedStore listNamespaces

hinthornw:
Couple things:
- Add support for listNamespaces in the asyncbatchstore
- Add tests
- Also think we should let you directly call batch()? I don't see any good reason to throw an error there.


libs/checkpoint/src/store/batch.ts
```diff
@@ -51,20 +54,13 @@ export class AsyncBatchedStore extends BaseStore {
...
+    operations: Op
   ): Promise<OperationResults<Op>> {
-    throw new Error(
-      "The `batch` method is not implemented on `AsyncBatchedStore`." +
-        "\n Instead, it calls the `batch` method on the wrapped store." +
-        "\n If you are seeing this error, something is wrong."
+    const promises = operations.map((operation) =>
+      this.enqueueOperation(operation)
     );
+    return Promise.all(promises) as Promise<OperationResults<Op>>;
```
jdrogers940, line 63: Might want to use `Promise.allSettled(promises)` so we get a result for each operation in the batch? Otherwise could just get the one rejected one and that's probably a weird UX to figure out which one (and status of others).

Probably worth adding a failure test around this.

libs/checkpoint/src/store/batch.ts
```diff
@@ -95,6 +91,7 @@ export class AsyncBatchedStore extends BaseStore {
     key: string,
     value: Record<string, any>
   ): Promise<void> {
+    validateNamespace(namespace);
```
jdrogers940, line 94: Should we be doing this in other methods too?

jdrogers940 (approved):

libs/checkpoint/src/store/batch.ts
```diff
@@ -51,20 +54,13 @@ export class AsyncBatchedStore extends BaseStore {
...
-   */
   async batch<Op extends Operation[]>(
-    _operations: Op
+    operations: Op
   ): Promise<OperationResults<Op>> {
-    throw new Error(
-      "The `batch` method is not implemented on `AsyncBatchedStore`." +
-        "\n Instead, it calls the `batch` method on the wrapped store." +
-        "\n If you are seeing this error, something is wrong."
+    const promises = operations.map((operation) =>
```
dqbd, line 60: I think we need to handle data dependencies here?

PR #1793, opened 2025-11-25
feat(js-sdk): skip auto loading api key when set to null on sdk client create

jdrogers940:
**Description:** There are times a user might want to create the client, but conditionally set the API key. For example, consider a complex auth situation where the system has user callers using jwts and system callers using API keys. This allows explicitly disabling the auto-loading behavior of API keys in the client today, so no key is set.
**Issue:** N/A
**Dependencies:** None
**Twitter handle:** N/A

### DIFF
--- a/libs/sdk/src/tests/fetch.test.ts
+++ b/libs/sdk/src/tests/fetch.test.ts
@@ -211,7 +211,21 @@ describe.each([["global"], ["mocked"]])(
         const client = new Client();
         await (client.threads as any).fetch("/test");
 
-        expect(expectedFetchMock).toHaveBeenCalledWith(
+        expect(expectedFetchMock).toHaveBeenNthCalledWith(
+          1,
+          expect.any(URL),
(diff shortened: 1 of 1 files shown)


libs/sdk/src/tests/fetch.test.ts
```diff
@@ -197,5 +198,78 @@ describe.each([["global"], ["mocked"]])(
...
+    describe("API key auto-load", () => {
+      it("should auto-load API key from environment when apiKey is undefined", async () => {
+        const getEnvSpy = vi
+          .spyOn(envUtils, "getEnvironmentVariable")
+          .mockImplementation((name: string) => {
+            if (name === "LANGGRAPH_API_KEY") return "env-api-key";
+            return undefined;
+          });
+
+        const client = new Client();
```
dqbd, line 211: _nit_: would test with `new Client({ apiKey: undefined })` as well

PR #1814, opened 2025-12-08
Bump zod on remaining packages

baptistejamin:
Hey there.

We are updating our codebase from 0.x to 1.x, and some remaining packages we are using are still pinning the previous zod version.

Just aligned packages still requiring zod v3


hntrl (approved):
thanks @baptistejamin!

PR #1817, opened 2025-12-11
fix(sdk): add support for retries on issues during streaming

jdrogers940:
Adds retries to the streaming endpoints of the JS client.

### DIFF
--- a/libs/sdk/src/client.ts
+++ b/libs/sdk/src/client.ts
@@ -359,7 +359,9 @@ class BaseClient {
    * Protected helper for streaming with automatic retry logic.
    * Handles both initial requests and reconnections with SSE.
    */
-  protected async *streamWithRetry<T extends { id?: string; event: string; data: unknown }>(config: {
+  protected async *streamWithRetry<
+    T extends { id?: string; event: string; data: unknown }
+  >(config: {
     endpoint: string;
     method?: string;
     signal?: AbortSignal;
@@ -367,7 +369,11 @@ class BaseClient {
     params?: Record<string, unknown>;
     json?: unknown;
     maxRetries?: number;
-    onReconnect?: (options: { attempt: number; lastEventId?: string; cause: unknown }) => void;
+    onReconnect?: (options: {
+      attempt: number;
+      lastEventId?: string;
+      cause: unknown;
+    }) => void;
     onInitialResponse?: (response: Response) => void | Promise<void>;
   }): AsyncGenerator<T> {
     const makeRequest = async (reconnectParams?: StreamRequestParams) => {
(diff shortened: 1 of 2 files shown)


libs/sdk/src/utils/stream.ts
```diff
@@ -1,6 +1,182 @@
...
+      const { response, stream } = isInitialRequest
+        ? await initialRequest()
+        : await reconnectRequest(lastEventId!, reconnectPath);
+
+      isInitialRequest = false;
+
+      // Check for Location header (server-provided reconnection path)
+      const locationHeader = response.headers.get("location");
+      if (locationHeader) {
+        reconnectPath = locationHeader;
```
dqbd, line 94: _question_: shouldn't we avoid reconnecting if the Location header is not present?

jdrogers940: Yeah you're right, don't want to kick off a new run

libs/sdk/src/utils/stream.ts
```diff
@@ -1,6 +1,187 @@
...
+  let isInitialRequest = true;
+
+  while (true) {
+    let shouldRetry = false;
+    let lastError: unknown;
+    let reader: ReadableStreamDefaultReader<T> | undefined;
+
+    try {
+      // Check if aborted before making request
+      if (options.signal?.aborted) return;
```
hinthornw, line 84: ooc: should this be before the try block?

libs/sdk/src/client.ts
```diff
@@ -1297,26 +1338,64 @@ export class RunsClient<
...
+    // Reconnect with GET request and Last-Event-ID header
+    // Uses reconnectPath (from Location header) if provided, otherwise original endpoint
+    const reconnectRequest = async (
+      lastEventId: string,
+      reconnectPath?: string
+    ) => {
+      const reconnectEndpoint = reconnectPath || endpoint;
+
+      // For reconnection, use GET (no body/content-type headers)
+      let [url, init] = this.prepareFetchOptions(reconnectEndpoint, {
```
hinthornw: There's so much shared btwn these two functions, I'd challenge you to see if we can refactor to write the code once and then curry the function for the initial call or something

libs/sdk/src/utils/stream.ts
jdrogers940: I think ok to do in either spot. Looks like just a boolean and we'd have to call a specific method to throw: https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal

libs/sdk/src/client.ts
jdrogers940: Yeah completely agreed - doing this in the background but don't want to block on this either, can be fast follow if needed

jdrogers940: Updated - kept one callback but now it's a helper on the base client so single use call everywhere.

dqbd (approved):

PR #1826, opened 2025-12-17
feat(sdk): add type-safe tool call streaming with agent type inference

christian-bromann:
This PR upgrades the LangGraph JS SDK’s React streaming DX with **type-safe tool call handling** and **agent-based type inference**, plus a set of related SDK reliability + core Zod interop improvements. It also overhauls `examples/ui-react` into a navigable demo app showcasing common streaming/branching patterns.

### Type-Safe Tool Call Streaming (React)

- **Agent type inference**: You can now pass `typeof agent` (from `createAgent`) to `useStream` and get **tool call name/args inferred from the agent’s tools** via `InferAgentToolCalls`.
- **Canonical tool-call typing for graphs**: For non-agent usage, tool call types are extracted from your state’s messages:
  - Define `messages: Message<MyToolCalls>[]` in your state
  - `GetToolCallsType<StateType>` is derived from that `messages` field
- **New tool call helpers on the stream**:
  - `stream.toolCalls`: tool calls paired with results as `ToolCallWithResult<ToolCall>[]` with lifecycle state (`pending` | `completed` | `error`)
  - `stream.getToolCalls(aiMessage)`: tool calls scoped to a specific AI message
- **New utility + exported types**:
  - `getToolCallsWithResults(messages)`
  - `ToolCallWithResult`, `ToolCallState`, plus helper types `ToolCallFromTool` / `ToolCallsFromTools` for inferring tool-call unions from tools

### SDK Reliability Improvements

- **Better connection errors**: When the SDK can’t reach the LangGraph server, it now throws a clearer `ConnectionError` message (instead of opaque fetch/network errors).

### Core: Improved Zod Interop

- **Zod v3/v4 plugin robustness**: Updates to the LangGraph Zod plugin to improve interop across Zod versions and reduce prototype extension edge cases.

### Example Application Overhaul (`examples/ui-react`)

Refactors the React example into a proper app shell (layout/sidebar/registry/components) and adds/updates examples:

| Example | Description |
|---------|-------------|
| `tool-calling-agent` | Basic tool-calling agent demo |
| `human-in-the-loop` | HITL pattern with approval flows |
| `multi-step-graph` | Multi-step state graph workflow |
| `parallel-research` | Parallel research tasks with streaming |
| `reasoning-agent` | Reasoning display example |
| `summarization-agent` | Summarization middleware pattern |
| `custom-streaming` | Custom event streaming with progress UI |
| `session-persistence` | Thread/session persistence patterns |
| `branching-chat` | Branching UX with branch switching |

## Usage

**With `createAgent` (recommended):**

```ts
import { agent } from "./agent";

function Chat() {
    assistantId: "agent",
  });

  // stream.toolCalls[0].call.name is typed from your tools
  // stream.toolCalls[0].call.args is typed from the tool schema/_call signature
}
```

**With a custom graph state (tool calls typed via `messages`):**

```ts
import type { Message } from "@langchain/langgraph-sdk";

type MyToolCalls =
  | { name: "get_weather"; args: { location: string }; id?: string }
  | { name: "search"; args: { query: string }; id?: string };

interface MyState {
  messages: Message<MyToolCalls>[];
}

const stream = useStream<MyState>({ assistantId: "my-graph" });
```

## Breaking Changes

None expected at runtime. (Tool-call typing is now derived from `StateType["messages"]` for graph-state usage, and inferred from `typeof agent` for `createAgent` usage.)

### DIFF
--- a/internal/environment_tests/test-exports-tsc/package.json
+++ b/internal/environment_tests/test-exports-tsc/package.json
@@ -23,8 +23,5 @@
     "@types/node": "^18.15.11",
     "typescript": "5.5.4",
     "zod": "^3.25.76"
-  },
-  "overrides": {
-    "zod": "$zod"
   }
 }
--- a/internal/environment_tests/test-exports-tsc/src/zod.ts
+++ b/internal/environment_tests/test-exports-tsc/src/zod.ts
@@ -32,8 +32,10 @@ const WeatherSchema = z.object({
 });
 type Weather = z.infer<typeof WeatherSchema>;
 
+// @ts-ignore - zod version mismatch between test and @langchain/core
 const WeatherGraphState = z.object({
   messages: z.array(z.any()),
+  // @ts-ignore - zod version mismatch between test and @langchain/core
   weather: withLangGraph(WeatherSchema, {
     default: () => ({ temperature: 0, condition: "unknown", location: "" }),
     reducer: {
@@ -44,6 +46,7 @@ const WeatherGraphState = z.object({
 
 type WeatherGraphStateType = z.infer<typeof WeatherGraphState>;
 
+// @ts-ignore - zod version mismatch between test and @langchain/core
 const weatherGraph = new StateGraph(WeatherGraphState)
   .addNode("get_weather", async (state: WeatherGraphStateType) => {
     return {


examples/ui-react/package.json
```diff
@@ -12,20 +12,24 @@
...
   "dependencies": {
-    "@langchain/core": "^1.0.1",
+    "@langchain/core": "1.2.0-dev-1765937705265",
     "@langchain/langgraph": "workspace:*",
     "@langchain/langgraph-sdk": "workspace:*",
-    "@langchain/openai": "^1.0.0",
+    "@langchain/openai": "^1.2.0",
+    "langchain": "2.0.0-dev-1765937705265",
+    "lucide-react": "^0.561.0",
     "react": "^19.0.0",
```
hntrl: are we waiting for langchain release to bump these to dev?

examples/ui-react/src/examples/custom-streaming/agent.ts
```diff
@@ -0,0 +1,186 @@
...
+      // The 'id' field makes this part persistent (added to message.parts)
+      config.writer?.({
+        type: "progress",
+        id: analysisId, // Same ID to update the progress in place
+        step: steps[i].step,
+        message: steps[i].message,
+        progress: Math.round(((i + 1) / steps.length) * 100),
+        totalSteps: steps.length,
+        currentStep: i + 1,
+      } satisfies ProgressData);
```
hntrl, line 55: wonder if we need to make stream writers type safe. Can be a separate push, but this would be good for type safety outside inside of StateGraph also

examples/ui-react/src/examples/parallel-research/agent.ts
```diff
@@ -0,0 +1,255 @@
...
+  model: "gpt-4o-mini",
+  temperature: 0.9,
+});
+const practicalModel = new ChatOpenAI({
+  model: "gpt-4o-mini",
+  temperature: 0.5,
+});
+
+// Define the state annotation with reducer for messages
+const StateAnnotation = Annotation.Root({
```
hntrl: can we use zod here?

libs/sdk/package.json
```diff
@@ -22,9 +22,9 @@
   },
   "license": "MIT",
   "dependencies": {
-    "p-queue": "^6.6.2",
-    "p-retry": "4",
-    "uuid": "^9.0.0"
+    "p-queue": "^9.0.1",
+    "p-retry": "^7.1.1",
```
hntrl, line 26: should be super careful with this change, we've learned our lesson before

libs/sdk/src/react/stream.custom.tsx
```diff
@@ -244,9 +251,28 @@ export function useStreamCustom<
...
+    get messages(): Message<ToolCallType>[] {
+      if (!stream.values) return [];
+      return getMessages(stream.values) as Message<ToolCallType>[];
+    },
+
+    get uiMessages() {
+      if (!stream.values) return [];
+      const msgs = getMessages(stream.values) as Message<ToolCallType>[];
+      return getUIMessagesWithReasoning<ToolCallType>(msgs);
+    },
```
hntrl, line 250: why are these separate?

libs/sdk/src/types.messages.ts
```diff
@@ -93,10 +133,140 @@ export type RemoveMessage = BaseMessage & {
...
+/**
+ * Messages suitable for UI rendering (excludes ToolMessage).
+ * ToolMessages are typically rendered via {@link ToolCallWithResult} instead of directly.
+ * Includes ReasoningMessage for rendering thinking/reasoning content.
+ *
+ * @template ToolCall The type of tool calls for AIMessage, defaults to DefaultToolCall.
+ */
+export type UIMessage<ToolCall = DefaultToolCall> =
+  | Exclude<Message<ToolCall>, ToolMessage>
+  | ReasoningMessage;
```
hntrl: My gut says we'll come to regret adding something like this. Would rather break this out into turns when we get there, and just deal with the rendering awkwardness for a short while

libs/sdk/src/ui/types.ts
```diff
@@ -0,0 +1,519 @@
...
+ * // Use with useStream: useStream<AgentState, Bag>({ ... })
+ * ```
+ */
+export type AgentToBag<T> = {
+  ToolCallsType: InferAgentToolCalls<T>;
+  ConfigurableType?: Record<string, unknown>;
+  InterruptType?: unknown;
+  CustomEventType?: unknown;
+  UpdateType?: unknown;
+};
```
hntrl: Instead of the agent bag type being its own thing, why don't we set these as types on ReactAgent? That way we don't need to have a translation layer between two things that represent the same thing pretty much:

```ts
declare class ReactAgent implements Partial<BagTemplate> {
  ToolCallsType: ...;
  StateType: ...;
}

useStream<TTemplate extends BagTemplate>(...)
// for backwards compat
useStream<GraphState extends Record<string, unknown>, TTemplate extends BagTemplate>(...)
```

That way we can later augment StateGraph to take on basically the same pattern
```ts
declare class CompiledStateGraph implements Partial<BagTemplate> {
  ToolCallsType: ...;
  StateType: ...;
}
```

examples/ui-react-transport/src/server.mts
```diff
@@ -1,17 +1,12 @@
...
 
 import { serve } from "@hono/node-server";
 import { Hono } from "hono";
 
 const llm = new ChatOpenAI({ model: "gpt-4o-mini" });
 
-const schema = z.object({
-  messages: z.custom<BaseMessage[]>().register(registry, MessagesZodMeta),
-});
+const schema = MessagesZodState;
```
dqbd: We need to deprecate this, since this implicitly uses Zod 3

libs/sdk/src/utils/reasoning.ts
hntrl, line 1: This seems like it's achieving the same thing as the block translators in core. Maybe its worth figuring out how to import those and leaning on those instead so we're not having to maintain two sets of the same logic

libs/langgraph-core/src/prebuilt/tool_node.ts
```diff
@@ -6,7 +6,11 @@ import {
   isAIMessage,
 } from "@langchain/core/messages";
 import { RunnableConfig, RunnableToolLike } from "@langchain/core/runnables";
-import { DynamicTool, StructuredToolInterface } from "@langchain/core/tools";
+import {
+  DynamicTool,
+  DynamicStructuredTool,
```
dqbd: _nit_: fix unused import

libs/sdk/src/types.reasoning.ts
hntrl, line 1: Any chance we can employ standard content here?

examples/ui-react/package.json
christian-bromann: Yep, this PR shouldn't be merged before releasing langchain

examples/ui-react/src/examples/custom-streaming/agent.ts
christian-bromann: Yep, type safe custom events + interrupt schemas is a great next push.

libs/sdk/package.json
christian-bromann: I checked the `@langchain/langgraph-sdk` doesn't export for CJS with Node Classic import.

libs/sdk/src/types.messages.ts
christian-bromann: As @dqbd suggested, we can make it experimental.

libs/sdk/src/react/stream.lgp.tsx
```diff
@@ -668,13 +674,32 @@ export function useStreamLGP<
       return interrupts.at(-1) as Interrupt<InterruptType> | undefined;
     },
 
-    get messages() {
+    get messages(): Message<ToolCallType>[] {
       trackStreamMode("messages-tuple", "values");
-      return getMessages(values);
+      return getMessages(values) as Message<ToolCallType>[];
```
dqbd: Can we avoid this typecast?

libs/sdk/src/react/utils.ts
```diff
@@ -0,0 +1,96 @@
...
+        if (
+          typed.type === "text" &&
+          typeof typed.text === "string" &&
+          typed.text.trim().length > 0
+        ) {
+          return true;
+        }
+      }
+    }
+  }
```
dqbd: I think as part of SDK, we should coerce `string` content to text blocks?

examples/ui-react/src/components/MessageInput.tsx
```diff
@@ -0,0 +1,85 @@
...
+export function MessageInput({
+  onSubmit,
+  disabled = false,
+  placeholder = "Send a message...",
+}: MessageInputProps) {
+  const textareaRef = useRef<HTMLTextAreaElement>(null);
+
+  const handleTextareaInput = (e: React.FormEvent<HTMLTextAreaElement>) => {
+    const target = e.target as HTMLTextAreaElement;
+    target.style.height = "auto";
```
dqbd: This could be replaced with a single tailwind class imo: `field-sizing-content`

examples/ui-react/src/components/MessageBubble.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  // Don't render if there's no content
+  if (!content) return null;
+
+  if (isReasoning) {
+    return <ReasoningBubble message={message} isStreaming={false} />;
+  }
+
+  return (
+    <div className="animate-fade-in">
```
dqbd, line 74: _nit_ it would be more readable if we didn't interleave styles for all kinds of messages

libs/sdk/src/react/stream.custom.tsx
christian-bromann: `messages` was there before (the diff is weird here). `uiMessages` was introduced to be a more optimized set of data structure for rendering in UI. More concretely `uiMessages`:

- don't carry `ToolMessage` as we ask user to get them from `toolCalls` see line 265
- carry a `ReasoningMessage` to populate reasoning tokens

examples/ui-react/src/examples/custom-streaming/agent.ts
dqbd: I think we already do for LangGraph?

examples/ui-react/src/examples/custom-streaming/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+              icon={Radio}
+              title="Custom Streaming Events"
+              description="Demonstrate streaming custom data from LangGraph nodes to the UI. Watch progress bars, status updates, and file operations stream in real-time as tools execute."
+              suggestions={CUSTOM_STREAMING_SUGGESTIONS}
+              onSuggestionClick={handleSubmit}
+            />
+          ) : (
+            <div className="flex flex-col gap-6">
+              {stream.uiMessages.map((message, idx) => (
+                <MessageBubble key={message.id ?? idx} message={message} />
```
dqbd, line 148: _question_: could we always ensure that we have `message.id`?

examples/ui-react/src/examples/human-in-the-loop/index.tsx
```diff
@@ -0,0 +1,296 @@
...
+
+  const messagesEndRef = useRef<HTMLDivElement>(null);
+  const [isProcessing, setIsProcessing] = useState(false);
+
+  /**
+   * Auto-scroll to bottom when new messages arrive
+   */
+  useEffect(() => {
+    messagesEndRef.current?.scrollIntoView({ behavior: "smooth" });
+  }, [stream.uiMessages, stream.isLoading, stream.interrupt]);
```
dqbd: Use `use-stick-to-bottom` rather than this

libs/sdk/src/react/tests/stream.test-d.ts
```diff
@@ -0,0 +1,85 @@
+import { describe, test, expectTypeOf } from "vitest";
+
+import type { Message } from "../../types.messages.js";
+import { useStream } from "../stream.js";
+
+describe("useStream", () => {
+  test("should properly type tool calls with explicit ToolCallsType", () => {
+    // Define tool call types as a discriminated union
+    type MyToolCalls =
```
dqbd: I think we also want to include the tool response types here?

examples/ui-react/src/examples/custom-streaming/index.tsx
christian-bromann: What would you suggest here? `BaseMessage` is defined as:
```ts
type BaseMessage = {
  additional_kwargs?: MessageAdditionalKwargs | undefined;
  content: MessageContent;
  id?: string | undefined;
  name?: string | undefined;
  response_metadata?: Record<string, unknown> | undefined;
};
```

Do we want to updated `BaseMessage` or create different message interfaces for `uiMessages`?

libs/sdk/src/react/utils.ts
christian-bromann: Should we tackle this in a separate PR?

libs/sdk/src/types.reasoning.ts
christian-bromann: I think this requires some larger effort and maybe move to a v2 design conversation. There are some problems with this:

1. we don't use reasoning standard content blocks in JS (at least not for streamed messages and not in all providers) - so we should have a push on consolidating this and make sure we do support this wherever possible
2. we have to get rid of allowing `content: string` for `AIMessage` otherwise there will be always wird `if (typeof ...` - ideally we should always return content blocks (maybe v2 thing?)

libs/sdk/src/utils/reasoning.ts
christian-bromann: Ideally we want to support reasoning as part of our standard content block. I suggest this as a separate effort.

libs/sdk/src/ui/types.ts
christian-bromann: I've consolidated `AgentToBag` and moved this all into `BagTemplate`.

hntrl (approved):

PR #1827, opened 2025-12-17
fix(langgraph): add resumable stream support to remote graph

jdrogers940:
- Adding support for the `resumableStreams` parameter in remote graph. Without this, streaming retry logic won't be triggered.
- Updated .gitignore files to more generally ignore generated files from builds.

### DIFF
--- a/libs/langgraph-core/src/pregel/remote.ts
+++ b/libs/langgraph-core/src/pregel/remote.ts
@@ -426,7 +426,6 @@ export class RemoteGraph<
       serializedInput = _serializeInputs(input);
     }
 
-    console.log("streamResumable - streamIterator", this.streamResumable);
     for await (const chunk of this.client.runs.stream(
       sanitizedConfig.configurable.thread_id as string,
       this.graphId,


libs/langgraph-core/src/pregel/remote.ts
```diff
@@ -422,6 +426,7 @@ export class RemoteGraph<
       serializedInput = _serializeInputs(input);
     }
 
+    console.log("streamResumable - streamIterator", this.streamResumable);
```
hntrl:
```suggestion
```

libs/langgraph-core/src/pregel/remote.ts
```diff
@@ -435,8 +440,10 @@ export class RemoteGraph<
         streamSubgraphs,
         ifNotExists: "create",
         signal: mergedConfig.signal,
+        streamResumable: this.streamResumable,
       }
     )) {
+      console.log("chunk", chunk);
```
hntrl:
```suggestion
```

hntrl (dismissed):

PR #1833, opened 2025-12-27
feat(create-langgraph): allow user to generate a langgraph config file

christian-bromann:
Extends the `create-langgraph` CLI with a new `config` subcommand that automatically scans a project for LangGraph agents and generates a ready-to-use `langgraph.json` configuration file.

#### Usage

```bash
# Generate config in current directory
npm init langgraph config

# Generate config for a specific project
npm init langgraph config ./path/to/project
```

#### Features

- **Automatic Agent Discovery**: Scans all TypeScript and JavaScript files for LangGraph agents
- **Pattern Detection**: Detects agents created with:
  - `createAgent({ ... })`
  - `new StateGraph(...).compile()`
  - `workflow.compile()` / `builder.compile()`
- **ESM & CJS Support**: Recognizes both module systems:
  - ESM: `export const agent = createAgent(...)`
  - CJS: `module.exports.agent = createAgent(...)` / `exports.agent = createAgent(...)`
- **Export Awareness**: Only includes exported agents in the config, warns about unexported ones with file path and line number
- **Unique Key Generation**: Automatically generates unique keys when multiple agents share the same name (uses parent directory name as fallback)
- **Auto-detection**: Detects Node.js version and `.env` file presence

#### Example Output

```
┌  🦜 create-langgraph config 
│
◇  Found 8 agent(s) in 39 file(s)

⚠️  Found 2 agent(s) that are not exported:
   • internalAgent at src/helpers/agent.ts:45
   • testAgent at src/test-utils.ts:12
   Add export keyword to include them in the configuration.

✓ Created langgraph.json with:
   • agent: ./src/agent.ts:agent
   • weather-agent: ./src/agents/weather/index.ts:agent
   • research-agent: ./src/agents/research/index.ts:agent

└  Configuration created successfully!
```

#### Generated Config

```json
{
  "node_version": "20",
  "graphs": {
    "agent": "./src/agent.ts:agent",
    "weather-agent": "./src/agents/weather/index.ts:agent"
  },
  "env": ".env"
}
```

#### Changes

- **New**: `src/config.ts` - Config generation logic with agent scanning
- **Modified**: `src/cli.ts` - Added `config` subcommand
- **New**: `src/tests/config.test.ts` - Comprehensive test suite (41 tests)
- **Modified**: `package.json` - Added test scripts, upgraded to ESM with `"type": "module"`
- **Modified**: `tsconfig.json` - Exclude test files from build
- Converted all source files from `.mts` to `.ts` extensions

### DIFF
--- a/libs/create-langgraph/README.md
+++ b/libs/create-langgraph/README.md
@@ -29,13 +29,13 @@ Follow the interactive prompts to select a template and configure your project.
 
 Choose from a variety of production-ready templates:
 
-| Template | Description |
-| -------- | ----------- |
-| [**New LangGraph Project**](https://github.com/langchain-ai/new-langgraphjs-project) | A simple, minimal chatbot with memory |
-| [**ReAct Agent**](https://github.com/langchain-ai/react-agent-js) | A flexible agent that can be extended with many tools |
-| [**Memory Agent**](https://github.com/langchain-ai/memory-agent-js) | A ReAct-style agent with persistent memory across conversations |
-| [**Retrieval Agent**](https://github.com/langchain-ai/retrieval-agent-template-js) | An agent with retrieval-based question-answering |
-| [**Data-enrichment Agent**](https://github.com/langchain-ai/data-enrichment-js) | An agent that performs web searches and organizes findings |
+| Template                                                                             | Description                                                     |
+| ------------------------------------------------------------------------------------ | --------------------------------------------------------------- |
+| [**New LangGraph Project**](https://github.com/langchain-ai/new-langgraphjs-project) | A simple, minimal chatbot with memory                           |
+| [**ReAct Agent**](https://github.com/langchain-ai/react-agent-js)                    | A flexible agent that can be extended with many tools           |
+| [**Memory Agent**](https://github.com/langchain-ai/memory-agent-js)                  | A ReAct-style agent with persistent memory across conversations |
+| [**Retrieval Agent**](https://github.com/langchain-ai/retrieval-agent-template-js)   | An agent with retrieval-based question-answering                |
+| [**Data-enrichment Agent**](https://github.com/langchain-ai/data-enrichment-js)      | An agent that performs web searches and organizes findings      |
 
 ### Using a Specific Template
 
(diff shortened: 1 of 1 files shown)


libs/create-langgraph/src/config.ts
```diff
@@ -0,0 +1,350 @@
...
+          Make sure your agents are defined using one of these patterns:
+          - ${color.cyan("createAgent({ ... })")}
+          - ${color.cyan("createReactAgent({ ... })")}
+          - ${color.cyan("new StateGraph(...).compile()")}
+          - ${color.cyan("workflow.compile()")}
+        `
+      )
+    );
+    return;
+  }
```
dqbd, line 216: _note_: would probably instead ask for an export here instead of warning and exiting early

libs/create-langgraph/src/config.ts
```diff
@@ -0,0 +1,350 @@
...
+
+  // Build the graphs config
+  const graphs: Record<string, string> = {};
+  const usedKeys = new Set<string>();
+
+  for (const agent of exportedAgents) {
+    const relativePath = path.relative(cwd, agent.filePath);
```
dqbd: _question_: should we do this from `cwd`? Or should we obtain the directory of root `package.json`? (thinking of monorepo setup).

dqbd (dismissed):

libs/create-langgraph/src/config.ts
christian-bromann: Let's do this in a sep PR, how do you imagine this to look like? Ask if you want to export the agent and if yes, modify the file?

libs/create-langgraph/src/config.ts
christian-bromann: The user can provide a root path to start looking from, e.g. via `npm init langgraph config ../../`. I will push a commit to change the `cwd` variable name to `targetRootPath` to make this more clear.

IMHO using `cwd` is the best default behavior here, users can easily copy the file to a better desired place. I am also not sure if everyone always puts the `langgraph.json` within the project root.

christian-bromann:
Docs PR: https://github.com/langchain-ai/docs/pull/2090

PR #1839, opened 2026-01-04
fix(langgraph): add equals method to channels for semantic comparison

killagu:
When using separate state/input/output schemas that share fields with reducers (e.g., middleware stateSchema in agents), the StateGraph would throw "Channel already exists with a different type" because it compared channel instances by reference.

This fix adds an `equals()` method to BaseChannel and implements it in BinaryOperatorAggregate to compare reducer function references, matching the Python implementation's `__eq__` behavior.

Changes:
- Add `equals()` method to BaseChannel (default: reference equality)
- Implement `equals()` in BinaryOperatorAggregate (compares operator refs)
- Update `_addSchema` to use `equals()` for semantic comparison
- Add unit tests for reducer field sharing across schemas

Fixes # (issue)
https://github.com/langchain-ai/langgraphjs/issues/1820


dqbd:
Thank you @killagu!

dqbd (approved):

dqbd:
Will need to update formatting and do minor tweaks, closing in favour of #1901

PR #1852, opened 2026-01-09
feat(langgraph): add type utilities

hntrl:
Adds utility types for defining typed graph nodes outside of the `StateGraph` builder, improving code organization and editor support.

### New Type Utilities

**`libs/langgraph-core/src/graph/types.ts`** (new file)

- **`ExtractStateType<Schema>`** - Extracts the state type from `StateSchema`, `AnnotationRoot`, or Zod object schemas
- **`ExtractUpdateType<Schema>`** - Extracts the update type (what nodes can return)
- **`GraphNode<Schema, Nodes?, Config?>`** - Type for node functions, with optional typed `Command.goto` routing
- **`ConditionalEdgeRouter<Schema, Nodes>`** - Type for conditional edge routing functions

### Updates

- **`StateSchema`** - Added `Node` type declaration for parity with `AnnotationRoot.Node`
- **`StateGraph`** - Now imports shared `ExtractStateType`/`ExtractUpdateType` from `types.ts`
- **Exports** - New types exported from `graph/index.ts` and `web.ts`

### Tests

- Comprehensive type-level tests in `types.test-d.ts` covering all new utilities with `StateSchema`, `Annotation`, and Zod schemas

---

## Usage

```typescript
import { StateGraph, GraphNode, Annotation } from "@langchain/langgraph";

const AgentState = Annotation.Root({
  messages: Annotation<BaseMessage[]>({ reducer: messagesReducer }),
  step: Annotation<number>({ default: () => 0 }),
});

// Define typed nodes outside the graph builder
const processNode: GraphNode<typeof AgentState> = (state, config) => {
  return { step: state.step + 1 };
};

// With typed Command routing
const routerNode: GraphNode<typeof AgentState, "agent" | "tool"> = (state) => {
  if (state.needsTool) {
    return new Command({ goto: "tool" }); // ✓ Type-safe
  }
  return new Command({ goto: "agent" });
};

// Build graph
const graph = new StateGraph(AgentState)
  .addNode("process", processNode)
  .addNode("router", routerNode)
  .compile();
```

### DIFF
--- a/libs/langgraph-core/src/pregel/loop.ts
+++ b/libs/langgraph-core/src/pregel/loop.ts
@@ -643,7 +643,11 @@ export class PregelLoop {
           // in `algo.ts` we append a bool to the task path to indicate
           // whether or not a call was present. If so, we don't emit the
           // the interrupt as it'll be emitted by the parent.
-          if (task.path?.[0] === PUSH && task.path?.[task.path.length - 1] === true) return;
+          if (
+            task.path?.[0] === PUSH &&
+            task.path?.[task.path.length - 1] === true
+          )
+            return;
 
           const interruptWrites = writes
             .filter((w) => w[0] === INTERRUPT)


libs/langgraph-core/src/tests/types.test-d.ts
```diff
@@ -0,0 +1,349 @@
...
+    };
+
+    // This should compile
+    const graph = new StateGraph(AgentAnnotation)
+      .addNode("myNode", myNode)
+      .addEdge(START, "myNode")
+      .addEdge("myNode", END)
+      .compile();
+
+    expectTypeOf(graph).not.toBeNever();
```
christian-bromann: I am not sure if this type check makes sense, I don't think it can ever be `never`?

libs/langgraph-core/src/tests/types.test-d.ts
```diff
@@ -0,0 +1,349 @@
...
+
+  it("should type node functions correctly", () => {
+    const myNode: GraphNode<typeof ZodState> = (state, _config) => {
+      expectTypeOf(state.count).toEqualTypeOf<number>();
+      expectTypeOf(state.name).toEqualTypeOf<string>();
+
+      return { count: state.count + 1 };
+    };
+
+    expectTypeOf(myNode).toBeFunction();
```
christian-bromann: same, `myNode` seems always to be a  `function` am I right?

libs/langgraph-core/src/graph/types.ts
```diff
@@ -0,0 +1,157 @@
...
+ * const graph = new StateGraph(AgentState)
+ *   .addNode("process", processNode)
+ *   .addNode("router", routerNode)
+ *   .compile();
+ * ```
+ */
+export type GraphNode<
+  Schema,
+  Nodes extends string = string,
+  Config extends LangGraphRunnableConfig = LangGraphRunnableConfig
```
dqbd: _question_: shouldn't it be `Runtime`?

libs/langgraph-core/src/graph/state.ts
```diff
@@ -81,6 +80,8 @@ import type {
   InferInterruptInputType,
 } from "../interrupt.js";
 import type { InferWriterType } from "../writer.js";
+import { ExtractStateType, ExtractUpdateType } from "./types.js";
+import { AnyStateSchema } from "../state/schema.js";
```
dqbd: _mega nit_: 

```suggestion
import type { ExtractStateType, ExtractUpdateType } from "./types.js";
import type { AnyStateSchema } from "../state/schema.js";
```

libs/langgraph-core/src/tests/types.test-d.ts
```diff
@@ -0,0 +1,640 @@
...
+      type State = ExtractStateType<typeof AgentAnnotation>;
+
+      expectTypeOf<State>().toEqualTypeOf<{
+        count: number;
+        name: string;
+      }>();
+    });
+  });
+
+  describe("with Zod object", () => {
```
dqbd, line 61: Can we test with Zod 3 as well? (also need to discuss when we stop supporting Zod 3)

libs/langgraph-core/src/graph/types.ts
```diff
@@ -0,0 +1,157 @@
...
+> = (
+  state: ExtractStateType<Schema>,
+  config: Config
+) =>
+  | ExtractUpdateType<Schema>
+  | Command<unknown, ExtractUpdateType<Schema>, Nodes>
```
dqbd: _question_: might be good to check for `CommandInstance` instead (so that InteliSense won't suggest bogus methods from `Command`)

libs/langgraph-core/src/tests/types.test-d.ts
hntrl: v2 IMO

PR #1860, opened 2026-01-09
fix: use yarn pack and publish

hinthornw:
`changeset publish` doesn't correctly resolve the `workspace:*` dependencies, leading to broken releases.

Confirmed this change *does* resolve the dependency correctly.
I'm less familiar with changeset so may e missing other features they're providing at publication step that may be missing here, however.


dqbd:
Closing since it seems like Yarn doesn't pick up OIDC, superseded by #1862

PR #1861, opened 2026-01-09
Fix reducer lookup failure with duplicate package installations

brettshollenberger:
## Fixes

* https://github.com/langchain-ai/deepagentsjs/issues/75#issuecomment-3713486856
* https://github.com/langchain-ai/langgraphjs/issues/1820

##  Summary

This PR fixes the "LastValue can only receive one value per step" error that occurs when multiple copies of @langchain/langgraph exist in a project's dependency tree.

##  The Problem

`schemaMetaRegistry` is a module-level singleton that stores reducer metadata in a WeakMap:

```typescript
  const schemaMetaRegistry = new SchemaMetaRegistry();
  export { schemaMetaRegistry };
```

When package managers create multiple copies of `@langchain/langgraph` (due to peer dependency resolution differences), each copy gets its own schemaMetaRegistry instance with its own WeakMap:

* Package Copy A: schemaMetaRegistry (WeakMap A) ← reducers registered here
* Package Copy B: schemaMetaRegistry (WeakMap B) ← reducer lookup happens here (not found!)

Result: Reducer lookup fails, falls back to LastValue, parallel tool calls fail with:

```typescript
  InvalidUpdateError: Invalid update for channel "files" with values [{},{},{},{}]:
  LastValue can only receive one value per step.
```

This issue is currently causing the deepagents repo to fail for lots of users on the basic example use cases.

##  Why duplicates occur

Package managers create separate copies when peer dependency versions differ. Common triggers:

* Different zod versions (v3 vs v4 -- deepagents uses 4)
* Different @langchain/core versions

Example from pnpm:

```
  deepagents → @langchain/langgraph@1.0.2 (with zod@4.3.5)
  my-app     → @langchain/langgraph@1.0.2 (with zod@3.25.76)
```

Same version, but two physical copies with separate registries.

##  The Solution

Use Symbol.for() to create a global symbol shared across all package copies, and attach the registry to globalThis:

```typescript
const SCHEMA_META_REGISTRY_KEY = Symbol.for('@langchain/langgraph/schemaMetaRegistry');

  function getGlobalSchemaMetaRegistry(): SchemaMetaRegistry {
    const existing = (globalThis as any)[SCHEMA_META_REGISTRY_KEY];
    if (existing instanceof SchemaMetaRegistry) {
      return existing;
    }

    const registry = new SchemaMetaRegistry();
    (globalThis as any)[SCHEMA_META_REGISTRY_KEY] = registry;
    return registry;
  }

  const schemaMetaRegistry = getGlobalSchemaMetaRegistry();
```

### Why this works

1. Symbol.for('key') returns the same symbol across all package copies (unlike Symbol('key'))
2. All copies look up the registry at globalThis[Symbol.for(...)]
3. First copy to run creates the registry; subsequent copies reuse it
4. One WeakMap shared everywhere = reducers registered anywhere are found everywhere


brettshollenberger:
@christian-bromann I know you'd said you were working on this issue - thanks for your work creating this great library!

christian-bromann:
@brettshollenberger sorry for the delayed response, this issue should be fixed when using the new [StateSchema primitive](https://reference-docs.vercel.app/javascript/langchain-langgraph/index/StateSchema). Can you give that a try?

christian-bromann:
Closing. Happy to answer any questions.

PR #1893, opened 2026-01-13
fix(sdk): infer agent types in stream.values

christian-bromann:
This patch enables proper type propagation of agent states within `useStream`, e.g.:

```ts
import { useStream } from "@langchain/langgraph/react";
import type { agent } from "./agent";

const stream = useStream<typeof agent>({ ... })
stream.values // no properly reflects the agent type
```

### DIFF
--- a/.changeset/clean-dogs-fly.md
+++ b/.changeset/clean-dogs-fly.md
@@ -1,5 +1,4 @@
 ---
-"@examples/ui-react": patch
 "@langchain/langgraph-sdk": patch
 ---
 
--- a/libs/sdk/src/tests/stream.test-d.ts
+++ b/libs/sdk/src/tests/stream.test-d.ts
@@ -1,5 +1,5 @@
 import { describe, test, expectTypeOf } from "vitest";
-import { z } from "zod";
+import { z } from "zod/v4";
 import type { Message } from "../types.messages.js";
 import type {
   InferAgentState,


.changeset/clean-dogs-fly.md
```diff
@@ -0,0 +1,6 @@
+---
+"@examples/ui-react": patch
```
dqbd: _nit_: we should not add packages from `@examples` to changesets

libs/sdk/src/tests/stream.test-d.ts
```diff
@@ -0,0 +1,335 @@
+import { describe, test, expectTypeOf } from "vitest";
+import { z } from "zod";
```
dqbd: _nit_: Good to specify exact version of `zod`?

libs/sdk/src/ui/types.ts
```diff
@@ -64,16 +65,145 @@ export type ExtractAgentConfig<T> = T extends { "~agentTypes": infer Config }
...
+ * Minimal interface to structurally match AgentMiddleware from langchain.
+ * We can't import AgentMiddleware due to circular dependencies, so we match
+ * against its structure to extract type information.
  */
-type InferSchemaInput<S> = S extends { _zod: { input: infer Args } }
-  ? Args
-  : S extends { _input: infer Args }
-  ? Args
-  : never;
+export interface AgentMiddlewareLike<
```
dqbd, line 79: _question_: couldn't we do merging of middleware state schema within `langchain` instead inside `sdk`?

christian-bromann: Yes ... but not yet. Eventually these type helpers should live in LC core and imported where needed. However I want to get an overview first of how these primitives look like before making a PR to core.

PR #1903, opened 2026-01-16
feat(sdk): add multi-subagent tracking to useStream

christian-bromann:
This PR adds comprehensive streaming enhancements to the `useStream` hook, enabling developers to:
- Track **node executions** in any graph
- Monitor **tool calls** with typed arguments in agents
- Display **multiple concurrent subagent executions** in deep agent scenarios

## Stream Type Hierarchy

The PR introduces a layered type system that automatically resolves based on your agent/graph type:

```
UseStreamOptions / BaseStream              ← Original useStream (unchanged)
    │
    ├── UseAgentStreamOptions / UseAgentStream
    │       └── Extends UseGraphStream
    │       Input:  (no new options)
    │       Output: + toolCalls, getToolCalls
    │
    └── UseDeepAgentStreamOptions / UseDeepAgentStream
            └── Extends UseAgentStream
            Input:  + subagentToolNames, filterSubagentMessages
            Output: + subagents, activeSubagents, getSubagent, getSubagentsByType
```

Usage automatically resolves to the correct interface:

```typescript
useStream<typeof graph>()        // → BaseStream
useStream<typeof agent>()        // → UseAgentStream (with toolCalls)
useStream<typeof deepAgent>()    // → UseDeepAgentStream (with subagents)
```

## Tool Call Tracking (for `createAgent`)

Track tool calls with full type inference from your agent's tools:

```typescript
  assistantId: "my-agent",
});

// All tool calls paired with results
stream.toolCalls.forEach(({ call, result }) => {
  if (call.name === "search") {
    // call.args is typed as { query: string }
    console.log("Searched for:", call.args.query);
  }
});

// Get tool calls for a specific AI message
stream.getToolCalls(aiMessage)
```

## Subagent Streaming (for `createDeepAgent`)

Track multiple concurrent subagent executions:

```typescript
  assistantId: "vacation-planner",
  subagentToolNames: ["task"],       // Tool names that spawn subagents
  filterSubagentMessages: true,       // Keep main messages clean
});

// All subagent executions (Map<string, SubagentStream>)
stream.subagents

// Currently running subagents
stream.activeSubagents

// Get specific subagent by tool call ID
stream.getSubagent("call_abc123")

// Get all subagents of a specific type (with typed state)
stream.getSubagentsByType("weather-scout")
```

### `SubagentStream` interface:

```typescript
interface SubagentStream<StateType, ToolCall> {
  id: string;                    // Tool call ID
  toolCall: {
    id: string;
    name: string;
    args: {
      description?: string;
      subagent_type?: string;
    };
  };
  status: "pending" | "running" | "complete" | "error";
  result: string | null;         // Final result when complete
  error: string | null;          // Error message if failed
  namespace: string[];           // LangGraph namespace path
  messages: Message<ToolCall>[]; // Subagent's message history
  values: StateType;             // Subagent's state (typed per-subagent)
  parentId: string | null;       // For nested subagents
  depth: number;                 // Nesting level (0 = direct child)
  startedAt: Date | null;
  completedAt: Date | null;
}
```

### `filterSubagentMessages` option:

When enabled, subagent messages are filtered from the main `stream.messages` array and only accessible via `SubagentStream.messages`. This allows clean separation of main agent and subagent streams in the UI.

## Usage Example

```tsx
function DeepAgentUI() {
    assistantId: "vacation-planner",
    filterSubagentMessages: true,
  });

  return (
    <div>
      {/* Main agent messages */}
      <MessageList messages={stream.messages} />
      
      {/* Subagent cards */}
      <div className="grid grid-cols-3 gap-4">
        {[...stream.subagents.values()].map((subagent) => (
          <SubagentCard 
            key={subagent.id}
            type={subagent.toolCall.args.subagent_type}
            status={subagent.status}
            content={subagent.messages
              .filter(m => m.type === "ai")
              .map(m => m.content)
              .join("")}
          />
        ))}
      </div>
    </div>
  );
}
```

### New Managers
- **`SubagentManager`** class handles all subagent state tracking

### Key Features
- **Namespace mapping**: Connects LangGraph's internal pregel task IDs to the original tool call IDs via description matching
- **Per-node/subagent `MessageTupleManager`**: Properly concatenates streaming message chunks
- **Status-based filtering**: Only shows subagents once they're actually running (filters out partial streaming artifacts)
- **Type inference**: Full TypeScript inference for node return types, tool call args, and subagent states

## Added Examples

Added deepagent examples displaying individual subagent streams:

Screenshot 2026-01-15 at 5 00 13 PM

### DIFF
--- a/examples/ui-react/src/examples/deepagent/index.tsx
+++ b/examples/ui-react/src/examples/deepagent/index.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useMemo, useState, useEffect } from "react";
+import { useCallback, useMemo, useState } from "react";
 import { useStickToBottom } from "use-stick-to-bottom";
 import { AlertCircle, Plane, Sparkles } from "lucide-react";
 import { useStream } from "@langchain/langgraph-sdk/react";
@@ -62,7 +62,6 @@ function useThreadIdParam() {
 
 export function DeepAgentDemo() {
   const { scrollRef, contentRef } = useStickToBottom();
-  const [pipelineShown, setPipelineShown] = useState(false);
   const [threadId, onThreadId] = useThreadIdParam();
 
   // Use filterSubagentMessages to keep main messages clean
@@ -79,20 +78,6 @@ export function DeepAgentDemo() {
 
   const hasMessages = stream.messages.length > 0;
   const hasSubagents = stream.subagents.size > 0;
-
-  // Reset pipeline shown state when messages are cleared
-  useEffect(() => {
-    if (!hasMessages && !hasSubagents) {
-      setPipelineShown(false);
-    }
-  }, [hasMessages, hasSubagents]);
-
-  // Mark pipeline as shown when we have subagents
-  useEffect(() => {
-    if (hasSubagents && !pipelineShown) {
-      setPipelineShown(true);
-    }
-  }, [hasSubagents, pipelineShown]);
 
   // Check if we're in the synthesis phase (subagents done, waiting for final response)
   const allSubagentsDone =
@@ -126,7 +111,6 @@ export function DeepAgentDemo() {
 
   const handleSubmit = useCallback(
     (content: string) => {
-      setPipelineShown(false);
       stream.submit(
         { messages: [{ content, type: "human" }] },
         {
@@ -140,15 +124,31 @@ export function DeepAgentDemo() {
     [stream]
   );
 
-  // Determine where to show pipeline (after the last human message)
-  const humanMessageIndex = useMemo(() => {
-    for (let i = displayMessages.length - 1; i >= 0; i--) {
-      if (displayMessages[i].type === "human") {
-        return i;
+  /**
+   * Build a map of human message ID -> subagents for that turn.
+   * A "turn" is everything between two consecutive human messages.
+   */
+  const subagentsByHumanMessage = useMemo(() => {
+    const result = new Map<
+      string,
+      ReturnType<typeof stream.getSubagentsByMessage>
+    >();
+    const msgs = stream.messages;
+
+    for (let i = 0; i < msgs.length; i++) {
+      if (msgs[i].type !== "human") continue;
+
+      // The next message in the turn is the AI message with tool_calls
+      const next = msgs[i + 1];
+      if (!next || next.type !== "ai" || !next.id) continue;
+
+      const subagents = stream.getSubagentsByMessage(next.id);
+      if (subagents.length > 0) {
+        result.set(msgs[i].id!, subagents);
       }
     }
-    return -1;
-  }, [displayMessages]);
+    return result;
+  }, [stream.messages, stream.subagents]);
 
   return (
     <div className="h-full flex flex-col">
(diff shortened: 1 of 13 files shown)


examples/ui-react/src/examples/custom-streaming/agent.ts
```diff
@@ -12,15 +12,22 @@ const model = new ChatOpenAI({
...
 /**
  * Data analysis tool - demonstrates custom streaming events
  * Emits progress updates during execution using typed custom events
  */
 const analyzeDataTool = tool(
-  async (
-    { dataSource, analysisType },
-    config: ToolRuntime
-  ): Promise<string> => {
+  async (input: { dataSource: string; analysisType: string }, config: ToolRuntime): Promise<string> => {
```
hntrl:
```suggestion
  async (input, config: ToolRuntime): Promise<string> => {
```

examples/ui-react/src/examples/custom-streaming/agent.ts
```diff
@@ -89,22 +96,23 @@ const analyzeDataTool = tool(
...
+    .enum(["read", "compress", "validate", "transform"])
+    .describe("The operation to perform"),
+});
+
 /**
  * File processing tool - demonstrates status updates
  */
 const processFileTool = tool(
-  async ({ filename, operation }, config: ToolRuntime) => {
+  async (input: { filename: string; operation: string }, config: ToolRuntime) => {
```
hntrl:
```suggestion
  async (input, config: ToolRuntime) => {
```

christian-bromann:
> I want to avoid overloading `useStream` to be doing a bunch of different things.

IMHO `useStream` is the primitive to help use stream content from graphs (regardless what nature, e.g. raw LangGraph nodes, createAgent, createDeepAgent). I am concerned that introducing a separate primitive for deepagents just confuses people. We have to be able to clearly line out the difference and when to use one primitive over the other. E.g. does `createAgentStream` work for `createAgent` and `createDeepAgent`? At this point the concepts are too identical that it makes sense to create a separate primitive. I rather throw an error if someone sets `filterSubagentMessages` or accesses `subagents` from `stream` but not using `deepagentjs`.

christian-bromann:
~Note to myself: make `SubagentExecution` look closer to a `useStream` result~

examples/ui-react/langgraph.json
```diff
@@ -3,6 +3,8 @@
   "graphs": {
     "agent": "./src/examples/branching-chat/agent.ts:agent",
     "custom-streaming": "./src/examples/custom-streaming/agent.ts:agent",
+    "deepagent": "./src/examples/deepagent/agent.ts:agent",
+    "deepagent-tools": "./src/examples/deepagent-tools/agent.ts:agent",
```
hntrl: feels like we should have these examples in the deepagents repo?

examples/ui-react/src/examples/deepagent/index.tsx
```diff
@@ -0,0 +1,288 @@
...
+                    hasSubagents && (
+                      <div className="mt-6">
+                        <SubagentPipeline
+                          subagents={stream.subagents}
+                          isLoading={stream.isLoading && !allSubagentsDone}
+                        />
+                      </div>
+                    )}
+                </div>
+              ))}
```
dqbd: Think that we don't want to keep subgraph list shifting with each human message? Ideally would just render those as special tool calls?

libs/sdk/src/ui/nodes.ts
```diff
@@ -0,0 +1,356 @@
...
+ * manager.recordUpdate("researcher", { research_notes: "..." });
+ *
+ * // Complete the node
+ * manager.completeNode("researcher");
+ *
+ * // Get all executions of a node
+ * const executions = manager.getNodeStreamsByName("researcher");
+ * ```
+ */
+export class NodeManager<
```
dqbd: Think that the node state disappears the moment the stream ends / the page is reloaded? Would rather keep this user-land instead?

dqbd (approved):

PR #1935, opened 2026-02-02
Adding Checkpointer and Store Implementation of Amazon Bedrock AgentCore Memory 

hasnainvirk:
Implements both BaseCheckpointSaver and BaseStore for AWS Bedrock AgentCore Memory:
    
    - Full checkpoint persistence with thread isolation via sessionId
    - Supports all standard operations: getTuple, list, put, putWrites, deleteThread
    - Handles checkpoint_ns filtering and metadata search
    - Automatic retry logic with exponential backoff for AWS API throttling
    - Rate limiting (16.7 req/sec) to stay within AgentCore Memory limits
    - Unique actor ID generation for test isolation
    - Proper Unicode handling with TextEncoder/TextDecoder
    
    - Hierarchical namespace organization for data isolation
    - Metadata filtering and pagination for search operations
    - Complex JSON value support with proper serialization
    - Vector similarity search capabilities via RetrieveMemoryRecordsCommand
    - Consistent rate limiting and error handling
    
    - thread_id maps directly to sessionId in AgentCore Memory
    - actor_id provides user-level isolation across sessions
    - checkpoint_ns stored in event payload for namespace filtering
    - Store namespaces map to sessionId/actorId for hierarchical organization
    
    - Proper TypeScript interfaces for all AWS SDK interactions
    - No explicit 'any' types - uses proper type definitions
    - Comprehensive error handling with typed AWS error interfaces


libs/checkpoint-aws-agentcore-memory/src/store.ts
```diff
@@ -0,0 +1,525 @@
...
+      } else {
+        throw new Error(`Unsupported operation: ${JSON.stringify(op)}`);
+      }
+    }
+
+    return results as OperationResults<Op>;
+  }
+
+  private async handlePut(op: PutOperation): Promise<void> {
+    if (op.value === null) {
```
sundargthb, line 308: The DeleteEvent API is already being used in saver.ts → deleteThread() via DeleteEventCommand. Since store items are stored as events, the same mechanism applies.

Right now store.delete() silently does nothing, which breaks the BaseStore contract. Any LangGraph workflow that calls store.delete(namespace, key) will think it succeeded but the item is still there.

you can ListEvents filtered by the key's metadata, find the matching event(s), then DeleteEvent to remove them.

libs/checkpoint-aws-agentcore-memory/src/store.ts
```diff
@@ -0,0 +1,525 @@
...
+      return items.slice(offset, offset + limit);
+    } catch (error) {
+      if ((error as AWSError).name === "ResourceNotFoundException") {
+        return [];
+      }
+      throw error;
+    }
+  }
+
+  private async handleListNamespaces(
```
sundargthb, line 569: This returns a hardcoded empty array, which means store.listNamespaces() is non-functional. This is a required part of the BaseStore interface and other providers (Postgres, Redis) implement it fully with support for prefix, suffix, maxDepth, limit, offset.

libs/checkpoint-aws-agentcore-memory/src/saver.ts
```diff
@@ -0,0 +1,878 @@
...
+    }
+
+    try {
+      const response = await this.client.send(
+        new ListEventsCommand({
+          memoryId: this.memoryId,
+          actorId: resolvedActorId,
+          sessionId,
+          includePayloads: true,
+          maxResults: 100,
```
sundargthb: For an active thread, it's easy to exceed 100 events — each putWrites call stores one event per channel, so 10 checkpoints with 10-channel writes = 100 events right at the boundary.

libs/checkpoint-aws-agentcore-memory/src/store.ts
```diff
@@ -0,0 +1,525 @@
...
+  region?: string;
+}
+
+/**
+ * AWS Bedrock AgentCore Memory implementation of BaseStore.
+ *
+ * This store uses AgentCore Memory for persistent key-value storage with
+ * optional vector similarity search capabilities.
+ */
+export class AgentCoreMemoryStore extends BaseStore {
```
sundargthb, line 65: Other LangGraphJS store providers (PostgresStore, RedisStore) implement setup() / start() / stop() lifecycle methods. setup() typically validates the backing resource exists and is ready.

This store has no initialization step, so if someone passes an invalid memoryId or the Memory resource is still in CREATING status, they won't find out until the first get/put/search call fails with a cryptic error.

[GetMemory API](https://docs.aws.amazon.com/bedrock-agentcore/latest/APIReference/API_GetMemory.html) returns the resource's status field (CREATING | ACTIVE | FAILED | DELETING). A setup() method could call this to validate the memory is ACTIVE before any operations.

libs/checkpoint-aws-agentcore-memory/src/saver.ts
sundargthb, line 1: can we add input validation for getSessionId() / getActorId().. 

Memory API enforces strict patterns on these fields:

sessionId: pattern [a-zA-Z0-9][a-zA-Z0-9-_]*, max 100 chars ([CreateEvent docs](https://docs.aws.amazon.com/bedrock-agentcore/latest/APIReference/API_agent-memory_CreateEvent.html))
actorId: pattern [a-zA-Z0-9][a-zA-Z0-9-_/]*(?::[a-zA-Z0-9-_/]+)*[a-zA-Z0-9-_/]*, max 255 chars

libs/checkpoint-aws-agentcore-memory/src/saver.ts
```diff
@@ -0,0 +1,878 @@
...
+          })
+        );
+      } catch (error) {
+        console.error(`Error storing write for channel ${channel}:`, error);
+        throw error;
+      }
+    }
+  }
+
+  async deleteThread(threadId: string, actorId?: string): Promise<void> {
```
sundargthb, line 980: Every other method in this class calls this.rateLimit() before API calls, but deleteThread doesn't

libs/checkpoint-aws-agentcore-memory/src/store.ts
```diff
@@ -0,0 +1,525 @@
...
+      });
+    } catch (error) {
+      if ((error as AWSError).name === "ResourceNotFoundException") {
+        return [];
+      }
+      throw error;
+    }
+  }
+
+  private async performMetadataSearch(
```
sundargthb, line 473: Same pagination issue as getTuple — this calls ListEventsCommand with maxResults but doesn't handle nextToken. If a namespace has more than op.limit || 100 store events, the search silently returns incomplete results. ListEvents API supports pagination via nextToken and the docs recommend using it.

libs/checkpoint-aws-agentcore-memory/src/saver.ts
sundargthb, line 1: checkpoint_ns is being stuffed into the event payload blob and filtered client-side, but our Memory has native branch support for this purpose. CreateEvent accepts a branch: { name, rootEventId } field and ListEvents supports filter.branch.name for server-side filtering. Using branches would avoid downloading ALL events across all namespaces just to filter to one.

sundargthb, line 1: getTuple() / querySpecificSession() - API supports eventMetadata filters with EQUALS_TO operator — you could filter by type=checkpoint when you only need checkpoints

libs/checkpoint-aws-agentcore-memory/src/store.ts
```diff
@@ -0,0 +1,525 @@
...
+    if (op.query) {
+      // Use vector search
+      return this.performVectorSearch(op);
+    } else {
+      // Use metadata filtering
+      return this.performMetadataSearch(op);
+    }
+  }
+
+  private async performVectorSearch(
```
sundargthb: store.put() writes items as Events (via CreateEventCommand), but store.search() with a query string reads from Memory Records (via RetrieveMemoryRecordsCommand).

sundargthb:
[StartMemoryExtractionJob](https://docs.aws.amazon.com/bedrock-agentcore/latest/APIReference/API_StartMemoryExtractionJob.html) processes raw events and produces structured, searchable Memory Records. The implementation can expose this capability. Even if it's out of scope for v1, it's worth noting in the README that extraction jobs need to be triggered separately for vector search to return results

libs/checkpoint-aws-agentcore-memory/src/store.ts
hasnainvirk: The change is added in  2ad6bfe

libs/checkpoint-aws-agentcore-memory/src/store.ts
hasnainvirk: This change is added in 52d30058c777616f524a101348f7bda5ff635aea

libs/checkpoint-aws-agentcore-memory/src/saver.ts
hasnainvirk: Pagination loop is added here b2288f56d81b252db58e28d6ee26e585bff1b32d

libs/checkpoint-aws-agentcore-memory/src/store.ts
hasnainvirk: GetMemory is control plane call. But I understand what you want to achieve here. The closest thing to catch an issue early enough is to use ListActors() API as it needs only _memory ID_ and serves as a smoke test of sort. 
In b1dcfccf196f12329b93f2c27be4f45f861c501c, I had introduced _start()_ and _stop()_ lifecycle events. stop() is a noop and is documented why. _start()_ invoked _ListSessions_ but it requires _actorId_ as well. So I later changed it in f4aab35c86b4604aa9dcf25279e84c83db0d12cc to _ListActors()_.

hasnainvirk: I am leaving it unresolved for now, because I am not sure if I have captured the spirit of your review here properly or not.

libs/checkpoint-aws-agentcore-memory/src/saver.ts
hasnainvirk: Input validation is added here b6a02860a3ac9ca444940e07abbef4ef103723fd

libs/checkpoint-aws-agentcore-memory/src/store.ts
hasnainvirk: Handled in 6beba21d6b91405d75411f7581b1e11ddf0660e9

libs/checkpoint-aws-agentcore-memory/src/saver.ts
hasnainvirk: This was more complex than I had expected originally. A lot of learning for me. I will write a blog about it later. Handled here f5a3643aca2fea810b0a4a61e3bd41a7faa6f3e8

hasnainvirk: The eventMetadata filter optimization is not applicable to querySpecificSession because list() needs both checkpoints and writes in the same scan to build pendingWrites. The only place it could work is getTuple — but as documented in f4aab35c86b4604aa9dcf25279e84c83db0d12cc, that would require two API calls instead of one. So the conclusion is: the eventMetadata type filter is not beneficial in either method given the current data model.

libs/checkpoint-aws-agentcore-memory/src/saver.ts
hasnainvirk: Handled in cc38a54fd5107442c76ba3c8c11046af4f2d7bda

libs/checkpoint-aws-agentcore-memory/src/store.ts
hasnainvirk: That was a very good catch. My concept about LTM was weak, _RetrieveMemoryRecordsCommand_ would always return empty in this context. I have removed _performVectorSearch_ in f845a114b3c7e05f577635a654cd3653f90d53fe. Now we support only metadata based search on Events.

hasnainvirk:
> [StartMemoryExtractionJob](https://docs.aws.amazon.com/bedrock-agentcore/latest/APIReference/API_StartMemoryExtractionJob.html) processes raw events and produces structured, searchable Memory Records. The implementation doesn't expose this capability at all. Even if it's out of scope for v1, it's worth noting in the README that extraction jobs need to be triggered separately for vector search to return results

I am not too sure about that. This API was introduced after I had created the PR. I am convinced that it is not required to start any LTM activities as they start when an LTM strategy is defined. I am unclear about the utility of this API.

libs/checkpoint-aws-agentcore-memory/src/store.ts
sundargthb: we are good.. this can be resolved

gremmer:
@sundargthb @hasnainvirk Love the initiative! Can someone help me understand what's required to move this forward?

hasnainvirk:
@gremmer from review perspective, it's ready. We need CI and guidance on merging/releasing.

gremmer:
Hey @sundargthb and @hasnainvirk ! Just wanted to bump this gently. It would be such a great addition to the LangGraph ecosystem! @sundargthb is there anything blocking this from moving forward? Happy to help with rebasing, testing or anything else needed to get this moving again

christian-bromann:
Thank you for the contribution. I don't think at this point we can add another checkpointer implementation to the langgraphjs repository due to maintainability reasons. I suggest to host and publish it yourself and raise a PR here to add references to it.

Thank you!

PR #1954, opened 2026-02-10
chore: support api_version in CLI

hinthornw:
Support version pinning.


christian-bromann:
Can you share a little bit about the motivation for this change?

Issue #1812, opened 2025-12-06
checkpoint_blob versions not calculated globally


carlosbaraza:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangGraph.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangGraph.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangGraph (or the specific integration package).

### Example Code

```
// Use DO UPDATE instead of DO NOTHING for blobs
// This ensures branches can overwrite blobs at the same channel version
const UPSERT_BLOBS_WITH_UPDATE_SQL = `
  INSERT INTO public.checkpoint_blobs (thread_id, checkpoint_ns, channel, version, type, blob)
  VALUES ($1, $2, $3, $4, $5, $6)
  ON CONFLICT (thread_id, checkpoint_ns, channel, version) DO UPDATE SET
    type = EXCLUDED.type,
    blob = EXCLUDED.blob
`;
```

### Description

I am building a chatbot and I want to be able to rerun messages from a particular time in the checkpoints. However, when new messages are created, they are not stored correctly in Postgres, because the same version number is calculated for the rerun of the message, and blobs are not silently ignored by the DO NOTHING.

A preferable solution would be to change the version number calculation for the blobs not to be based on the previous state version, but something global.

Not sure if I might be doing something wrong here.

 ## Why Branches Have The Same Version

  When you rerun (branch from a checkpoint), LangGraph:

  1. Loads the parent checkpoint's state (e.g., messages at version 2 after first user
  message)
  2. Runs the AI again, which modifies messages
  3. Increments version: 2 → 3... BUT wait, version 3 already exists from the original
  branch!

  The problem is LangGraph calculates the next version based on the current state being 
  modified, not globally. So:

```
  Original:  User "Hello" (v1) → AI response (v2) → User "Hi" (v3) → AI response (v4)
                                      ↓
  Rerun from v2:                     AI new response → also gets v3!
                                      ↓
  Rerun again from v2:               AI another response → also gets v3!
```

  All reruns from the same point calculate the same next version.

### System Info

Node version: v22.12.0
Operating system: darwin arm64
Package manager: npm
Package manager version: N/A
--------------------
zod -> @3.25.55, , @"^3.25.55", @"^3.0.0", @"^3.23.8", @"^3.22.4", @"^3.24.1"

[...]

PR #1968, opened 2026-02-17
feat(checkpoint-postgres)!: align version format and inline primitive storage with Python

WilliamPeralta:
## Summary

**BREAKING:** Align checkpoint version format and channel_values storage with the Python implementation.

### Breaking changes:

1. **Version format:** `getNextVersion` now produces zero-padded string versions (e.g. `"00000000000000000000000000000001.0482910384729105"`) instead of integer versions (`1`, `2`, `3`).

   **Migration impact:** Existing checkpoints with integer versions will still be readable. New checkpoints will use string versions. Active threads may need to be completed before upgrading, as `getNextVersion` needs to parse the current version format.

2. **Inline primitives:** Primitive channel values (`string`, `number`, `boolean`, `null`) are now stored inline in the checkpoint JSONB column instead of in `checkpoint_blobs`. Requires #1967 (read-side merge) to be deployed first, so existing readers can handle both formats.

### Why:
These changes enable cross-compatibility between Python and JS checkpoint implementations sharing the same database, which is required for hybrid Python/JS LangGraph deployments.

### Upgrade path:
1. Deploy #1967 (non-breaking, adds read-side support) to all readers
2. Then deploy this PR to writers
3. Old integer-versioned checkpoints remain readable

### Test plan
- [ ] Verify `getNextVersion` produces correct zero-padded string format
- [ ] Verify primitive values are stored inline in checkpoint JSONB
- [ ] Verify complex values are still stored in checkpoint_blobs
- [ ] Verify read-side correctly merges inline + blob values
- [ ] Verify existing integer-versioned checkpoints still readable

### DIFF
--- a/.changeset/align-version-format-python.md
+++ b/.changeset/align-version-format-python.md
@@ -1,0 +1,13 @@
+---
+"@langchain/langgraph-checkpoint-postgres": minor
+---
+
+feat(checkpoint-postgres)!: align version format and inline primitive storage with Python
+
+Align checkpoint version format and channel_values storage with the Python implementation:
+
+1. **Version format:** `getNextVersion` now produces zero-padded string versions (e.g. `"00000000000000000000000000000001.0482910384729105"`) instead of integer versions (`1`, `2`, `3`).
+
+2. **Inline primitives:** Primitive channel values (`string`, `number`, `boolean`, `null`) are now stored inline in the checkpoint JSONB column instead of in `checkpoint_blobs`.
+
+These changes enable cross-compatibility between Python and JS checkpoint implementations sharing the same database, required for hybrid Python/JS LangGraph deployments.
--- a/libs/checkpoint-postgres/src/index.ts
+++ b/libs/checkpoint-postgres/src/index.ts
@@ -81,7 +81,7 @@ const { Pool } = pg;
  * }, config);
  * ```
  */
-export class PostgresSaver extends BaseCheckpointSaver {
+export class PostgresSaver extends BaseCheckpointSaver<string | number> {
   private readonly pool: pg.Pool;
 
   private readonly options: PostgresSaverOptions;
@@ -175,13 +175,39 @@ export class PostgresSaver extends BaseCheckpointSaver {
     }
   }
 
+  /**
+   * Generate the next version ID for a channel.
+   *
+   * Uses zero-padded string versions with random suffix for compatibility
+   * with the Python checkpoint-postgres implementation.
+   */
+  getNextVersion(
+    current: string | number | undefined
+  ): string | number {
+    if (current === undefined) {
+      return `${"0".repeat(31)}1.${Math.random().toString().slice(2, 18).padEnd(16, "0")}`;
+    }
+    const currentStr = String(current);
+    const currentV = parseInt(currentStr.split(".")[0], 10) || 0;
+    const nextV = currentV + 1;
+    const nextH = Math.random().toString().slice(2, 18).padEnd(16, "0");
+    return `${String(nextV).padStart(32, "0")}.${nextH}`;
+  }
+
   protected async _loadCheckpoint(
-    checkpoint: Omit<Checkpoint, "pending_sends" | "channel_values">,
+    checkpoint: Omit<Checkpoint, "pending_sends" | "channel_values"> & {
+      channel_values?: Record<string, unknown>;
+    },
     channelValues: [Uint8Array, Uint8Array, Uint8Array][]
   ): Promise<Checkpoint> {
     return {
       ...checkpoint,
-      channel_values: await this._loadBlobs(channelValues),
+      channel_values: {
+        // Merge inline primitives from checkpoint JSONB
+        ...(checkpoint.channel_values || {}),
+        // Blob values override inline primitives
+        ...(await this._loadBlobs(channelValues)),
+      },
     };
   }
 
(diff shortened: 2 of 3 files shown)


WilliamPeralta:
This PR directly addresses [#1812](https://github.com/langchain-ai/langgraphjs/issues/1812) — the branching bug caused by deterministic integer versions in `getNextVersion()`.

**Root cause:** When branching from the same checkpoint, `getNextVersion()` produces the same next integer version for both branches (e.g., both get version `3`). Since `checkpoint_blobs` uses `ON CONFLICT ... DO NOTHING`, the second branch's blob data is silently discarded, corrupting state. `MemorySaver` is unaffected because it doesn't use version-keyed blob storage.

**Fix:** `getNextVersion()` now produces globally unique string versions (`"00000000000000000000000000000003.7482910384729105"`) — zero-padded counter + random hash — ensuring no collisions across branches.

In the meantime, we've published a **temporary drop-in replacement** package with all fixes from both #1967 and this PR:

- **npm:** `@logicpanel/langgraph-checkpoint-postgres`
- **Source:** https://github.com/WilliamPeralta/langgraph-checkpoint-postgres

Usage is a one-line import change:

```typescript
// Before:
import { PostgresSaver } from "@langchain/langgraph-checkpoint-postgres";
// After:
import { PostgresSaver } from "@logicpanel/langgraph-checkpoint-postgres";
```

We'll deprecate the temporary package once these PRs are merged.

christian-bromann (commented):
Can we have some unit tests here to ensure we don't regress in the future?

WilliamPeralta:
@christian-bromann Christian BromannAdded unit tests in `src/tests/postgres-saver.test.ts` (21 tests) covering:

**`getNextVersion`** (8 tests)
- Produces correct format: 32-char zero-padded counter + dot + 16-char random hash
- Starts at counter 1 when `current` is `undefined`
- Increments counter correctly from string versions
- Backward compat: handles legacy integer versions (e.g. `3` → `"0...04.xxx"`)
- Produces unique versions across calls (random hash uniqueness)
- Monotonically increasing counters
- String sorting works correctly (critical for Postgres `ORDER BY`)
- Format matches Python's `f"{next_v:032}.{next_h:016}"`

**`_dumpCheckpoint`** (4 tests)
- Inlines primitive `channel_values` (string, number, boolean, null) in checkpoint JSONB
- Excludes complex values (arrays, objects) from inline storage
- Handles empty `channel_values`
- Preserves all other checkpoint fields

**`_dumpBlobs`** (5 tests)
- Skips primitive values (they go inline in checkpoint)
- Includes complex values in blob output
- Correctly separates primitives vs complex in mixed scenarios
- Returns empty array when no versions
- Handles channels with version but missing from values (empty type)

**`_loadCheckpoint`** (4 tests)
- Merges inline primitives from checkpoint JSONB with blob values
- Blob values override inline primitives on key collision
- Works with empty blob values (inline-only)
- Works with no inline `channel_values` (backward compat with pre-inline checkpoints)

All passing: `pnpm test` ✓, `pnpm lint` ✓, `pnpm format:check` ✓

WilliamPeralta:
so?

libs/checkpoint-postgres/src/index.ts
```diff
@@ -235,27 +262,60 @@ export class PostgresSaver extends BaseCheckpointSaver {
...
-    );
+    const results: [
+      string,
+      string,
+      string,
+      string,
+      string,
+      Uint8Array | undefined
+    ][] = [];
+    for (const [k, ver] of Object.entries(versions)) {
```
christian-bromann, line 273: We should keep `Promise.all`. For checkpoints with many non-primitive channel values, this serializes all dumpsTyped calls that were previously concurrent.

.changeset/align-version-format-python.md
```diff
@@ -0,0 +1,13 @@
+---
+"@langchain/langgraph-checkpoint-postgres": minor
+---
+
+feat(checkpoint-postgres)!: align version format and inline primitive storage with Python
+
+Align checkpoint version format and channel_values storage with the Python implementation:
+
+1. **Version format:** `getNextVersion` now produces zero-padded string versions (e.g. `"00000000000000000000000000000001.0482910384729105"`) instead of integer versions (`1`, `2`, `3`).
```
christian-bromann, line 9: I don't think this can be released as `minor` change. The `getNextVersion` signature changes the version format stored in Postgres, and the inline primitive storage changes the shape of the checkpoint JSONB column. Both are breaking changes for users who upgrade without first deploying #1967. The PR description itself uses the ! breaking-change marker.

We currently don't plan major releases , so I can't merge this PR in the current form.

libs/checkpoint-postgres/src/index.ts
```diff
@@ -81,7 +81,7 @@ const { Pool } = pg;
  * }, config);
  * ```
  */
-export class PostgresSaver extends BaseCheckpointSaver {
+export class PostgresSaver extends BaseCheckpointSaver<string | number> {
```
christian-bromann, line 84: Since getNextVersion always returns a string, the type parameter should be `BaseCheckpointSaver<string>`, not `BaseCheckpointSaver<string | number>`. The union is needed to accept legacy integer values as input, but the output is always a string. The union type leaks into callers and weakens type safety. We should handle the integer input case internally and keep the declared version type as string.

libs/checkpoint-postgres/src/index.ts
```diff
@@ -175,13 +175,40 @@ export class PostgresSaver extends BaseCheckpointSaver {
...
+   */
+  getNextVersion(current: string | number | undefined): string | number {
+    if (current === undefined) {
+      return `${"0".repeat(31)}1.${Math.random()
+        .toString()
+        .slice(2, 18)
+        .padEnd(16, "0")}`;
+    }
+    const currentStr = String(current);
+    const currentV = parseInt(currentStr.split(".")[0], 10) || 0;
```
christian-bromann, line 192: The `|| 0` fallback silently swallows a `NaN` and a legitimate `0`. If `parseInt` returns `NaN` (e.g. a malformed version), this produces `nextV = 1`, which looks like a fresh start and could silently cause a version regression. Use `Number.isNaN(parsed) ? 0 : parsed` instead:

```ts
const parsed = parseInt(currentStr.split(".")[0], 10);
const currentV = Number.isNaN(parsed) ? 0 : parsed;
```

libs/checkpoint-postgres/src/tests/postgres-saver.test.ts
```diff
@@ -0,0 +1,401 @@
...
+      versions_seen: {},
+    };
+
+    const encoder = new TextEncoder();
+    // Simulate a blob for a complex value — use JSON serde format
+    const channelValues: [Uint8Array, Uint8Array, Uint8Array][] = [
+      [
+        encoder.encode("blobKey"),
+        encoder.encode("json"),
+        encoder.encode(JSON.stringify({ complex: true })),
```
christian-bromann, line 322: This test bypasses the real serde (the default JsonPlusSerializer) and manually encodes values as json type. The _loadBlobs method calls `this.serde.loadsTyped("json", bytes)`, if the default serde doesn't handle plain "json" type strings (it uses "msgpack" or "json+" under the hood), this test could pass while the real runtime path fails. Consider using the actual serde round-trip (await saver.serde.dumpsTyped(...)) to produce the blob values in the test, to ensure you're exercising the real deserialization path.

christian-bromann (changes requested):
I don't think the PR is mergable at current state due to its breaking nature.

PR #1971, opened 2026-02-17
fix(sdk): select branch with newest checkpoint as default in getBranchView

WilliamPeralta:
## Summary

- Fix `getBranchView` to select the fork branch with the **newest checkpoint** when the requested branch path can't be matched (e.g. `branch=""` after page refresh)
- Previously, `at(-1)` was used as fallback which picks the **oldest** fork due to `unshift` ordering in `getBranchSequence`
- Add `findLatestForkIndex` helper that compares checkpoint IDs (UUID v1 strings are time-sortable) across all fork branches to find the most recently active one

## Bug

When `useStream` initializes with `branch=""`:
1. `forkStack = [""]` → `forkId = ""` (not null, so `findIndex` runs)
2. `findIndex(checkpoint_id === "")` → `-1` (never matches)
3. `item.items.at(-1)` → picks last in array = **oldest** fork (because `unshift` inserts newest first at line 134)

This causes the UI to show an old/stale branch instead of the most recent one after a page refresh or stream end.

## Userland workaround

Before this fix, consumers had to work around the issue by traversing `experimental_branchTree` themselves to find the correct branch:

```tsx
function StreamProvider({ threadId, children }) {
  const stream = useStream({ threadId, /* ... */ });
  const savedBranch = useRef("");

  useLayoutEffect(() => {
    // branch resets to "" after submit() and on page refresh
    if (stream.branch) {
      savedBranch.current = stream.branch;
      return;
    }
    if (stream.isLoading) return;

    // Traverse experimental_branchTree to find the node with the
    // highest checkpoint_id, then call setBranch() with its path
    if (savedBranch.current) {
      stream.setBranch(savedBranch.current);
    } else {
      const latest = findLatestBranchInTree(stream.experimental_branchTree);
      if (latest) stream.setBranch(latest);
    }
  });

  return <StreamContext.Provider value={stream}>{children}</StreamContext.Provider>;
}
```

This workaround is fragile and should not be necessary — the SDK should pick the correct default branch internally.

## Fix

When `findIndex` returns `-1`, instead of blindly using `at(-1)`, call `findLatestForkIndex()` which traverses all nodes in each fork branch and picks the one containing the highest (most recent) checkpoint ID.

This handles:
- **Fresh page load with `branch=""`**: picks the fork with the newest activity
- **Old branch that was continued**: correctly picks the continued branch even if it wasn't the last-created fork

## Test plan

- [x] `getBranchView` with `branch=""` selects the fork with the newest checkpoint
- [x] `getBranchView` with `branch=""` picks a continued older branch over a newer-created fork
- [x] `getBranchView` with explicit branch path still works correctly (regression)
- [x] All existing tests pass
- [x] Lint and format checks pass


christian-bromann:
Thanks @WilliamPeralta for the contribution.

We no longer plan to maintain the legacy UI primitives in `@langchain/langgraph-sdk` and recommend folks to move to the `@langchain/react` package. I am planning on working on a dedicated hook to simplify branching. You can find some preliminary work in our [streaming cookbook](https://github.com/langchain-ai/streaming-cookbook/pull/11/changes#diff-a270efb7ed3bd67d677bb0b701c76cc5dcf69620026a7ee984f1b51c1c354752). Let me know if you have any questions.

WilliamPeralta:
> Thanks WilliamPeralta for the contribution.
> 
> We no longer plan to maintain the legacy UI primitives in `@langchain/langgraph-sdk` and recommend folks to move to the `@langchain/react` package. I am planning on working on a dedicated hook to simplify branching. You can find some preliminary work in our [streaming cookbook](https://github.com/langchain-ai/streaming-cookbook/pull/11/changes#diff-a270efb7ed3bd67d677bb0b701c76cc5dcf69620026a7ee984f1b51c1c354752). Let me know if you have any questions.

That's a shame. I was expecting the opposite: better support for the primitives rather than the specific UI layer. I actually moved to the primitives because I find the React library too opinionated, somewhat buggy, and not very optimized.

PR #1984, opened 2026-02-19
feat: add tools stream mode for tool lifecycle events

colifran:
### Description
Adds a tools stream mode so LangGraph can stream tool lifecycle events (start, partial, end, error) to the client when running in tools stream mode.

#### Stream mode and events
- Stream mode: "tools" is added to the StreamMode union in core, API schemas, and storage types. Callers can pass streamMode: ["tools"] (or include "tools" with other modes) when starting a run.
- Events emitted: For each tool run, the stream emits:
  - on_tool_start – name, input, and optional toolCallId
  - on_tool_event – optional data (one chunk per yield when using async-generator tools from @langchain/core)
  - on_tool_end – name, output, and optional toolCallId
  - on_tool_error – name, error, and optional toolCallId

#### Implementation
- StreamToolsHandler (in libs/langgraph-core/src/pregel/stream.ts): Extends BaseCallbackHandler and implements handleToolStart, handleToolEvent, handleToolEnd, and handleToolError. It pushes StreamChunk tuples [namespace, "tools", payload] to a streamFn, with payload containing the event type and the fields above. It uses an optional toolCallId argument on handleToolStart (no longer from metadata) so it works with the explicit-param API in @langchain/core once that is released.
- Pregel integration (in libs/langgraph-core/src/pregel/index.ts): When streamMode includes "tools", a StreamToolsHandler is instantiated with a callback that pushes chunks to the run’s stream and is added to config.callbacks (merged with any existing callbacks).
- ToolNode (libs/langgraph-core/src/prebuilt/tool_node.ts): Invokes tools with tool.invoke({ ...call, type: "tool_call" }, config) so @langchain/core can pass toolCallId into handleToolStart when that support is available.
- Types: StreamToolsOutput and the stream payload map in libs/langgraph-core/src/pregel/types.ts are extended so tools-mode chunks are typed as [string[], "tools", { event, toolCallId?, name, input?, data?, output?, error? }].

#### SDK and API
- SDK (libs/sdk): ToolsStreamEvent, ToolProgress, and the stream-mode map include the tools events; the React stream layer handles on_tool_start / on_tool_event / on_tool_end / on_tool_error and updates toolProgress for the UI.
- API (libs/langgraph-api): Run/create schemas and storage types allow "tools" in streamMode; embed tests assert that tools stream mode emits at least on_tool_start and on_tool_end.

### Testing
- Unit: StreamToolsHandler is tested in isolation (namespace, toolCallId, partial/end/error, no emit when metadata is undefined or when tags include TAG_HIDDEN).
- Integration: Pregel test “streamMode tools emits on_tool_start and on_tool_end for ToolNode runs” checks that streaming with streamMode: ["updates", "tools"] yields tools chunks with the expected events; toolCallId is asserted when present (optional until @langchain/core passes it). A prebuilt test that asserts toolCallId is passed to handleToolStart is skipped until @langchain/core exposes that parameter (see langchainjs PR #10102).

### Dependencies
Works with current @langchain/core; toolCallId in events is optional. Full toolCallId support will align with langchainjs PR #10102 (explicit handleToolStart(..., toolCallId) and tools stream mode in core).

### DIFF
--- a/libs/langgraph-api/package.json
+++ b/libs/langgraph-api/package.json
@@ -96,7 +96,7 @@
     }
   },
   "devDependencies": {
-    "@langchain/core": "^1.1.27",
+    "@langchain/core": "^1.1.28",
     "@langchain/langgraph": "workspace:*",
     "@langchain/langgraph-checkpoint": "workspace:*",
     "@langchain/langgraph-sdk": "workspace:*",
--- a/libs/langgraph-core/package.json
+++ b/libs/langgraph-core/package.json
@@ -50,7 +50,7 @@
   },
   "devDependencies": {
     "@langchain/anthropic": "^1.0.0",
-    "@langchain/core": "^1.1.27",
+    "@langchain/core": "^1.1.28",
     "@langchain/langgraph-checkpoint-postgres": "workspace:*",
     "@langchain/langgraph-checkpoint-sqlite": "workspace:*",
     "@langchain/openai": "^1.0.0",
--- a/libs/sdk/package.json
+++ b/libs/sdk/package.json
@@ -28,7 +28,7 @@
     "uuid": "^13.0.0"
   },
   "devDependencies": {
-    "@langchain/core": "^1.1.27",
+    "@langchain/core": "^1.1.28",
     "@langchain/scripts": "^0.1.4",
     "deepagents": "^1.5.0",
     "langchain": "^1.2.11",
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
   },
   "pnpm": {
     "overrides": {
-      "@langchain/core": "^1.1.27",
+      "@langchain/core": "^1.1.28",
       "@langchain/langgraph-sdk": "workspace:*",
       "@langchain/langgraph-checkpoint": "workspace:*",
       "cheerio": "^1.0.0-rc.12",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -5,7 +5,7 @@ settings:
   excludeLinksFromLockfile: false
 
 overrides:
-  '@langchain/core': ^1.1.27
+  '@langchain/core': ^1.1.28
   '@langchain/langgraph-sdk': workspace:*
   '@langchain/langgraph-checkpoint': workspace:*
   cheerio: ^1.0.0-rc.12
@@ -73,7 +73,7 @@ importers:
         specifier: ^1.0.0
         version: 1.3.11(@langchain/core@1.1.28(openai@6.16.0(ws@8.19.0)(zod@4.3.5)))
       '@langchain/core':
-        specifier: ^1.1.27
+        specifier: ^1.1.28
         version: 1.1.28(openai@6.16.0(ws@8.19.0)(zod@4.3.5))
       '@langchain/langgraph':
         specifier: workspace:*
@@ -100,7 +100,7 @@ importers:
         specifier: ^1.0.0
         version: 1.3.11(@langchain/core@1.1.28(openai@6.16.0(ws@8.19.0)(zod@4.3.5)))
       '@langchain/core':
-        specifier: ^1.1.27
+        specifier: ^1.1.28
         version: 1.1.28(openai@6.16.0(ws@8.19.0)(zod@4.3.5))
       '@langchain/langgraph':
         specifier: workspace:*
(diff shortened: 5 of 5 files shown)


libs/langgraph-core/src/pregel/stream.ts
```diff
@@ -149,6 +160,107 @@ export class IterableReadableWritableStream extends IterableReadableStream<Strea
...
+      {
+        event: "on_tool_start",
+        toolCallId: info.toolCallId,
+        name: info.toolName,
+        input,
+      },
+    ]);
+  }
+
+  handleToolStream(chunk: unknown, runId: string) {
```
colifran: Need to update this should be handleToolEvent.

libs/sdk/src/types.stream.ts
```diff
@@ -246,6 +247,50 @@ export type FeedbackStreamEvent = {
...
+  data: ToolStreamEventData;
+};
+
+export type ToolProgress = {
+  toolCallId?: string;
+  name: string;
+  state: "starting" | "running" | "completed" | "error";
+  input?: unknown;
+  data?: unknown;
+  result?: unknown;
```
christian-bromann, line 287: IMHO this should be typed

colifran: ToolProgress is a runtime streaming container for events from arbitrary user-defined tools. The SDK has no knowledge of what tools exist or what they return/yield. One tool might return a string, another a { temperature: number; unit: string }, another an array. Typing result or data as anything more specific than unknown would either be wrong, e.g., Record<string, unknown> breaks for tools that return/yields strings or arrays) or force tool authors to conform to a shape the SDK dictates which defeats the purpose. This is the same reason our callback handlers type these as unknown or any. The narrowing happens at the consumer level where they know and own their own tool shapes.

I considered making ToolProgress generic, e.g., ToolProgress<TInput, TData, TResult>), but it doesn't work well in practice since agents can have multiple tools with different return types so a single generic can't represent per-tool result shapes.

libs/langgraph-core/src/pregel/types.ts
```diff
@@ -72,6 +73,32 @@ type StreamTasksOutput<StreamUpdates, StreamValues, Nodes = string> =
...
 
+export type StreamToolsOutput =
+  | {
+      event: "on_tool_start";
+      toolCallId?: string;
+      name: string;
+      input: unknown;
+    }
+  | {
+      event: "on_tool_partial";
```
hntrl: thoughts on this? since that's what we're calling it in callbacks
```suggestion
      event: "on_tool_event";
```

colifran: Yeah makes sense. Good catch.

hntrl (dismissed):

libs/sdk/src/types.stream.ts
christian-bromann: > The SDK has no knowledge of what tools exist or what they return/yield.

It actually does if you define a stream like this:

```ts
    assistantId: "tool-calling-agent",
  });
```
the stream object yields information about the structured tool, e.g.:

```ts
const stream: UseAgentStream<BaseAgentState<{
    name: "get_weather";
    args: {
        location: string;
    };
    id?: string;
    type?: "tool_call";
}>, {
    name: "get_weather";
    args: {
        location: string;
    };
    id?: string;
    type?: "tool_call";
}, BagTemplate>
```

This is because the ReactAgent or DeepAgent yield these type information:

```ts
ReactAgent<
    AgentTypeConfig<
        ResponseFormatUndefined,
        undefined,
        AnyAnnotationRoot,
        readonly AgentMiddleware<any, any, any,
        readonly (ClientTool | ServerTool)[]>[],
        readonly [DynamicStructuredTool<ZodObject<{
            location: ZodString;
        }, $strip>, {
            location: string;
        }, {
            location: string;
        }, string, "get_weather">, ServerTool]
    >
>
```

It should be technically possible to extend the `DynamicStructuredTool` structure to hold type information of the generator and pass it all the way into the stream primitives.

colifran: Ahh okay fair enough - thanks for the example!

hntrl (approved):

shlroland:
This modification seems not to have been synchronized to the Python version. I am currently using the langgraphjs SDK to interface with the langgraph-api of the Python version, and if the tools' stream_mode occurs, it will cause the server's data verification to fail.
@colifran

colifran:
Hey @shlroland. Yeah we don't have support for this on the Python side yet. Thanks for calling this out.

PR #1987, opened 2026-02-20
feat(sdk): add extract parameter to threads.search()

hinthornw:
## Summary
- Adds `extract` parameter to `threads.search()` in the JS/TS SDK client
- Adds `extracted` field to the `Thread` interface response type
- The `extract` parameter accepts a `Record<string, string>` mapping aliases to JSONB paths (e.g., `{"last_msg": "values.messages[-1]"}`)
- Depends on server-side support in https://github.com/langchain-ai/langgraph-api/pull/2609

## Test plan
- [ ] Verify types are correct via build (already passing)
- [ ] Integration test against server with extract feature enabled

Release Notes: Add `extract` parameter to `threads.search()` for extracting nested values from thread data during search.

### DIFF
--- a/.changeset/wicked-beers-hug.md
+++ b/.changeset/wicked-beers-hug.md
@@ -1,0 +1,5 @@
+---
+"@langchain/langgraph-sdk": patch
+---
+
+feat(sdk): add extract parameter to threads.search()


christian-bromann (dismissed):
LGTM 👍

PR #1989, opened 2026-02-20
Fix issue: Runtime.store always undefined #1952

Axadali:
Fix: runtime.store Always Undefined

    🐛 Issue
    When passing a store to createAgent() or compiling a graph with a store, accessing runtime.store from within node
    functions or middleware always returned undefined.

    Related Issue: #1952

    🔍 Root Cause
    While the store was properly passed to the graph and available internally, it wasn't being explicitly set on the
    config object that gets passed to node functions as the runtime parameter.

    Other runtime properties like writer and interrupt were explicitly set in the invoke method, but store was missing
     this assignment.

    ✅ Solution
    Added a single line in pregel/index.ts to ensure config.store is explicitly set on the runtime config:

     1 // Ensure store is explicitly set on config for runtime access
     2 config.store ??= store;

    This follows the existing pattern used for config.writer and config.interrupt.

    📝 Changes
     - libs/langgraph-core/src/pregel/index.ts: Added explicit store assignment (1 line + comment)
     - libs/langgraph-core/src/tests/pregel.test.ts: Added test case to verify runtime.store accessibility

    🧪 Testing
     - ✅ New test: runtime.store is accessible in nodes
     - ✅ All existing store-related tests pass
     - ✅ No regressions detected

    🎯 Scope
    This fix applies to all graph types:
     - StateGraph
     - MessageGraph
     - Prebuilt agents (createReactAgent, createAgent, etc.)
     - Custom graphs

    🔄 Backward Compatibility
     - ✅ No breaking changes
     - ✅ Uses ??= to respect user-provided store values
     - ✅ No type definition changes required
     - ✅ No documentation updates needed

    💡 Why This Approach?
    Rather than fixing individual agent creator functions (like createAgent), this fix addresses the root cause at the
     Pregel level where the runtime config is constructed. This ensures consistency across all graph types and
    prevents similar issues in the future.

### DIFF
--- a/.changeset/fix-runtime-store-undefined.md
+++ b/.changeset/fix-runtime-store-undefined.md
@@ -1,0 +1,5 @@
+---
+"@langchain/langgraph": patch
+---
+
+Fix `runtime.store` being undefined when accessing the store from within node functions or middleware. The store is now properly attached to the runtime config, making it accessible via `runtime.store` in all graph types (StateGraph, prebuilt agents, etc.).


Axadali:
Hi @hntrl, just following up on this PR.
Last week I opened it and haven’t seen any updates yet.
If it’s good to merge, we can proceed; otherwise, I’m happy to make any needed changes.

Thanks!

PR #1991, opened 2026-02-20
feat(checkpoint-mongodb): add opt-in `enableTimestamps` for `upserted_at` field

vanb:
## Summary
- Adds an optional `enableTimestamps` parameter to `MongoDBSaver`
- When enabled, writes an `upserted_at` BSON date to checkpoint and checkpoint_writes documents on every upsert
- Includes unit tests for the `timestampFields` getter

## Motivation
Long-running applications accumulate checkpoint data over time. MongoDB's built-in [TTL index](https://www.mongodb.com/docs/manual/core/index-ttl/) mechanism provides a reliable, zero-maintenance way to manage data retention — expired documents are automatically cleaned up by MongoDB's background thread without requiring application-side cron jobs or manual purges.

This approach writes an `upserted_at` timestamp on every upsert and delegates the retention policy entirely to MongoDB's `expireAfterSeconds` setting on the TTL index. This keeps the application code minimal and gives operators full control over retention at the database level. The timestamp is also useful for auditing and debugging independently of TTL.

Usage:
```ts
const saver = new MongoDBSaver({ client, enableTimestamps: true });
```

To enable TTL, create indexes on both collections:
```js
// Expire documents 7 days after last upsert
db.checkpoints.createIndex({ "upserted_at": 1 }, { expireAfterSeconds: 604800 })
db.checkpoint_writes.createIndex({ "upserted_at": 1 }, { expireAfterSeconds: 604800 })
```

## Test plan
- [x] `pnpm format:check` passes
- [x] `pnpm lint` passes
- [x] `pnpm build` succeeds
- [x] `pnpm test` passes (6/6 including 2 new timestamp tests)

### DIFF
--- a/.changeset/add-overwrite-class.md
+++ b/.changeset/add-overwrite-class.md
@@ -1,0 +1,11 @@
+---
+"@langchain/langgraph": minor
+---
+
+feat(langgraph): add Overwrite class for bypassing channel reducers
+
+Adds an `Overwrite` class and `OverwriteValue` type that allow nodes to bypass reducers in `BinaryOperatorAggregate` channels, writing values directly instead of passing them through the reducer function. This is useful when a node needs to replace accumulated state rather than append to it.
+
+- New `Overwrite` class exported from `@langchain/langgraph`
+- `BinaryOperatorAggregate` channel detects `OverwriteValue` and sets the value directly
+- `Annotation`, `StateSchema`, and zod schema type mappings updated to include `OverwriteValue` in update types
--- a/examples/ui-react/langgraph.json
+++ b/examples/ui-react/langgraph.json
@@ -10,7 +10,8 @@
     "reasoning-agent": "./src/examples/reasoning-agent/agent.ts:agent",
     "session-persistence": "./src/examples/session-persistence/agent.ts:agent",
     "summarization-agent": "./src/examples/summarization-agent/agent.ts:agent",
-    "tool-calling-agent": "./src/examples/tool-calling-agent/agent.ts:agent"
+    "tool-calling-agent": "./src/examples/tool-calling-agent/agent.ts:agent",
+    "tool-streaming": "./src/examples/tool-streaming/agent.ts:agent"
   },
   "env": ".env"
 }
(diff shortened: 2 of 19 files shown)


libs/checkpoint-mongodb/src/index.ts
```diff
@@ -16,6 +16,8 @@ export type MongoDBSaverParams = {
   dbName?: string;
   checkpointCollectionName?: string;
   checkpointWritesCollectionName?: string;
+  /** When true, writes an `upserted_at` BSON date to documents on every upsert. Useful for MongoDB TTL indexes, auditing, or debugging. */
```
hntrl: nit: (our jsdocs sometimes freaks out with formatting)
```suggestion
  /**
   * When true, writes an `upserted_at` BSON date to documents on every upsert.
   * Useful for MongoDB TTL indexes, auditing, or debugging.
   */
```

libs/checkpoint-mongodb/src/index.ts
```diff
@@ -30,12 +32,19 @@ export class MongoDBSaver extends BaseCheckpointSaver {
 
   checkpointWritesCollectionName = "checkpoint_writes";
 
+  protected enableTimestamps: boolean;
+
+  private get timestampFields(): { upserted_at: Date } | Record<string, never> {
```
hntrl: this return signature seems odd? can we just do `Record<string, unknown>`?

vanb:
Updated to use MongoDB's `$currentDate` operator instead of `new Date()` for the `upserted_at` field. The timestamp is now set server-side, avoiding clock skew across distributed app servers.

Changes in this push:
- Renamed `timestampFields` → `timestampOp` (returns `{ $currentDate: { upserted_at: true } }`)
- Moved the operator from inside `$set` to alongside it in both `put()` and `putWrites()`
- Simplified tests (exact match on operator object, no timing bounds needed)
- Updated JSDoc and changeset description

libs/checkpoint-mongodb/src/index.ts
vanb: yea, i think it literally means either `{ upserted_at: Date }` or `{}` so it couldn't accidentally return `{upserrrrted_at: Date}`. But agree it reads a little off; happy to make this more readable.

vanb: i couldn't use unknown so just using type inference for return type on this simple function.

> unknown won't work here because the return value gets spread into MongoDB update objects, and TypeScript won't let you spread Record<string, unknown> into a typed object. But we can just drop the explicit return type entirely and let TypeScript infer it.

hntrl:
Thanks for the PR!

PR #2004, opened 2026-02-26
feat: implement type inference for tool streaming

colifran:
### Summary
Adds type inference for toolProgress in useStream, so that useStream<typeof agent>() produces typed tool names, inputs, streaming data, and results based on the agent's tool definitions without manual casting or explicit type parameters.

#### Problem
When using useStream<typeof agent>(), the toolProgress array typed all fields as unknown and name as string. Users had no way to get typed narrowing on tool progress entries based on the agent's actual tool definitions. For streaming tools that use AsyncGenerator<YieldT, ReturnT>, the yield and return types were completely erased.

#### Solution
- New ToolProgress generic type with discriminated union on state ("starting" | "running" | "completed" | "error"), enabling TypeScript narrowing while keeping all fields optional for backward compatibility.
- InferToolMapFromAgent<T> type utility in ui/types.ts that extracts { [toolName]: { input, data, result } } from an agent's tools array, including AsyncGenerator yield/return types for streaming tools.
- DeriveToolProgress<T> and GetToolProgressType<Bag> helpers in types.stream.ts that convert an inferred tool map into a typed discriminated union of ToolProgress entries.
- Added optional ToolMap to BagTemplate in types.template.ts, wired through InferBag in ui/stream/index.ts. Uses Omit to avoid index signature pollution from the base template.
- Updated toolProgress property type in BaseStream and UseStream to use GetToolProgressType<Bag>[].
- Fixed handleToolEvent in react/stream.lgp.tsx to explicitly set undefined for fields that don't apply in each state transition, preventing stale values from leaking across states via object spread.

### Testing
New type-level tests (stream.test-d.ts) verify tool progress inference:
- Simple agent infers tool name and input; data/result remain unknown for non-streaming tools
- Multi-tool agent infers all tool entries with correct input types
- Streaming tool (AsyncGenerator) infers typed data and result
- useStream toolProgress has literal tool names (not generic string)
- State narrowing (state === "running") produces typed data and result

### Example
This example shows how these changes enable type inference for tool names and tool events:
type-inference

### DIFF
--- a/libs/sdk/src/types.template.ts
+++ b/libs/sdk/src/types.template.ts
@@ -12,4 +12,8 @@ export type BagTemplate = {
   CustomEventType?: unknown;
   UpdateType?: unknown;
   MetaType?: unknown;
+  ToolMap?: Record<
+    string,
+    { input?: unknown; data?: unknown; result?: unknown }
+  >;
 };
--- a/libs/sdk/src/react/stream.lgp.tsx
+++ b/libs/sdk/src/react/stream.lgp.tsx
@@ -319,7 +319,13 @@ export function useStreamLGP<
         }
         case "on_tool_event": {
           if (existing) {
-            next.set(key, { ...existing, state: "running", data: data.data });
+            next.set(key, {
+              ...existing,
+              state: "running",
+              data: data.data,
+              result: undefined,
+              error: undefined,
+            });
           }
           break;
         }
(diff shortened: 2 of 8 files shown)


libs/sdk/src/types.template.ts
```diff
@@ -12,4 +12,8 @@ export type BagTemplate = {
   CustomEventType?: unknown;
   UpdateType?: unknown;
   MetaType?: unknown;
+  ToolMap?: Record<
```
hntrl, line 15: nit: can we call this `ToolTypes`? also if we could update the jsdoc

christian-bromann (approved):
Thank you 👍

PR #2009, opened 2026-02-27
feat(prebuilt): add "system-prompt" mode to withAgentName

### LINKED ISSUE #1178: prebuilt: `includeAgentName` and `withAgentName` won't work if provider disallows editing AIMessage history
Supervisor and co rely on setting the `name` field of `AIMessage` to indicate which agent generated which `AIMessage` in the message history. Few providers actually observe this field, so the `includeAgentName` parameter was introduced (e.g. see #1174 and #1176) that modifies past AIMessages to follow the format `<name>${agentName}</name><content>${content}</content>`.

This is fine in cases when the chat model provider allows updates of previous AIMessages in context, but there are cases where this is disallowed (e.g. when referencing past messages using the OpenAI responses API, or when using thinking blocks with Anthropic).

There may be multiple ways to fix this (including provider-specific approaches), but the most reliable strategy that I can think of to get around this issue would be to get the model to generate its output in this format. This could involve modifying the system message to include prompting that tells the model to use this format using structured output when available.

RajeshKumar11:
## Summary

Fixes #1178

The existing `"inline"` mode in `withAgentName` / `includeAgentName` works by retroactively **mutating past `AIMessage` content** to inject XML tags:

```
<name>agentName</name><content>original response</content>
```

This breaks with providers that disallow editing message history:
- **OpenAI Responses API** — past message IDs are immutable
- **Anthropic thinking blocks** — thinking blocks in history cannot be modified/replayed

### Fix

Add a new `"system-prompt"` mode to `AgentNameMode` that:
1. **Does NOT mutate past messages** — leaves the entire conversation history untouched
2. **Injects an instruction into the system prompt** telling the model to format its own output in `<name>agentName</name><content>response</content>` format
3. **Reuses the existing `_removeInlineAgentName` output processor** — same XML extraction logic on the new AI message

### Changes

- `libs/langgraph-core/src/prebuilt/agentName.ts`
  - Extend `AgentNameMode = "inline" | "system-prompt"`
  - Add `_addSystemPromptAgentName(messages, agentName)` — appends instruction to existing `SystemMessage`, or prepends a new one if absent
  - Update `withAgentName(model, mode, agentName?)` to handle the new mode (optional `agentName` param, required for `"system-prompt"`)

- `libs/langgraph-core/src/prebuilt/react_agent_executor.ts`
  - Update `includeAgentName` type from `"inline" | undefined` → `AgentNameMode`
  - Pass `name` through to `withAgentName` at both call sites

- `libs/langgraph-supervisor/src/supervisor.ts`
  - Pass `supervisorName` to `withAgentName` (needed for `"system-prompt"` mode)

- Tests: 4 new unit tests for `_addSystemPromptAgentName` + 3 new parametrized supervisor test cases for `"system-prompt"` mode

### Backward Compatibility

- `"inline"` mode is **completely unchanged** — existing behaviour preserved
- `agentName` param on `withAgentName` is **optional** — no breaking change to the signature
- `includeAgentName` type widens from `"inline" | undefined` → `AgentNameMode` — fully compatible

### Test Results

```
@langchain/langgraph:  1117 passed (0 failed)
@langchain/langgraph-supervisor: 10 passed (0 failed) — includes 3 new system-prompt cases
@langchain/langgraph-swarm: 1 passed
```


christian-bromann:
@RajeshKumar11 mind sharing why you are using the (outdated) prebuilt primitives rather than [`createAgent`](https://docs.langchain.com/oss/javascript/langchain/quickstart) or [`createDeepAgent`](https://docs.langchain.com/oss/javascript/deepagents/quickstart)?

christian-bromann:
Closing due to inactivity.

PR #2001, opened 2026-02-25
feat(sdk): add vue, svelte, angular implementation of useStream hook

christian-bromann:
This PR introduces first-class **Vue**, **Svelte**, and **Angular** integrations for the LangGraph `useStream` hook — bringing the same streaming chat experience previously available only in React to three additional frontend frameworks. It also restructures the SDK internals to extract a framework-agnostic UI core, splits the React integration into its own dedicated package, and adds a client-side submission queue system.

### New Packages

| Package | Framework | Reactivity Model | Peer Dependencies |
|---|---|---|---|
| `@langchain/react` | React 18/19 | Hooks (`useState`, `useSyncExternalStore`) | `react`, `react-dom` |
| `@langchain/vue` | Vue 3+ | Composition API (`ref`, `computed`, `watch`) | `vue` |
| `@langchain/svelte` | Svelte 4/5 | Stores (`writable`, `derived`) / Runes | `svelte` |
| `@langchain/angular` | Angular 18-20 | Signals (`signal`, `computed`, `effect`) | `@angular/core` |

All four packages expose a unified `useStream` API adapted to each framework's idiomatic reactivity system, providing:

- **Real-time streaming** of LangGraph agent state via SSE
- **Optimistic updates** for instant UI feedback
- **Interrupt handling** for human-in-the-loop workflows
- **Thread management** including history, branching, and thread switching
- **Subagent tracking** for multi-agent graphs
- **Tool call lifecycle** tracking
- **Submission queue** for batching rapid user inputs (new feature)
- **Custom transport** support via `FetchStreamTransport` or user-defined implementations

### SDK Core Refactoring (`@langchain/langgraph-sdk`)

The React-specific streaming implementation has been extracted from the core SDK into the new `@langchain/react` package. In its place, the core SDK now exports a **framework-agnostic UI module** (`@langchain/langgraph-sdk/ui`) containing shared primitives:

- **`StreamManager`** — Core stream lifecycle and event processing
- **`MessageTupleManager`** — Message state normalization between dict and class formats
- **`SubmitQueue`** — Client-side queue that drains submissions sequentially when the agent is idle, with configurable error handling (`"continue"` / `"stop"`)
- **`FetchStreamTransport`** — Pluggable SSE transport abstraction using the Fetch API
- **`extractInterrupts`** — Unified interrupt extraction from stream state and thread state
- **`getBranchContext` / `getMessagesMetadataMap`** — Thread history branching utilities
[...]

PR #2010, opened 2026-02-27
fix(sdk-react): clear stale stream values after joinStream history refresh

### LINKED ISSUE #1969: joinStream's onSuccess doesn't clear stream values, causing stale state after reconnect
## Bug Description

After a stream reconnection via `joinStream` (either through the built-in `reconnectOnMount` or manual `joinStream()` calls), `useStream` can provide stale state to the consuming component even though the hook correctly fetches the latest state from the server.

In practice, both `submit` and `joinStream` should see the same `values` events for a given run — `values` events are emitted after each superstep that produces output changes, not as a completion signal. Under normal conditions both paths work fine. The difference is fault tolerance: if the final `values` event is missed or never delivered for any reason (network issues, backend edge cases, etc.), `submit`'s `onSuccess` gracefully falls back to the freshly fetched state via `history.mutate()` + `return null`, while `joinStream`'s `onSuccess` fetches the same fresh state but fails to use it because it doesn't clear `stream.values`.

## Root Cause

In `submit`'s `onSuccess`, when history is successfully refetched, `return null` is used to signal the `StreamManager` to clear `stream.values`:

```javascript
// submit's onSuccess (stream.lgp.js)
async onSuccess() {
    if (rejoinKey) runMetadataStorage?.removeItem(rejoinKey);
    if (shouldRefetch) {
        const lastHead = (await history.mutate(usableThreadId))?.at(0);
        if (lastHead) {
            options.onFinish?.(lastHead, callbackMeta);
            return null; // ✅ Clears stream.values in StreamManager
        }
    }
}
```

However, `joinStream`'s `onSuccess` is missing this `return null`:

```javascript
// joinStream's onSuccess (stream.lgp.js)
async onSuccess() {
    runMetadataStorage?.removeItem(`lg:stream:${threadId}`);
    const lastHead = (await history.mutate(threadId))?.at(0);
    if (lastHead) options.onFinish?.(lastHead, callbackMeta);
    // ❌ No return null — implicitly returns undefined
}
```

In `StreamManager.start()` (manager.js), the return value of `onSuccess` determines whether stream values are cleared:

```javascript
const values = await options.onSuccess?.();
if (typeof values !== "undefined") this.setStreamValues(values);
```

When `onSuccess` returns `undefined` (as `joinStream`'s always does), `setStreamValues` is never called. The stale values accumulated during SSE streaming remain in `stream.values`.

Since the hook's values getter uses `stream.values ?? historyValues`, the stale `stream.values` shadows the fresh `historyValues` that was correctly fetched by `history.mutate()`.

## Reproduction

1. Start a stream with `useStream` (via `submit`)
2. Have the backend update state during the run (e.g., set `generating: true`)
3. Disconnect the SSE stream mid-run (e.g., network interruption) before the backend sends a final `values` event reflecting the completed state
4. Reconnect via `joinStream` (either manually or via `reconnectOnMount`)
5. The stream reconnects and the run has already completed on the server
6. Observe that `useStream`'s `values` still contains the mid-run state (e.g., `generating: true`) even though the server's state has `generating: false`

## Expected Behavior

After `joinStream` completes and successfully fetches the latest state via `history.mutate()`, `stream.values` should be cleared so that the fresh `historyValues` is used.

## Suggested Fix

Add `return null` to `joinStream`'s `onSuccess`, matching the pattern used by `submit`:

```javascript
async onSuccess() {
    runMetadataStorage?.removeItem(`lg:stream:${threadId}`);
    const lastHead = (await history.mutate(threadId))?.at(0);
    if (lastHead) {
        options.onFinish?.(lastHead, callbackMeta);
        return null; // Clear stream values, fall back to fresh historyValues
    }
}
```

## Affected Versions

Verified present in v1.0.0 and v1.6.2 (latest at time of writing).

## Impact

This affects any use of `joinStream`, including:
- The built-in `reconnectOnMount` behavior
- Manual `joinStream()` calls for retry/reconnect logic

pawel-twardziak:
## Summary

  This PR fixes a stale state issue in useStream/useStreamLGP after joinStream completes.

  When joinStream succeeds and history is refetched, the hook could keep showing stale local stream
  values if no final values event was applied.
  This change clears local stream state after a successful history refresh so the hook falls back to the
  latest /history values.

  ## What Changed

  - Updated libs/sdk/src/react/stream.lgp.tsx to reset local stream state (return null) after
    history.mutate(threadId) returns a latest head.
  - Added a regression test in libs/sdk-validation/src/tests/stream.test.tsx covering:
      - stale local values after submit
      - joinStream path where no final values event is delivered
      - expected fallback to refreshed history values

  ## Issue Fixed

  Fixes a bug where joinStream could leave UI state stale (e.g. generating: true, outdated messages) even
  after history was successfully refreshed.

  ## Relevant Context

  This is most visible in slow/interrupted stream flows where joinStream returns but no terminal values
  event updates local stream state.

  ## Testing

  - Added integration-style regression test: joinStream clears stale values when history is refetched.

Fix #1969


christian-bromann:
Thank you for raising the PR, we are currently working on new SDK packages to integrate in different frontend frameworks, see https://github.com/langchain-ai/langgraphjs/pull/2001. I will incorporate your PR in there manually.

pawel-twardziak:
Hi @christian-bromann thanks for that info, I forgot for a moment about that PR

PR #2015, opened 2026-03-02
[Feature] Add onerr callback for submit

hieusmiths:
Current Situation:
We are currently using the onError callback within useStream to handle all errors. However, this approach seems too generic and is primarily suitable for logging purposes rather than robust error handling.

Problem:
The generic error handling is not sufficient for user-facing scenarios where specific error conditions require particular user actions, such as retrying an operation.

Sample Use Case:

- A user submits a request to initiate a stream.
- The stream operation fails.
- The application needs to display a specific error message to the user, along with an option to retry the operation.
- No handle error, detect submit failed since internal submit using callback and don't throw the exeception

### DIFF
--- a/libs/sdk/src/react/stream.custom.tsx
+++ b/libs/sdk/src/react/stream.custom.tsx
@@ -231,6 +231,7 @@ export function useStreamCustom<
 
         onSuccess: () => undefined,
         onError(error) {
+          submitOptions?.onError?.(error, undefined);
           options.onError?.(error, undefined);
         },
       }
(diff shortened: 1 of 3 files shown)


christian-bromann:
Thank you for raising the PR, we are currently working on new SDK packages to integrate in different frontend frameworks, see #2001. I will incorporate your PR in there manually.

christian-bromann:
Added to my branch. Will go ahead and close. Please follow the PR for updates on release. Thank you!

PR #2019, opened 2026-03-02
fix: add explicit : symbol annotations to symbols used as property keys

hntrl:
## Description

TypeScript infers `unique symbol` type when `Symbol.for()` is used without an explicit type annotation. When these symbols are used as computed property keys on classes, this causes type incompatibility when multiple versions of the same package are present in a dependency tree.

This PR adds explicit `: symbol` annotations **only** to symbols that are:
1. Used as computed property keys on classes (using `[SYMBOL]: value` syntax)
2. Exported or used in exported classes that could cross package boundaries

## Changes

Updated 3 symbols in langgraph-core:

- `COMMAND_SYMBOL`: Used as property key on `CommandInstance` class
- `REDUCED_VALUE_SYMBOL`: Exported symbol used as property key on `ReducedValue` class  
- `UNTRACKED_VALUE_SYMBOL`: Exported symbol used as property key on `UntrackedValue` class

## Impact

- Fixes type compatibility issues when multiple package versions exist in dependency tree
- No runtime behavior changes - only type annotations
- Minimal, targeted fix that only touches symbols that actually cause cross-version issues

## Related

Follows the same pattern as langchain-ai/langchainjs#10243


hntrl:
Closing in favor of #2022 with rebased commits

PR #2023, opened 2026-03-02
fix: add explicit : symbol annotations to symbols used as property keys

hntrl:
Same as #2019 but on fresh branch to test CI

Adds explicit  type annotations to symbols used as computed property keys on classes to fix cross-version type compatibility.


hntrl:
Closing: The  type annotation approach causes TypeScript errors when symbols are used as index keys. The symbols already use Symbol.for() which provides runtime compatibility across versions. The unique symbol type incompatibility is a compile-time issue that doesn't affect runtime behavior.

PR #2035, opened 2026-03-06
fix(core): prevent Zod schema defaults from overwriting checkpoint state in Command.update

JadenKim-dev:
Fixes #2031

When using `Command({ update })` with a Zod schema that has `.default()` on fields, Zod's `parse()` injects default values for missing fields into the update object. These injected keys then overwrite values restored from the checkpoint.

**Root cause:** `_validateInput` passes `Command.update` through `interopParse(schema, input.update)`, which triggers Zod defaults for any field not present in the update.

**Fix:** After parsing, filter the result to only include keys that were present in the original update input — matching the behavior of `StateSchema.validateInput` which already iterates only over `Object.entries(data)`.

Added regression tests for both Zod v3 and v4.


christian-bromann (approved):
LGTM 👍 

Thanks for your contribution!

PR #2039, opened 2026-03-08
fix(langgraph-checkpoint-redis): use blob storage for channel values to match Python implementation

JadenKim-dev:
## Summary

Replaces the parent-chain traversal approach (`mergeParentChannelValues`) with a blob storage pattern that mirrors the [Python checkpoint saver architecture](https://github.com/langchain-ai/langgraph/blob/main/libs/checkpoint-postgres/langgraph/store/postgres/base.py).

### Changes

- **`dumpBlobs()`**: At `put()` time, each channel value is stored separately under a `checkpoint_blob:{threadId}:{ns}:{channel}:{version}` key. Only channels present in `newVersions` are written; unchanged channels keep their existing blobs.
- **`loadBlobs()`**: At `getTuple()` time, `channel_values` are restored by looking up each channel's blob directly via its version number, eliminating the need to walk the parent chain.
- **Backward-compatibility fallback**: If no blob keys are found (`hasBlobKeys=false`), the reader falls back to `channel_values` embedded in the checkpoint document, preserving support for data written by older versions of the saver.
- **TTL support**: Blob keys are included in TTL application alongside the checkpoint key.

### Motivation

The previous implementation stored filtered `channel_values` inline in the checkpoint document and reconstructed the full state by traversing the parent chain (`mergeParentChannelValues`). This diverged from the Python implementation and caused issues in multi-backend deployments where Python and JS savers share the same Redis instance.

This change makes the JS `RedisSaver` structurally compatible with the Python implementation.

### Related

- Similar alignment work for Postgres: #1968
- Fixes langchainjs issue: langchain-ai/langchainjs#10291

## Test plan

- [x] Integration tests cover `put()` + `getTuple()` round-trip with blob storage
- [x] Backward-compatibility fallback tested (legacy data without blob keys)
- [x] `ShallowRedisSaver` behavior verified (uses inline storage, unaffected)
- [x] TTL applied to blob keys


etay2000:
Seems like this has stalled. Is there a better option than the Checkpoint Redis that I should be using?

christian-bromann:
@JadenKim-dev would you be open addressing failing unit/lint checks?

PR #2166, opened 2026-03-11
fix(core,sdk): Reduce streaming overhead: metadata dedup, empty field stripping, subagent values filtering

christian-bromann:
This PR reduces LangGraph streaming overhead by introducing an opt-in `stream_protocol_version: "v2"` and applying a set of protocol optimizations behind it.

The main problem was that the stream was spending far more bytes on repeated metadata, repeated default fields, and repeated state snapshots than on actual user-visible content. The goal here is to reduce payload size without breaking existing clients.

### `v2` protocol opt-in

Added `stream_protocol_version: "v2"` across the API and SDKs.

- `v1` remains the default and preserves current behavior
- `v2` enables the new optimized stream behavior

### Metadata deduplication

For streamed LLM message chunks, `v2` sends full metadata only on the first chunk and omits it on subsequent chunks.

This removes a large amount of repeated per-chunk overhead while keeping reconnect / message reconstruction intact.

### Sparse message serialization

Under `v2`, the API strips empty/default fields from serialized message chunks, including:

- `tool_calls: []`
- `invalid_tool_calls: []`
- `tool_call_chunks: []`
- `additional_kwargs: {}`

This reduces noise on the wire while preserving the same client-side semantics.

### Subgraph values patches

For subgraph streams, `v2` now distinguishes between:

- `values|...` → full snapshot
- `values-patch|tools:*` → incremental patch

The first subgraph values event remains a bootstrap snapshot. Later updates can be emitted as patches containing only changed fields plus optional deleted keys.

On the SDK side, subagent values are merged when `values-patch` is received.

## Why this approach

The changes are versioned because some clients may rely on the existing protocol shape, especially metadata being present on every chunk.

The patching work is currently scoped to subgraph values only. Root `values` events still behave as full snapshots, which keeps the rollout safer and avoids changing broader thread-state semantics all at once.

I also explicitly moved away from embedding delta markers inside user state and instead use a separate `values-patch` event, which makes the protocol cleaner and avoids mixing transport metadata with application data.

## Result

This gives us a substantial reduction in streaming overhead while preserving backward compatibility and creating a cleaner path toward more semantic, patch-oriented stream protocols in the future.

### Estimated impact

| Optimization | Savings per event | Cumulative (typical stream) |
|---|---|---|
| Metadata dedup | ~1,156 bytes | ~500 KB for 450-event stream |
| Empty field stripping | ~84 bytes | ~38 KB |
| Subagent values (SDK) | — | ~259 KB (messages portion) |
| Subagent values (server) | — | ~292 KB (full values events) |

### Testing

I ran a test with the following examples:

#### Basic Deep Agent

From our [React example](https://github.com/langchain-ai/langgraphjs/blob/8a84c0448eb2db7619c43be7c6e35daa058ac613/examples/ui-react/src/examples/deepagent):

__Before:__ 12.7 MB
__After:__ 4.1MB

[before.txt](https://github.com/user-attachments/files/25919893/before.txt)
[after.txt](https://github.com/user-attachments/files/25919895/after.txt)


christian-bromann:
There are still larger opportunities to reduce stream overhead, especially around `values` events, but I intentionally stopped short of making those changes in this PR because they would shift the protocol too far semantically.

In particular, I considered going further on subgraph and state streaming by:
- removing full `values|tools:*` snapshots entirely
- switching more of the protocol from snapshot-style events to patch-only events
- generalizing state streaming around explicit mutation semantics rather than repeated materialized state

I decided not to include those changes here because they would move the protocol from “same semantics, smaller payloads” into “different semantics, different client expectations.” That is a much larger contract change than the other optimizations in this PR.

Instead, this PR keeps the current protocol shape mostly intact and limits `v2` to optimizations that are still reasonably compatible:
- metadata deduplication
- sparse message serialization
- incremental subgraph value patches layered on top of an initial bootstrap snapshot

Rather than continuing to optimize the current protocol incrementally I am planning to explore a redesign. That work should be scoped as its own protocol effort, with dedicated design discussion and compatibility planning, instead of being folded into this PR.

PR #2208, opened 2026-03-18
fix(langgraph-checkpoint-redis): fix deleteThread using wrong key pattern for writes

jackjin1997:
## Summary

Fixes #2207

- Fix `deleteThread()` using incorrect `writes:` prefix instead of `checkpoint_write:` for write key deletion
- Add missing cleanup of `write_keys_zset:` entries, matching the correct implementation in `ShallowRedisSaver`

The bug was found by comparing `RedisSaver.deleteThread()` with `ShallowRedisSaver.deleteThread()` in `shallow.ts`, which correctly uses `checkpoint_write:` prefix and also cleans up zset keys.

## AI Disclosure

This bug was identified through code review with AI assistance. The fix aligns the standard `RedisSaver` implementation with the existing correct `ShallowRedisSaver` implementation.


christian-bromann (commented):
Thank you for the contribution!

Can we have tests for this?

jackjin1997:
Added unit tests in `deleteThread.test.ts` — verifies all three key patterns (`checkpoint:`, `checkpoint_write:`, `write_keys_zset:`) and the empty-keys edge case. Also updated the existing mock test in `checkpoint.int.test.ts` to use the corrected patterns.

jackjin1997:
@christian-bromann Thanks for the earlier feedback — unit tests are in place covering all three key patterns and the edge case. CI is passing. Anything else blocking this from moving forward?

christian-bromann (approved):
LGTM 👍

PR #2248, opened 2026-03-26
feat(react): add thread-subscribed stream hook

hntrl:
## Summary

This adds a new `useThreadStream` hook for React consumers who want to subscribe to thread-level SSE updates instead of run-level streams. It preserves the existing `useStream` behavior while providing an alternate API that still supports thread creation, queueing, reconnect metadata, and stream callbacks.

## Changes

- Add `useThreadStream` export in `@langchain/react` and `@langchain/langgraph-sdk/react` entrypoints.
- Extend the React LGP stream implementations to support `subscribeTo: "thread"` mode, which creates runs via `runs.create(...)` and consumes events via `threads.joinStream(...)`.
- Update join/reconnect behavior in thread mode to use thread stream subscription while preserving existing run-mode behavior.
- Replace `@ts-expect-error` checkpoint mutation with a typed property delete to keep TypeScript checks clean.


christian-bromann:
We published new streaming primitives that are purely thread based. Closing.

PR #2261, opened 2026-04-01
fix(redis): use correct key prefix in deleteThread() to clean up writes

venkat22022202:
## Summary

Fixes #2207

`RedisSaver.deleteThread()` uses the wrong key pattern (`writes:`) to find checkpoint writes, but writes are stored with the `checkpoint_write:` prefix. Additionally, it doesn't clean up `write_keys_zset:` entries. This causes orphaned keys and memory leaks in Redis.

## Root Cause

Key pattern mismatch between `putWrites()` (stores with `checkpoint_write:` prefix) and `deleteThread()` (scans with `writes:` prefix).

## Fix

- Changed `writes:${threadId}:*` to `checkpoint_write:${threadId}:*`
- Added `write_keys_zset:` cleanup matching the correct implementation in `ShallowRedisSaver`

The correct pattern already exists in `shallow.ts` L474-487 -- this PR aligns `RedisSaver` with `ShallowRedisSaver`.

## Test Plan

- Store checkpoints and writes for a thread via RedisSaver
- Call deleteThread()
- Before: checkpoint_write:* and write_keys_zset:* keys remain orphaned
- After: all keys for the thread are correctly deleted

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/libs/checkpoint-redis/src/index.ts
+++ b/libs/checkpoint-redis/src/index.ts
@@ -675,8 +675,6 @@ export class RedisSaver extends BaseCheckpointSaver {
   async deleteThread(threadId: string): Promise<void> {
     // Delete checkpoints
     const checkpointPattern = `checkpoint:${threadId}:*`;
-    // Use scan for better performance and cluster compatibility
-    // Use keys for simplicity - scan would be better for large datasets
     const checkpointKeys = await (this.client as any).keys(checkpointPattern);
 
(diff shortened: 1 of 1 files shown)


christian-bromann (changes requested):
Can we add proper unit tests for this change? Also let's rebase on top of current `main`.

PR #2262, opened 2026-04-01
fix(sdk): expose `toolCalls` and `getToolCalls` on `BaseStream` for all `useStream` usages

### LINKED ISSUE #10483: docs: useStream doesn't provide a toolCalls attribute despite it being mentioned in the docs
### Checklist

- [x] I included a link to the documentation page I am referring to (if applicable).

### Issue with current documentation:

The frontend documentation  for tool patterns names that stream.toolCalls can be used to directly render the toolcall with its associated message to the frontend. However, when testing this in the actual code, there is no attribute "toolCalls" that is available. 

https://docs.langchain.com/oss/javascript/langchain/frontend/tool-calling

### Idea or request for content:

_No response_

abhishekDeshmukh74:
Thank you for contributing to LangGraph.js! Your PR will appear in our next release under the title you set above.

## Problem

When users call `useStream<AgentState>` with a plain TypeScript interface (e.g. `interface AgentState { messages: BaseMessage[] }`), the returned object does **not** expose `toolCalls` or `getToolCalls`. But the docs show these properties as available and they do work when using `useStream<typeof myAgent>`.

Root cause: `toolCalls` and `getToolCalls` were only defined on `UseAgentStream`, which is the return type when using `useStream<typeof myAgent>` (a compiled `createAgent` result). When users pass a plain interface, `ResolveStreamInterface<T>` resolves to `BaseStream` — which was missing those properties despite them being **implemented at runtime** in both `stream.lgp.tsx` and `stream.custom.tsx`.

## Fix

Move `toolCalls` and `getToolCalls` from `UseAgentStream` to `BaseStream` so they are available for all `useStream<T>` usages.

## Changes

- **`libs/sdk/src/ui/stream/base.ts`**: Added `AIMessage` and `ToolCallWithResult` to imports; added `toolCalls: ToolCallWithResult<ToolCall>[]` and `getToolCalls(message)` to `BaseStream<StateType, ToolCall, Bag>`.
- **`libs/sdk/src/ui/stream/agent.ts`**: Removed now-duplicate `toolCalls`/`getToolCalls` from `UseAgentStream` (now inherited from `BaseStream`) and cleaned up unused imports.
- **`libs/sdk/src/ui/stream/index.ts`**: Updated `ResolveStreamInterface` docstring that incorrectly stated `BaseStream` does NOT include `toolCalls`.
- **`libs/sdk/src/tests/stream-types.test-d.ts`**: Added two type regression tests confirming `toolCalls` and `getToolCalls` are present on plain-state-type streams.

## Testing

All 298 existing tests pass. Two new type-level tests added.

Fixes langchain-ai/langchainjs#10483

### DIFF
--- a/libs/sdk/src/tests/stream-types.test-d.ts
+++ b/libs/sdk/src/tests/stream-types.test-d.ts
@@ -1,6 +1,10 @@
 import { describe, test, expectTypeOf } from "vitest";
 import type { BaseMessage } from "@langchain/core/messages";
-import type { Message } from "../types.messages.js";
+import type {
+  Message,
+  DefaultToolCall,
+  ToolCallWithResult,
+} from "../types.messages.js";
 import type { ThreadState } from "../schema.js";
 import type { BagTemplate } from "../types.template.js";
 import type { BaseStream, ResolveStreamInterface } from "../ui/stream/index.js";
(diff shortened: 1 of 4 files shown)


christian-bromann:
> Move `toolCalls` and `getToolCalls` from `UseAgentStream` to `BaseStream` so they are available for all `useStream<T>` usages.

I don't think this is the right fix. We intentionally separated `toolCall ` primitives from the `BaseStream` so that these properties wouldn't be available when streaming a StateGraph. I suggest we find a way to have the `useStream` function look into the attached type bag and if it contains tool information we can return a `UseAgentStream`.

christian-bromann:
Closing due to inactivity.

PR #2298, opened 2026-04-07
docs(README): replace `url` with `apiUrl` for FetchStreamTransport

Li13:
### Summary

In the README document, the parameter for FetchStreamTransport is used incorrectly; it should be apiUrl but url is mistakenly used instead.

### Testing

not run (documentation-only change)


christian-bromann:
@Li13 thanks for the contribution. We have updated our docs in a separate PR.

PR #2304, opened 2026-04-09
fix(sdk): resolve incorrect fork point for repeated branch edits

### LINKED ISSUE #2295: The branching feature in `useStream` is broken
### Checked other resources

### Example Code

In the docs page `https://docs.langchain.com/oss/python/langchain/frontend/branching-chat` , where example is there, even that itself is broken

_No response_

### Description

When i use chat branching, and keep editing the human message, if i edit more than 3 times, its broken and stuck at `2/2`. I tried increasing the history loading to more than default 10 , but still did not solve, since more than 10 checkpoints come only after 4 edits, and at 3rd edit itself, it is broke. 

I have attached the video here for reference. Note that after the 3rd edit, the branch switcher is stuck at `2/2`.

https://github.com/user-attachments/assets/9b642ec1-9d8c-4f5b-98c9-f9918ad2d805

### System Info

Ran the example from docs itself, and it fully broken.

seungwoo-ji-03:
When forking a message, the server inserts an intermediate INPUT checkpoint between the fork point and the new message. The existing UI pattern uses `firstSeenState.parent_checkpoint` as the fork target, but this points to the intermediate checkpoint rather than the actual shared ancestor — causing every subsequent edit to nest under the previous fork instead of branching flat.

This PR adds two fixes:

- **`forkParentCheckpoint` field on `MessageMetadata`**: starting from `firstSeenState.parent_checkpoint`, climbs up through parent checkpoints while the current one is part of the message's own branch path (i.e. an intermediate fork artifact). Stops once it leaves the branch path, giving the true fork point.
- **Adaptive history fetch**: when the default `limit=10` clips branch roots (detected by orphaned parent count > 1), automatically doubles the limit until the full branch tree is visible. Explicit user-specified limits are preserved.

All framework adapters (React, Vue, Svelte, Angular) and the docs example updated to use `forkParentCheckpoint ?? firstSeenState.parent_checkpoint` for backward compatibility.

Fixes #2295


christian-bromann (approved):
LGTM 👍

Jourdelune:
Is it still not merged?

christian-bromann:
@seungwoo-ji-03 sorry for the delay, we have been working on a better streaming protocol which caused this patch to fall behind. I have an [`updated example`](https://github.com/langchain-ai/streaming-cookbook/tree/cb/branching/typescript/ui-react/src/branching) that uses the v1 SDK packages. Given the complexity the provider has gotten I am probably will be adding a `useBranchedMessages` or something like that to encapsulate this into its own, well tested hook.

seungwoo-ji-03:
@christian-bromann Thanks for the update and the example link! Since you'll be adding `useBranchedMessages` natively, I'll close this PR.

PR #2314, opened 2026-04-13
feat(core): event based streaming

christian-bromann:
All stream v2 changes consolidated.


examples/ui-react-transport/src/app/server.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /**
+     * If user is subscribing to the other channels, return the stream of the other channels.
+     */
+    return new Response(
+      await this.#graph.streamEvents(body.input, {
+        version: "v3",
+        encoding: "text/event-stream",
+      })
```
hntrl: x2 make sure encoding documented

examples/ui-react-transport/src/transport.ts
hntrl, line 1: either:
* simplify it
* or make it a export from somewhere

assume everything that we put in examples is stuff that lives in user code, so it should be pretty easy to understand

libs/langgraph-core/src/pregel/index.ts
```diff
@@ -125,6 +140,59 @@ import { interrupt } from "../interrupt.js";
...
+> & {
+  version: "v3";
+  /** User-supplied transformer factories for custom projections. */
+  transformers?: TTransformers;
+  /**
+   * Configuration forwarded to the built-in lifecycle transformer
+   * that powers `run.lifecycle` and the `lifecycle` channel events
+   * in the protocol stream.
+   */
+  lifecycle?: LifecycleTransformerOptions;
```
hntrl: what is this needed for?

libs/langgraph-core/src/pregel/index.ts
```diff
@@ -1893,9 +2011,123 @@ export class Pregel<
...
+      ...restOptions
+    } = options;
+
+    const streamOptions = {
+      recursionLimit: this.config?.recursionLimit,
+      ...restOptions,
+      configurable: {
+        ...this.config?.configurable,
+        ...restOptions?.configurable,
+        [PROTOCOL_MESSAGES_STREAM_CONFIG_KEY]: true,
```
hntrl: whats this for

libs/langgraph-core/src/pregel/loop.ts
```diff
@@ -1203,6 +1228,53 @@ export class PregelLoop {
...
+    entries: [StreamMode, unknown][]
+  ): void {
+    const meta = this._currentCheckpointMeta();
+    if (!meta) {
+      this._emit(entries);
+      return;
+    }
+    this._emit(
+      entries.map(([mode, payload]) =>
+        mode === "values" ? [mode, payload, meta] : [mode, payload]
```
hntrl, line 1273: why is meta only attached on values stream?

libs/langgraph-core/src/pregel/loop.ts
hntrl, line 1: we might not need these changes?

libs/langgraph-core/src/pregel/messages-v2.ts
hntrl, line 1: we should either:
* upstream most of of the changes here to @langchain/core (ripping from ChatModelStream)
* create the js equivalent of `_V2StreamingCallbackHandler` and use that instead of inferring from handleLLMNewToken

libs/langgraph-core/src/stream/convert.ts
```diff
@@ -0,0 +1,273 @@
...
+ *   paired with a `values` chunk).
+ * @returns An ordered list of {@link ProtocolEvent}s ready for downstream
+ *   reducers. Callers advance their `seq` counter by `result.length`.
+ */
+export function convertToProtocolEvent(
+  ns: Namespace,
+  mode: StreamMode,
+  payload: unknown,
+  seq: number,
+  meta?: StreamChunkMeta
```
hntrl: uber nit: dont use positional args

libs/langgraph-core/src/stream/mux.ts
```diff
@@ -0,0 +1,527 @@
...
+    (typeof value === "object" || typeof value === "function") &&
+    typeof (value as { then?: unknown }).then === "function"
+  );
+}
+
+/**
+ * Symbol key used by {@link StreamMux} to resolve the values promise on a
+ * stream handle. Using a symbol keeps this off the public autocomplete surface.
+ */
+export const RESOLVE_VALUES: unique symbol = Symbol("resolveValues");
```
hntrl, line 50:
```suggestion
export const RESOLVE_VALUES: symbol = Symbol.for("resolveValues");
```

libs/sdk-react/docs/context.md
```diff
@@ -0,0 +1,99 @@
...
+- [Custom adapters through `StreamProvider`](#custom-adapters-through-streamprovider)
+
+## Basic usage
+
+```tsx
+import { StreamProvider, useStreamContext } from "@langchain/react";
+
+function App() {
+  return (
+    <StreamProvider assistantId="agent" apiUrl="http://localhost:2024">
```
hntrl, line 19: do we want assistantId + apiUrl parameters to be configured at the transport level? E.g. what transport we call is not determined by the props that get passed in

libs/sdk-react/docs/interrupts.md
```diff
@@ -0,0 +1,122 @@
...
+- [`respond(response, target?)`](#respondresponse-target)
+- [Headless tools](#headless-tools)
+- [Lower-level helpers](#lower-level-helpers)
+
+## Reading interrupts
+
+The root hook exposes the latest interrupt and the full list:
+
+```tsx
+const { messages, interrupt, submit } = useStream<
```
hntrl, line 18: in docs can we use `interrupts`?

hntrl (commented):
intentionally glossing over a lot of the example + sdk-* code

libs/langgraph-core/src/pregel/loop.ts
christian-bromann: `meta` is only attached to `values` because this checkpoint envelope describes the persisted state snapshot, not every emitted event. A `values` chunk is the stream item that corresponds to “this is the graph state captured by checkpoint X, whose parent is Y.” Other modes like updates, messages, tools, and tasks are partial/lifecycle events and don’t map cleanly to a checkpointed state that users can fork from.

The new stream layer then turns that sidecar into a separate lightweight checkpoints protocol event immediately before the paired values event. That gives clients a dedicated checkpoints channel without overloading the legacy checkpoints stream mode, which still has the full debug checkpoint shape.

libs/langgraph-core/src/stream/mux.ts
christian-bromann: `Symbol.for("resolveValues")` changes runtime behavior: it uses the global symbol registry. That is useful when independently loaded code needs to rediscover the same symbol by string, but it also creates possible collisions, especially with a generic key like "resolveValues". Here the code already shares the key by importing the exported constant, so Symbol() is cleaner.

I also don't think we have to be concerned about type collisions as this is all internal logic within the same package, nothing reaches to a different package.

libs/sdk-react/docs/context.md
christian-bromann: Providing an `apiUrl` expects the server to implement an Agent Server interface that respects `assistantId` as a form to identify the target agent. If you would use a custom endpoint/server implementation this would just be:

```ts
const transport = new HttpAgentServerAdapter({
  apiUrl: window.location.origin,
  threadId,
  paths: {
    commands: `/api/threads/commands`,
    stream: `/api/threads/stream`,
  },
});
<StreamProvider transport={transport}>
```

libs/sdk-react/docs/interrupts.md
christian-bromann: Looking at this for the 2nd time, I think this is a nice convenience API. In reality a user can only respond to one interrupt at a time. Unless the application requires to respond to specific interrupts before responding to others, in most cases developers would just do:

```ts
if (interrupt) {
  return <InterruptForm interrupt={interrupt} />
}
```

instead of:

```ts
if (interrupt.length > 0) {
  return <InterruptForm interrupt={interrupt[0]} />
}
```

libs/langgraph-core/src/pregel/messages-v2.ts
christian-bromann: Implemented.

libs/langgraph-core/src/pregel/index.ts
christian-bromann: removed

libs/langgraph-core/src/graph/state.ts
```diff
@@ -1248,6 +1263,10 @@ export class CompiledStateGraph<
   NodeReturnType = unknown,
   InterruptType = unknown,
   WriterType = unknown,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  TStreamTransformers extends ReadonlyArray<
+    () => import("../stream/types.js").StreamTransformer<any>
```
hntrl: ?

hntrl (dismissed):

hntrl (approved):

christian-bromann:
🎉

PR #2316, opened 2026-04-14
fix(sdk): Backfill truncated history for regenerate branching

### LINKED ISSUE #2306: The regenerate feature in useStream is broken
### Checked other resources

### Example Code

In the page `https://docs.langchain.com/oss/python/langchain/frontend/branching-chat`, with the given exemple

_No response_

### Description

When I send multiple messages in the chat (e.g. 7 messages) and then try to regenerate the first one, the system doesn’t regenerate from the correct point. Instead, it ends up regenerating from a later message (for example the 4th one).

One workaround is to increase the fetchStateHistory limit, but this only partially fixes the issue. It works up to a certain conversation length, and then the problem appears again once the number of checkpoints exceeds that limit.

https://github.com/user-attachments/assets/6df0a8ef-e6f6-4f4f-a9e1-0d179665d219

### System Info

Ran the example from docs itself, and it fully broken.

ibrahim1023:
## Summary

This fixes incorrect regenerate branching in `useStream` when `fetchStateHistory`
is enabled with the default bounded history window.

In longer chats, the initial history fetch can exclude the earliest checkpoint
for older visible messages. When that happens, `getMessagesMetadata()` can
resolve an incorrect `firstSeenState`, which causes regenerate to branch from a
later checkpoint instead of the selected message's actual parent checkpoint.

This change backfills older history pages when the initial history response is
truncated, so message metadata can resolve against the full checkpoint chain.
The fix is applied in both:

- `libs/sdk-react/src/stream.lgp.tsx`
- `libs/sdk/src/react/stream.lgp.tsx`

## Testing

- Added a regression test covering a 7-turn conversation and regenerating the
  first AI message in `libs/sdk-react/src/tests/stream.test.tsx`
- Ran `pnpm format:check`
- Ran `pnpm build`
- Ran `pnpm --filter @langchain/react run test`

## Notes

This is a correctness-first fix for truncated history. Any optimization to avoid
eagerly backfilling older history pages can be handled separately.

Related: #2304

`#2304` addresses repeated forks / intermediate fork checkpoints. This PR is
still needed for `#2306` because it fixes incorrect regenerate behavior when
older checkpoints fall outside the initial history window in long chats.

Fixes #2306

### DIFF
--- a/libs/sdk-angular/src/tests/stream.test.ts
+++ b/libs/sdk-angular/src/tests/stream.test.ts
@@ -601,6 +601,37 @@ it("branching", async () => {
     .toHaveTextContent("1 / 2");
 });
 
+it("branching: regenerating the first ai message after a long chat resets to that checkpoint", async () => {
+  const screen = await render(BranchingComponent);
+
+  for (let i = 0; i < 7; i += 1) {
+    await screen.getByTestId("submit").click();
+    await expect
+      .element(screen.getByTestId(`content-${i * 2}`), { timeout: 10_000 })
+      .toHaveTextContent("Hello");
+    await expect
+      .element(screen.getByTestId(`content-${i * 2 + 1}`), { timeout: 10_000 })
+      .toHaveTextContent("Hey");
+  }
+
+  await expect
+    .element(screen.getByTestId("message-13"), { timeout: 10_000 })
+    .toHaveTextContent("Hey");
+
+  await screen.getByTestId("regenerate-1").click();
+
+  await expect
+    .element(screen.getByTestId("message-0"), { timeout: 10_000 })
+    .toHaveTextContent("Hello");
+  await expect
+    .element(screen.getByTestId("message-1"), { timeout: 10_000 })
+    .toHaveTextContent("Hey");
+  await expect
+    .element(screen.getByTestId("branch-info-1"), { timeout: 10_000 })
+    .toHaveTextContent("2 / 2");
+  await expect.element(screen.getByTestId("message-2")).not.toBeInTheDocument();
+});
+
 it("fetchStateHistory: { limit: 2 }", async () => {
   const requests: Array<{ url: string; body?: Record<string, unknown> }> = [];
   const client = new Client({
--- a/libs/sdk-react/src/stream.lgp.tsx
+++ b/libs/sdk-react/src/stream.lgp.tsx
@@ -25,6 +25,7 @@ import type {
 import { Client, getClientConfigHash } from "@langchain/langgraph-sdk/client";
 import {
   filterStream,
+  fetchHistory,
   unique,
   StreamError,
   getBranchContext,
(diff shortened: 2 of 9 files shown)


Jourdelune:
Could be nice to fix the issue for other frontend framework too, like Vue!

ibrahim1023:
> Could be nice to fix the issue for other frontend framework too, like Vue!

Yeah I can add that!

Jourdelune:
Ty! But langgraph is also available in angular and svelte, maybe there is a cleaner way to handle this that editing each of the frontend lib directly.

ibrahim1023:
Good point. I cleaned this up so the truncated-history backfill now lives in a shared SDK helper instead of being duplicated as framework-specific logic.

That shared path is now used by the stream orchestration/history flow across the frontend packages, and I verified the long-chat regenerate case in React, Vue, Angular, and Svelte as part of the update.

christian-bromann:
Thanks @ibrahim1023 for digging into this and for the thorough fix + regression tests across React, Vue, Angular, and Svelte. The diagnosis is spot on: with the legacy `useStream` API, regenerate/edit flows depend on `getMessagesMetadata()` scanning `fetchStateHistory` to resolve each message’s `firstSeenState`. When the bounded history window truncates older checkpoints, that lookup can point at the wrong parent checkpoint — exactly what you reproduced in #2306.

We’re going to close this PR because the issue is addressed in the **new streaming primitives** (v1 SDK / v2 protocol), which take a different approach rather than backfilling truncated history pages.

**What changed in the new model:**

1. **Per-message checkpoint metadata at stream time** — The stream controller tracks `parentCheckpointId` for each message via `MessageMetadataTracker`, pairing `checkpoints` events with their companion `values` events as they arrive. This is recorded when the message is first observed, not inferred later from a partial history fetch.

2. **Fork/edit/regenerate without `fetchStateHistory`** — Branching is expressed as:
   - `useMessageMetadata(stream, msg.id)` → `{ parentCheckpointId }`
   - `submit(input, { forkFrom: { checkpointId } })`

   This removes the dependency on scanning thread history to find the correct fork point, so long chats no longer break when the initial history window is too small.

3. **Legacy API removed** — `fetchStateHistory`, `getMessagesMetadata`, and `branch` / `setBranch` are dropped in v1. See the migration guide and fork docs:
   - [`libs/sdk-react/docs/v1-migration.md`](https://github.com/langchain-ai/langgraphjs/blob/main/libs/sdk-react/docs/v1-migration.md) (§4–§6)
   - [`libs/sdk-react/docs/fork-from-checkpoint.md`](https://github.com/langchain-ai/langgraphjs/blob/main/libs/sdk-react/docs/fork-from-checkpoint.md)

The core also now emits checkpoint metadata alongside `values` (via `_currentCheckpointMeta()` in the pregel loop), which backs the client-side metadata tracker.

**Net:** Your backfill approach is a reasonable correctness fix for the legacy API, but the new primitives fix the root cause by design — regenerate always forks from the message’s recorded parent checkpoint, independent of how much history was prefetched.

Closing as superseded. #2306 should be resolved for apps on `@langchain/react` v1+ (and the other v1 frontend packages). Thanks again for the contribution and for calling out the Vue/Angular/Svelte coverage — that feedback helped shape the shared orchestrator work.

PR #2317, opened 2026-04-14
feat(supervisor): widen agents type to accept createAgent graphs

fish895623:
## Summary

- Add a broader `CompiledStateGraph<any, any, string, any, any>` to the `agents` union in `CreateSupervisorParams`
- Graphs produced by `createAgent` from `langchain` (via `.graph`) are now accepted alongside existing `createReactAgent` graphs and `RemoteGraph`
- The original `AnnotationRootT`-parameterized `CompiledStateGraph` type is preserved for backward compatibility

## Motivation

The new `createAgent` API in the `langchain` package returns a `ReactAgent` whose `.graph` property is a `CompiledStateGraph` with a different state schema (`BuiltInState`) than the `MessagesAnnotation`-based state from `createReactAgent`. Since `createReactAgent` is deprecated in favor of `createAgent`, `createSupervisor` needs to accept both graph types.

At runtime this already works — `makeCallAgent` types its `agent` parameter as `any` and only accesses `.name`, `.invoke()`, and optionally `.description`. The type constraint on the `agents` parameter was simply too narrow for the new API.

### `@langchain/langgraph-supervisor` (`libs/langgraph-supervisor`)

- Updated `CreateSupervisorParams.agents` type to include `CompiledStateGraph<any, any, string, any, any>` in the union alongside the existing strictly-typed `CompiledStateGraph` and `RemoteGraph`

## Test plan

- [x] `pnpm build` passes for `@langchain/langgraph-supervisor`
- [x] Consuming project using `createAgent` + `createSupervisor` compiles without errors
- [x] Existing supervisor tests still pass


christian-bromann (approved):
👍

PR #2320, opened 2026-04-14
feat: add MongoDBStore, new long term memory store

PavelSafronov:
This PR adds MongoDBStore, new long term memory store

Fixes MongoDB issue: https://jira.mongodb.org/browse/NODE-7497


libs/checkpoint-mongodb/src/store.ts
```diff
@@ -0,0 +1,797 @@
...
+    {
+      client,
+      dbName,
+      collectionName,
+      enableTimestamps,
+      embeddings,
+      indexConfig,
+      ttl,
+    }: MongoDBStoreParams,
+    serde: SerializerProtocol
```
RaschidJFR: Why is a sede required? 
can it be optional to reduce user friction?

libs/checkpoint-mongodb/src/tests/store.test.ts
```diff
@@ -0,0 +1,572 @@
...
+        namespace: ["documents"],
+        key: "doc1",
+        value: { title: "Test", secret: "hidden", content: "body" },
+        index: ["title", "content"], // Only embed title and content
+      };
+
+      await storeWithEmbeddings.batch([operation]);
+
+      // Should have called embedDocuments with only the indexed fields
+      expect(mockCollection.bulkWrite).toHaveBeenCalled();
```
RaschidJFR, line 186: Is this assertion complete?

libs/checkpoint-mongodb/src/store.ts
```diff
@@ -0,0 +1,797 @@
...
+        const now = new Date();
+        const doc: Record<string, any> = {
+          namespace,
+          key,
+          value: serializedValue,
+          type,
+          valueIndex: value,
+          namespacePath: computeNamespacePath(namespace),
+          updatedAt: now,
+        };
```
RaschidJFR, line 303: - This is saving duplicate data as `value` (binary) and `valueIndex` (object). Do we need both?
- Similarly with `namespace` and `namespacePath`. I understand the latter is needed to apply filters in Atlas vector search. Do we still need `namespace` then?

libs/checkpoint-mongodb/src/tests/store.test.ts
PavelSafronov: Yup. `toHaveBeenCalled` returns `void`, and commenting out the call on line 210 will result in the correct assertion being thrown from this spot.

libs/checkpoint-mongodb/src/store.ts
PavelSafronov: Dang, I followed the checkpoint pattern a bit closely. Stores are dealing with JSON data, so no need for a serializer. Removing.

libs/checkpoint-mongodb/src/store.ts
PavelSafronov: Removing the binary value, we don't need them both.

Correct, the denormalized `namespacePath` allows quicker namespace filtering.
It's also something that the Python MongoDBStore does as well: https://github.com/langchain-ai/langchain-mongodb/blob/main/libs/langgraph-store-mongodb/langgraph/store/mongodb/base.py#L690-L700

But do we still need `namespace`? Yes, it helps in other scenarios:
- Put would have to perform an `$elemMatch` query
- List would have to perform a string split in the pipeline
- Search would have to do array parsing

We would potentially save a dozen bytes per item by not storing `namespace`, but we'd end up with worse queries.

libs/checkpoint-mongodb/src/store.ts
```diff
@@ -477,7 +462,13 @@ export class MongoDBStore extends BaseStore {
...
    * Search for items by namespace prefix and filter criteria.
    * Supports both field-based filtering (structured search) and vector similarity search.
-   * If a query string is provided and embeddings are configured, performs semantic search.
+   *
+   * Field-based search (filter parameter): Always available, searches valueIndex fields.
+   * Vector search (query parameter): Requires embeddings to be configured.
+   *   - If embeddings are provided: Performs semantic search on the query
+   *   - If embeddings are NOT provided: Throws an error. In this mode, MongoDB Atlas
+   *     auto-embedding is assumed to be enabled, but semantic search is not supported.
+   *     Use field-based filtering instead.
```
RaschidJFR, line 471: Not quite sure how Auto-embeddings would work here... will users be able to perform a vector search at all with autoembeddings? If so, could you add a test case to explain?

libs/checkpoint-mongodb/src/store.ts
```diff
@@ -0,0 +1,783 @@
...
+    }
+
+    // Otherwise, perform structured field-based search
+    // Build MongoDB query for namespace and field filtering
+    const mongoQuery: Record<string, any> = {};
+
+    // Filter by namespace if provided (exact match)
+    if (namespacePrefix && namespacePrefix.length > 0) {
+      mongoQuery.namespace = namespacePrefix;
+    }
```
tadjik1, line 503: The comment and implementation are using exact match, however, from the method signature (parameter called `namespacePrefix`), as well as from [in-memory](https://github.com/langchain-ai/langgraphjs/blob/main/libs/checkpoint/src/store/memory.ts#L253-L257) imeplementation it seems that this method should search for keys in **all** potentially nested namespaces behind the prefix. Is exact match - deliberate choice here, and if so, can you elaborate?

libs/checkpoint-mongodb/src/store.ts
```diff
@@ -0,0 +1,783 @@
...
+    };
+  }
+
+  /**
+   * List namespaces in the store, optionally filtered by prefix/suffix/depth.
+   */
+  private async listNamespacesOp(
+    op: ListNamespacesOperation
+  ): Promise<string[][]> {
+    const { limit, offset, namespacePrefix, maxDepth } =
```
tadjik1, line 395: From the caller side I see the following code:
```
const readOp = op;
if ("key" in readOp && !("namespacePrefix" in readOp)) {
  results[i] = await this.getOp(readOp as GetOperation);
} else if ("namespacePrefix" in readOp) {
  results[i] = await this.searchOp(readOp as SearchOperation);
} else {
  results[i] = await this.listNamespacesOp(
    readOp as ListNamespacesOperation
  );
}
```
which basically means we get into this method when no `namespacePrefix` is presented (last `else` block).

Consequently it means that the `namespacePrefix=undefined` and first stage of aggregation (`$match`), is not in the pipeline, so we return unfiltered namespaces.

libs/checkpoint-mongodb/src/store.ts
```diff
@@ -0,0 +1,783 @@
...
+          value, // Store JSON directly
+          valueIndex: value,
+          namespacePath: computeNamespacePath(namespace),
+          updatedAt: now,
+        };
+
+        // Add embedding if available and this operation was embedded
+        const embeddingIndex = Array.from(embeddingMap.entries()).find(
+          ([_, opIdx]) => opIdx === i
+        )?.[0];
```
tadjik1, line 308: Since it's inside `for (let i = 0; i < opsList.length; i++) {` it's going to be O(n*n), maybe we should build lookup map first for embeddings beforehand?
At the same time I'm not sure how big are typically batches (probably small), let me know what you think

libs/checkpoint-mongodb/src/store.ts
RaschidJFR: Great. `value` is no longer binary but `valueIndex` still exists. Shouldn't one of them go?

libs/checkpoint-mongodb/src/store.ts
RaschidJFR, line 631: I think this is wrong and should be:

```js
namespacePath: { $in: [namespacePrefixFilter] },
```

libs/checkpoint-mongodb/src/store.ts
RaschidJFR: I realized this wasn't possible with the current implementation so I implemented in this [commit](https://github.com/RaschidJFR/langgraphjs/commit/348739703414a666cc9a40b37572b8e4a39a010d) as a reference.

libs/checkpoint-mongodb/src/store.ts
```diff
@@ -0,0 +1,783 @@
...
+    if (namespacePrefix && namespacePrefix.length > 0) {
+      namespacePrefixFilter = namespacePrefix.join("/");
+    }
+
+    // Stage 1: Vector search with namespace filtering
+    const vectorSearchStage: Record<string, any> = {
+      $vectorSearch: {
+        index: this.indexConfig!.name,
+        path: "embedding",
+        queryVector: queryEmbedding,
```
tadjik1, line 622: Autoembeddings feature requires different qiery format `query: {text: ''}` and embeddings are calculated server-side.

tadjik1 (commented):
Thanks @PavelSafronov! I have one big concern here, autoembeddings seem to not working with the current implementation.

libs/checkpoint-mongodb/src/store.ts
```diff
@@ -0,0 +1,783 @@
...
+      });
+    }
+
+    return items;
+  }
+
+  /**
+   * Perform vector search using in-memory cosine similarity.
+   */
+  private async inMemoryVectorSearch(
```
tadjik1, line 683: This can be very expensive as we do this for all docs in memory. Why do we need this?

dariakp:
This can be closed in favor of https://github.com/langchain-ai/langgraphjs/pull/2326

PR #2326, opened 2026-04-17
feat(langgraph-checkpoint-mongodb): add MongoDBStore for long-term memory with manual and auto embedding

tadjik1:
## Summary

Adds `MongoDBStore` - a `BaseStore` implementation backed by MongoDB for long-term cross-thread memory. While the existing `MongoDBSaver` persists graph execution state within a single thread (checkpoints), `MongoDBStore` enables agents to store and retrieve data across threads and sessions - user preferences, learned facts, agent instructions, and more.

Fixes [NODE-7497](https://jira.mongodb.org/browse/NODE-7497)

Built on top of the initial implementation by @PavelSafronov and @RaschidJFR (PR #2320). Their work provided the foundation for the store, auto embedding integration, and Docker test setup.

## Review notes

This PR is organized into two self-contained commits that can be reviewed independently:

1. **[`add MongoDBStore with core CRUD and field-based search`](https://github.com/langchain-ai/langgraphjs/commit/1415f7f6c7df0338644444feffa14f8123e2f0a8)** - the store itself with all non-embedding functionality, namespace prefix matching, matchConditions, tests
2. **[`add manual and auto embedding support`](https://github.com/langchain-ai/langgraphjs/pull/2326/changes/7bc2ecd80840b1bd291d7dfcdc012a704b614040)** - adds vector search on top of the first commit, including IndexConfig, embedding logic in put/search, vector search index creation, tests

### Core store (commit 1)
- `MongoDBStore` extending `BaseStore` with Put, Get, Delete, Search, and ListNamespaces operations
- Hierarchical namespace organization with unique `(namespace, key)` index
- Field-based search with comparison operators (`$eq`, `$ne`, `$gt`, `$gte`, `$lt`, `$lte`)
- Namespace prefix matching in search using dot-notation array indexing
- ListNamespaces with `matchConditions` support (prefix, suffix, and wildcard filtering)
- TTL support via MongoDB TTL index with optional refresh-on-read
- Factory method `fromConnString` for convenient setup
- Refactors existing `MongoDBSaver` into separate `checkpoint.ts` file

### Embedding support (commit 2)
Two modes for vector similarity search via MongoDB `$vectorSearch`, both requiring `indexConfig`:

**Manual embedding** - app provides an `EmbeddingsInterface`:
  - Store computes vectors on put via `embedDocuments()`
  - Search computes query vector via `embedQuery()` and uses `queryVector` in `$vectorSearch`

**Auto embedding** - no `EmbeddingsInterface` needed:
  - Store writes plain text to the embedding field
  - MongoDB generates embeddings server-side via Voyage AI (requires `model` in `indexConfig`)
  - Search sends query text via `query.text` in `$vectorSearch`
  - Requires MongoDB 8.2+ with auto embedding support


libs/checkpoint-mongodb/src/store.ts
```diff
@@ -0,0 +1,799 @@
...
+   * Automatically creates and connects a MongoDB client, and calls
+   * {@link start} to ensure required indexes exist.
+   * @param connString MongoDB connection string
+   * @param params Store configuration
+   */
+  static async fromConnString(
+    connString: string,
+    params?: Omit<MongoDBStoreParams, "client">
+  ): Promise<MongoDBStore> {
+    const { MongoClient: MC } = await import("mongodb");
```
hntrl: whats with the dynamic import?

libs/checkpoint-mongodb/src/store.ts
```diff
@@ -0,0 +1,799 @@
...
+      const indexName = this.indexConfig.name;
+      const deadline = Date.now() + 120_000;
+      while (Date.now() < deadline) {
+        const indexes = await collection.listSearchIndexes(indexName).toArray();
+        const ready = indexes.some(
+          (idx: any) => idx.name === indexName && idx.status === "READY"
+        );
+        if (ready) break;
+        await new Promise((resolve) => setTimeout(resolve, 5_000));
+      }
```
hntrl: This polling mechanism feels awkward in the starting path? Do we need strong guarantees that search indexes are created before we allow write traffic?

libs/checkpoint-mongodb/src/store.ts
```diff
@@ -0,0 +1,799 @@
...
+    }
+  }
+
+  /**
+   * Clean up the store (no-op for now).
+   */
+  async stop(): Promise<void> {
+    // No-op
+  }
```
hntrl, line 797: We don't need to dispose of the client object?

hntrl (commented):
Thanks for the add! Few comments:

libs/checkpoint-mongodb/src/store.ts
tadjik1: Leftover from earlier work - there's no reason for it. mongodb is a regular dependency, so the static import is fine.

libs/checkpoint-mongodb/src/store.ts
tadjik1: Agreed, dropped it. Writes don't depend on the search index being ready - only `vectorSearch` does, and blocking `start()` (and therefore `fromConnString()`) for up to few minutes is too heavy for a path that's also valid for non-vector users.

libs/checkpoint-mongodb/src/store.ts
tadjik1: Good catch - the `fromConnString` path was leaking the client. Added an `ownsClient` flag set to `true` only in `fromConnString`. `stop()` now calls `this.client.close()` when the store created the client; user-supplied clients (passed into the constructor) are left alone since the caller owns their lifecycle.

hntrl (approved):

PR #2327, opened 2026-04-18
fix(sdk): clear MessageTupleManager on joinStream full replay

pawel-twardziak:
## What

When a network error interrupted an SSE stream and `useStream`'s `onError` handler called `joinStream(runId)` to reconnect, the server replayed all events from the beginning (default `lastEventId = "-1"`) and `MessageTupleManager.add()` concatenated the replay onto the chunks still held from the pre-disconnect stream - doubling assistant message content (`"Hello"` → `"HelloHello"`) and corrupting streamed `tool_call` args into unparseable JSON. The same bug affected per-subagent chunk managers inside `SubagentManager`, so subagent output doubled too on replay.

## Root cause

`StreamManager.enqueue()` had no reset of `MessageTupleManager.chunks` between runs. `joinStream` defaulted `lastEventId` to `"-1"` in both the orchestrator and the React hook, so every reconnect against servers without the SSE `Location` header (e.g. LangGraph Platform ≤ 0.4.46) re-delivered every chunk and collided with stale state. `streamWithRetry` did not catch this because it requires the server to advertise a reconnect path via `Location`.

## Fix

- Added `StreamManager.resetChunkAccumulator(existingMessages?)` in `libs/sdk/src/ui/manager.ts`. It clears `MessageTupleManager.chunks` and the per-subagent chunk managers, without touching `state.values`, the abort controller, or the subagent registry. When `existingMessages` is supplied, each id is seeded with its current index so the replay overwrites the visible message in place instead of appending a duplicate.
- Added `SubagentManager.resetChunkAccumulator()` in `libs/sdk/src/ui/subagents.ts`. Drops `messageManagers` + `pendingMessages` and zeroes each subagent's `messages` array, preserving the idempotent registry and namespace mapping.
- `StreamOrchestrator.joinStream` (`libs/sdk/src/ui/orchestrator.ts`) and the parallel React hook `joinStream` (`libs/sdk-react/src/stream.lgp.tsx`) call the reset before `stream.start(...)` when `lastEventId === "-1"`, passing the currently visible messages for index seeding.
- Left a `TODO` in the orchestrator pointing at the longer-term follow-up: have `StreamManager.enqueue` remember the last seen event id (the same approach `streamWithRetry` already uses at `libs/sdk/src/utils/stream.ts`) and thread it into `joinStream` automatically, so servers resume at the correct cursor instead of replaying from zero.

## Scope decisions

- The reporter suggested unconditionally calling `this.messages.clear()` at the top of every `enqueue()`. I did not take that approach - it also runs on normal `submit()` calls and hides the intent. Gating on `lastEventId === "-1"` inside `joinStream` is smaller, explicit about when it fires, and leaves the normal submit path untouched.
- `StreamManager.clear()` was intentionally not reused: it also aborts the in-flight controller and nulls `state.values`, which would produce a visible UI blink on reconnect.
- Subagent coverage is included in this PR because it is the same root cause and the symmetrically-named method keeps the semantics obvious. Subagent registrations remain idempotent by `tool_call_id`.
- Issue #1969 (`joinStream`'s `onSuccess` not clearing stream values) is related but independent and out of scope here.

## Testing

- `pnpm test` in `libs/sdk`: 310 tests pass (up from 307). New tests added:
  - `libs/sdk/src/ui/orchestrator.test.ts` - `joinStream replay (issue #2028)`: asserts the main stream's message content is not doubled after a full replay, and that streamed `tool_call` args survive the replay (they previously concatenated into invalid JSON, producing an empty `tool_calls` array).
  - `libs/sdk/src/ui/manager.test.ts` - `SubagentManager.resetChunkAccumulator drops per-subagent chunks without losing the registry (issue #2028)`: asserts per-subagent replay does not double subagent messages and that the subagent registry survives the reset.
- All three new tests fail on `main` (verified before the fix was wired in) and pass with the fix.
- `pnpm build:internal` passes clean for both `@langchain/langgraph-sdk` and `@langchain/react`; `attw` and `publint` report no problems; TypeScript type-check passes.
- Repo-level `pnpm lint` and `pnpm format:check` were not exercised locally because the environment used for this change did not have `oxlint` / `oxfmt` on `PATH`; those will run in CI.

## Changeset

`.changeset/joinstream-replay-doubles-messages.md` - patch bump for `@langchain/langgraph-sdk` and `@langchain/react`.

Fixes #2028

### DIFF
--- a/.changeset/joinstream-replay-doubles-messages.md
+++ b/.changeset/joinstream-replay-doubles-messages.md
@@ -1,0 +1,24 @@
+---
+"@langchain/langgraph-sdk": patch
+"@langchain/react": patch
+---
+
+fix(sdk, sdk-react): clear `MessageTupleManager` on `joinStream` full replay
+
+When a network error interrupted a stream and `joinStream` reconnected
+without a `lastEventId`, the server replayed events from the beginning
+and `MessageTupleManager.add()` concatenated the replay onto the chunks
+still held from the pre-disconnect stream — doubling the assistant's
+message content (`"Hello"` → `"HelloHello"`) and corrupting streamed
+tool_call args. The same doubling affected per-subagent chunk managers.
+
+`StreamManager` now exposes `resetChunkAccumulator(existingMessages?)`
+which drops chunk state (plus per-subagent chunks) without touching
+`state.values`, the abort controller, or the subagent registry. When
+`existingMessages` is supplied, each id is seeded with its current
+index so the replay overwrites messages at their original positions
+instead of appending duplicates. Both the orchestrator and the React
+hook's `joinStream` now call this before `stream.start(...)` when
+`lastEventId === "-1"`.
+
+Closes https://github.com/langchain-ai/langgraphjs/issues/2028.
--- a/libs/sdk-react/src/stream.lgp.tsx
+++ b/libs/sdk-react/src/stream.lgp.tsx
@@ -762,6 +762,15 @@ export function useStreamLGP<
     lastEventId ??= "-1";
     if (!threadId) return;
 
+    // See libs/sdk/src/ui/orchestrator.ts joinStream() and issue #2028:
+    // on a full replay the MessageTupleManager would concat the replay
+    // onto the chunks from the pre-disconnect stream, doubling content.
+    // Seed current messages' indexes so the replay overwrites them in
+    // place rather than appending duplicates.
+    if (lastEventId === "-1") {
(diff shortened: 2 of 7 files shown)


christian-bromann:
Thanks @pawel-twardziak for the contributions.

We have published a new major version for all framework SDKs. They all are using different primitives underneath , which makes given changes irrelevant. We will be deprecating the manager and orchestrator class soon.

Issue #2334, opened 2026-04-22
RedisSaver.put() delta-filters channel_values but getTuple() does not reconstruct them — channels not written by the last node are lost


MohMaherId:
# RedisSaver.put() delta-filters channel_values but getTuple() does not reconstruct them — channels not written by the last node are lost

### Checked other resources

- [x] I added a very descriptive title to this issue
- [x] I searched the LangGraph.js documentation with the integrated search
- [x] I used the GitHub search to find a similar question and found #594 (SqliteSaver variant) and #1996 (related has_writes overwrite)
- [x] I am sure that this is a bug in LangGraph.js rather than my code
- [x] The bug is not resolved by updating to the latest stable version (`@langchain/langgraph-checkpoint-redis@1.0.4`)

### Example Code

```typescript
import { Annotation, StateGraph } from "@langchain/langgraph";
import { MemorySaver } from "@langchain/langgraph-checkpoint";
import { RedisSaver } from "@langchain/langgraph-checkpoint-redis";
import { AIMessage, BaseMessage, HumanMessage, SystemMessage } from "@langchain/core/messages";

// Multi-channel state: different nodes write different channels
const GraphState = Annotation.Root({
  messages: Annotation<BaseMessage[]>({
    reducer: (left, right) => left.concat(right),
    default: () => [],
  }),
  category: Annotation<string>({
    reducer: (_, right) => right,
    default: () => "",
  }),
  status: Annotation<string>({
    reducer: (_, right) => right,
    default: () => "idle",
  }),
});

// Agent node: writes { messages } only
function agent(state: typeof GraphState.State) {
  const lastMsg = state.messages[state.messages.length - 1];
  const content = typeof lastMsg?.content === "string" ? lastMsg.content : "?";
  console.log(`[agent] messages before reply=${state.messages.length}`);
  return { messages: [new AIMessage(`I heard: "${content}"`)] };
}

// Post-process node: writes { status, category } — does NOT write messages
function postProcess(state: typeof GraphState.State) {
  console.log(`[post_process] messages=${state.messages.length}`);
  return { status: "completed", category: "2" };
}

function buildGraph(checkpointer: any) {
  return new StateGraph(GraphState)
    .addNode("agent", agent)
    .addNode("post_process", postProcess)
    .addEdge("__start__", "agent")
    .addEdge("agent", "post_process")
    .addEdge("post_process", "__end__")
    .compile({ checkpointer });
}

async function main() {
[...]

PR #2335, opened 2026-04-22
fix(checkpoint-redis): preserve all channel_values in multi-node graphs

MohMaherId:
## Summary

Old Fixes for 2334

`RedisSaver.put()` delta-filters `channel_values` based on `newVersions`, keeping only channels written by the current node. This causes **silent data loss** in any multi-node graph where the last node writes a subset of channels.

### Root cause

The delta-filter mirrors `PostgresSaver.put()` behavior, which only persists changed channel blobs. In PostgresSaver this works because blobs are stored in a **separate `checkpoint_blobs` table** keyed by `(thread_id, channel, version)`, and `getTuple()` reconstructs full state via a SQL JOIN across all blob versions.

`RedisSaver` stores the entire checkpoint (including `channel_values`) in a **single JSON document**. There is no separate blob storage and no reconstruction logic in `getTuple()`. When the final node writes its checkpoint, the delta filter strips channels not written by that node. Those channels are permanently lost.

### Fix

Remove the delta-filter block entirely. The `checkpoint` object passed to `put()` by the Pregel loop already contains the **complete** `channel_values` for all channels. The delta filter was discarding data that was correctly assembled upstream.

This is consistent with `MemorySaver` and `ShallowRedisSaver`, neither of which filter `channel_values`.

### Trade-off

Each checkpoint JSON document will be slightly larger (it includes unchanged channel data). For Redis's single-document model this is correct behavior.

### Test

Added integration test: multi-channel state where node B writes only `{status, category}` (not `messages`), verifying that `messages` survives in the stored checkpoint.

### Reproduction

```typescript
// 3-channel state: messages (reducer), status (last-write), category (last-write)
// Node A writes { messages }
// Node B writes { status, category }
// getState() → messages: [] ← BUG (should be the accumulated messages)
```

See #2334 for full reproduction code and root cause analysis.

### DIFF
--- a/libs/checkpoint-redis/src/index.ts
+++ b/libs/checkpoint-redis/src/index.ts
@@ -189,29 +189,7 @@ export class RedisSaver extends BaseCheckpointSaver {
     const checkpointId = checkpoint.id || uuid6(0);
     const key = `checkpoint:${threadId}:${checkpointNs}:${checkpointId}`;
 
-    // Copy checkpoint and filter channel_values to only include changed channels
     const storedCheckpoint = copyCheckpoint(checkpoint);
-
-    // If newVersions is provided and has keys, only store those channels that changed
-    // If newVersions is empty {}, store no channel values
-    // If newVersions is not provided (undefined), keep all channel_values as-is
-    if (storedCheckpoint.channel_values && newVersions !== undefined) {
-      if (Object.keys(newVersions).length === 0) {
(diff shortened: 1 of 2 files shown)


MohMaherId:
Closing this PR pending discussion on the preferred fix direction — see updated issue #2334 with both options (fix in put() vs fix in getTuple()).

MohMaherId:
I will submit a better fix that targets getTuple() blob based construction

PR #2336, opened 2026-04-22
fix(langgraph-checkpoint-redis): add blob reconstruction for channel_values in multi-node graphs

### LINKED ISSUE #2334: RedisSaver.put() delta-filters channel_values but getTuple() does not reconstruct them — channels not written by the last node are lost

MohMaherId:
## Summary

Fixes #2334

`RedisSaver.put()` delta-filters `channel_values` based on `newVersions`, storing only channels written by the current node. Unlike `PostgresSaver` (which stores channel blobs separately and reconstructs in `getTuple()`), `RedisSaver` had no reconstruction logic — channels not written by the last node were silently lost.

This PR completes the blob storage layer:

- **`put()`**: writes each changed channel as a version-keyed Redis JSON key (`checkpoint_blob:{thread_id}:{ns}:{channel}:{version}`)
- **`loadCheckpointWithWrites()`**: for each channel in `channel_versions` missing from `channel_values`, looks up the blob at the recorded version and reconstructs full state

The `checkpoint_blobs` RediSearch index already existed with the correct schema (`thread_id`, `checkpoint_ns`, `channel`, `version`, `type`) — it just wasn't wired up.

### Reproduction

```json
{
  "name": "redis-checkpoint-bug",
  "private": true,
  "type": "module",
  "scripts": {
    "dev": "tsx redis-bug.ts",
    "dev:redis": "REDIS_URL=redis://localhost:6379 tsx redis-bug.ts"
  },
  "dependencies": {
    "@langchain/core": "^1.1.40",
    "@langchain/langgraph": "^1.2.9",
    "@langchain/langgraph-checkpoint": "^1.0.1",
    "@langchain/langgraph-checkpoint-redis": "1.0.4"
  },
  "devDependencies": {
    "tsx": "^4.19.0"
  }
}
```

Save the reproduction code from #2334 as `redis-bug.ts`, then:
```bash
docker run -d --name redis-test -p 6379:6379 redis:7-alpine
pnpm install && pnpm dev:redis
```

### Test plan

- [x] Added integration test: multi-channel state where node B writes only `{status, category}` (not `messages`), verifying `messages` survives in `getTuple()` result
- [x] Passes `oxfmt --check` and `oxlint`
- [x] Includes changeset
- [x] CI: unit tests
- [x] CI: lint + format
- [x] CI: build


christian-bromann (changes requested):
The PR introduces `checkpoint_blob:*` keys but deleteThread() never deletes them. For parity, `PostgresSaver.deleteThread()` explicitly [deletes](https://github.com/langchain-ai/langgraphjs/blob/1bc53a69dfc003ccdcad4e9b1b7010b11da8e4d2/libs/checkpoint-postgres/src/index.ts#L668-L670) blobs first. Without this, blobs orphan forever (memory growth), and thread deletion is no longer complete (a problem for data-deletion/GDPR semantics).

MohMaherId:
Thanks for the review, @christian-bromann — good catch on the orphaned blobs.

Fixed: deleteThread() now also deletes the checkpoint_blob:${threadId}:* keys written by put(), so blobs no longer orphan and thread deletion is complete — matching PostgresSaver.deleteThread() parity.

Test coverage added:

deleteThread.test.ts — unit test asserting the checkpoint_blob:* keys are deleted.
checkpoint.int.test.ts — integration test against real Redis: writes a multi-node checkpoint so a blob is created, then asserts zero thread keys remain after deleteThread().
Also added a changeset (patch bump) covering the full feature. Rebased onto the latest main. Ready for another look

christian-bromann:
Thanks for addressing the `deleteThread()` blob cleanup issue. I think there’s one remaining blocker around TTL behavior before this can merge.

With `ttlConfig.refreshOnRead` enabled, `getTuple()` currently refreshes only the checkpoint key’s TTL. This PR makes checkpoints depend on separate `checkpoint_blob:*` keys for unchanged channel values, and those blob keys get a TTL on write. That means a read can keep the checkpoint alive while the blob keys it needs still expire. After that, later reads return the checkpoint but silently omit the reconstructed channels, which brings back the state-loss behavior for TTL users.

Could you update the reconstruction path to refresh TTL for every blob key it reads when `refreshOnRead` is enabled, and add a TTL integration test for the multi-node subset-channel case?

MohMaherId:
Thanks @christian-bromann — fixed in 7e770097.

Reads now refresh the TTL of every reconstructed checkpoint_blob:* key, gated on the same refreshOnRead && defaultTTL condition getTuple() uses for the checkpoint key, so the checkpoint and its blobs expire in lockstep — closing the silent data-loss window for TTL users.

Since the refresh lives in the shared reconstruction path (loadCheckpointWithWrites), reconstructed blob TTLs are refreshed on getTuple() reads (and also when list() reconstructs them). Note the checkpoint key's own TTL refresh remains in getTuple() only, as before — I kept that pre-existing behavior unchanged.

Added an integration test for the multi-node subset-channel case (erodes a reconstructed blob's TTL, reads, asserts it's refreshed back to the default — verified it fails without the fix), and updated the changeset.

libs/checkpoint-redis/src/index.ts
```diff
@@ -247,6 +247,30 @@ export class RedisSaver extends BaseCheckpointSaver {
     // Store metadata fields at top-level for searching
     this.addSearchableMetadataFields(jsonDoc, metadata);
 
+    // The newVersions filter above strips unchanged channels from the checkpoint document.
+    // Persist each changed channel as a separate blob so loadCheckpointWithWrites()
+    // can reconstruct the full channel_values from blobs written by earlier nodes.
+    if (newVersions !== undefined && checkpoint.channel_values) {
+      for (const [channel, version] of Object.entries(newVersions)) {
```
christian-bromann: Instead of a for loop, can we run this via `Promise.all` to run this in parallel?

christian-bromann (commented):
Also can we refresh the TTL of every blob referenced by checkpoint.channel_versions (not just the newVersions subset) so each blob a freshly-written checkpoint depends on stays alive in lockstep with the checkpoint doc. This is a write-side refresh, so it's correct independent of refreshOnRead.

libs/checkpoint-redis/src/index.ts
MohMaherId: One related edge case I want to flag for your input, @christian-bromann — separate from the write-side refresh I will add:

Both the read-side and write-side TTL refreshes use EXPIRE, which is a no-op on a key that's already gone. So they keep a blob alive only if it still exists at refresh time. There's still a narrow window where a carried-over blob can be lost entirely:

Node A writes the messages blob (TTL = 60m).
~60m pass with no read of this thread.
Node B writes a new checkpoint (changing only status). If the messages blob expired in the gap before this write, the write-side refresh EXPIREs a key that no longer exists → messages is permanently lost, even though the new checkpoint still references it.
Closing this fully would mean re-persisting any missing referenced blob from the in-memory channel_values during put() (it's still available there), rather than only refreshing TTL. That's a larger change with some correctness considerations (re-writing under the exact :channel:version key, and trusting channel_values holds the carried-over value at write time), so I've kept it out of this PR.

Do you want me to handle that re-persist case in this PR, or is the TTL-refresh approach sufficient here and we track the re-persist as a follow-up?

christian-bromann: Thanks! One thing though: re-persisting in `put()` doesn't actually close the window you described. By the time node B runs, it's already loaded state via `getTuple()`, which reconstructs `messages` from that same blob and hits the identical no-op-on-expired-key problem, so if the blob expired in the gap, `channel_values` no longer holds `messages` at `put()` time and there's nothing to re-persist.

The gap is structural (blob TTL is independent of the checkpoint, `EXPIRE` can't resurrect, reads have no source to re-persist from). The real fix is converging on the `PostgresSaver` model: every `put()` upserts blobs for all channels and drops inline `channel_values`, which deserves its own PR.

So let's keep TTL-refresh here and track re-persist as a follow-up. This PR already turns unconditional loss into a narrow expire-during-idle window, and re-persist as proposed wouldn't close it anyway.

Two small asks before I approve:
1. Document the limitation (comment + changeset line): with TTL on, a carried-over blob can still be lost if it expires during an idle gap > `defaultTTL`. Don't want a new silent loss path.
2. A boundary int test: erode a blob to expiry, read, assert the channel is cleanly absent (not corrupt).

MohMaherId: @christian-bromann , both done:

**1. Documented the limitation (comment + changeset line)** — added a comment at the reconstruction skip-path and a "Known limitation" line in the changeset: with TTL on, a carried-over blob can still be lost if it expires during an idle gap > `defaultTTL`. When that happens the channel is left cleanly absent (not corrupt) on read, and re-persisting expired blobs is tracked as a follow-up.

**2. Boundary integration test** — `should drop a channel cleanly when its blob has expired`: deletes the carried-over blob, reads, and asserts the read doesn't error, the other channel is intact, and the expired channel is cleanly absent.

libs/checkpoint-redis/src/index.ts
```diff
@@ -850,6 +904,52 @@ export class RedisSaver extends BaseCheckpointSaver {
...
+      // Collected and refreshed in a single applyTTL() call so the per-key EXPIREs
+      // run concurrently and the refreshOnRead check is evaluated once, not per
+      // channel. This path also runs for list(), which reconstructs via this method.
+      if (
+        this.ttlConfig?.refreshOnRead &&
+        this.ttlConfig?.defaultTTL &&
+        usedBlobKeys.length > 0
+      ) {
+        await this.applyTTL(...usedBlobKeys);
+      }
```
christian-bromann: Can we wrap this into `Promise.all` as well?

christian-bromann (changes requested):
Changes look good! One last thing:

libs/checkpoint-redis/src/index.ts
MohMaherId: @christian-bromann, 
Done in `80beb38`, `8446a98` (formatting fix) — `loadCheckpointWithWrites()` now fetches the missing channels' blobs concurrently via `Promise.all`, then assigns them in (the expired-blob clean-skip behavior is preserved). That covers the reconstruction read path alongside the parallel blob writes already in `put()`.

christian-bromann (approved):
LGTM 👍 

Thanks a lot!

PR #2337, opened 2026-04-27
fix(checkpoint): restrict load() to allowlisted LangChain id namespaces

etairl:
## Summary

`JsonPlusSerializer._loads` walks every parsed checkpoint object and calls `@langchain/core/load` on anything matching the LangChain serialization envelope `{lc:1, type:"constructor", id:[...]}` with no allowlist of acceptable class IDs. Because graph state can contain attacker-influenced fields (a message body, a tool payload, anything that survives JSON serialization), a crafted envelope nested inside state causes `load()` to instantiate an arbitrary registered LangChain class with attacker-chosen `kwargs` on the next checkpoint restore — a classic insecure-deserialization sink usable as a stepping stone to RCE / SSRF / credential exfiltration via tool, retriever, loader, or LLM constructors with side effects.

This affects every checkpoint backend (Memory, SQLite, Postgres, Redis, MongoDB) because `JsonPlusSerializer` is the default serde set on `BaseCheckpointSaver`.

## Fix

Gate the `load()` call with a default allowlist of LangChain `id` namespace prefixes whose classes are pure data containers — no I/O, no env access, no network in their constructors:

- `["langchain_core", "messages"]` — AIMessage, HumanMessage, ToolMessage, SystemMessage, etc.
- `["langchain_core", "documents"]`
- `["langchain_core", "prompt_values"]`
- `["langchain_core", "outputs"]`

Envelopes outside the allowlist (e.g. `langchain_core/tools/*`, `langchain_core/retrievers/*`, `langchain_core/language_models/*`, `langchain/document_loaders/*`, `langchain/vectorstores/*`, etc.) pass through as plain objects rather than being instantiated. Apps that explicitly trust their checkpoint store and need to rehydrate additional namespaces can extend the allowlist via a new `JsonPlusSerializerOptions.loadableLangChainPrefixes` constructor option.

## Test plan

- [x] Existing AIMessage / HumanMessage / complex-nested-state roundtrip tests still pass (`pnpm test src/serde/tests/jsonplus.test.ts`).
- [x] New test: a `langchain_core.tools.AttackerTool` envelope nested inside graph state is preserved as a plain object on deserialization and never reaches `load()`.
- [x] New test: an unknown-namespace envelope is preserved as plain data instead of crashing the restore.
- [x] New test: opting in via `loadableLangChainPrefixes` makes `load()` reachable again, proving the gate (not the registry) is what blocks the call.
- [x] Full `@langchain/langgraph-checkpoint` package suite green (74 tests).
- [x] Package builds (`pnpm --filter @langchain/langgraph-checkpoint build`).

> Scanned automatically with https://github.com/etairl/Probus

### DIFF
--- a/.changeset/secure-jsonplus-load.md
+++ b/.changeset/secure-jsonplus-load.md
@@ -1,0 +1,5 @@
+---
+"@langchain/langgraph-checkpoint": patch
+---
+
+fix: restrict JsonPlusSerializer's load() call to an allowlist of safe LangChain id namespaces (messages, documents, prompt_values, outputs) to close an insecure-deserialization sink in checkpoint restore. Other namespaces pass through as plain objects; apps can extend the allowlist via the new `JsonPlusSerializerOptions.loadableLangChainPrefixes` option.
--- a/libs/checkpoint/src/serde/jsonplus.ts
+++ b/libs/checkpoint/src/serde/jsonplus.ts
@@ -14,6 +14,44 @@ function isLangChainSerializedObject(value: Record<string, unknown>) {
 }
 
 /**
+ * Default LangChain `id` namespace prefixes whose classes are pure data
+ * containers (no I/O, no network, no env access in their constructors) and
+ * are therefore safe to instantiate when reviving a checkpoint.
+ *
+ * Other LangChain namespaces (tools, retrievers, vectorstores, language
+ * models, document loaders, agents, callbacks, etc.) can have construction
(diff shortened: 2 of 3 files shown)


etairl:
Tracking issue: #2346 (Fixes #2346)

hntrl (commented):
See https://github.com/langchain-ai/langgraphjs/issues/2346#issuecomment-4465547550

PR #2342, opened 2026-04-28
fix(sdk): avoid reusing history message array references during submit

### LINKED ISSUE #2341: @langchain/vue useStream marks historical assistant message as   streaming after submit
### Package

  @langchain/vue 0.4.5
  @langchain/langgraph-sdk 1.8.9

  ### Description

  When `useStream` is initialized with `fetchStateHistory: true` and an
  existing `threadId`, submitting a new user message can cause historical
  messages returned from thread state/history to appear updated in the
  UI.

  In our Vue integration, the last historical assistant message becomes
  `streaming` immediately after `submit()`, even before a new assistant
  message is produced. This makes consumers observe a history message
  update when the intent is to only stream the new run.

  ### Suspected cause

  `StreamOrchestrator.submitDirect()` seeds stream values from
  `historyValues` using a shallow copy:

  ```ts
  const prev = {
    ...this.historyValues,
    ...this.stream.values
  };

  return { ...prev };

  Because historyValues[messagesKey] keeps the same message array/object
  references from fetched history until a stream event replaces them,
  consumers reading values/messages can observe historical messages as
  part of the active loading state.

  ### Expected behavior

  Calling submit() should not make historical messages appear updated or
  streaming. Only messages created or updated by the new run should
  change.

  ### Workaround

  We worked around this by snapshotting message IDs before submit() and
  only marking assistant messages as streaming when they were not present
  in the pre-submit baseline.

zhouchuan315-png:
Closes #2341

### Summary

When a new run is started, the SDK UI layer seeds stream state from history or initial values using a shallow object copy, but it keeps the same `messages` array reference.

This can make consumers observe historical messages as part of the new active stream state.

### Current behavior

During `submit()`, stream state is initialized from history/initial values.

The state object itself is copied, but `messages` still points to the same array reference from history.

As a result, consumers that derive UI state from `values/messages` can treat historical messages as if they were part of the new in-flight run.

In our case, a historical assistant message could appear updated during submit, even though no new assistant message had been produced yet.

### Expected behavior

Starting a new run should not reuse the same history `messages` array reference in stream state.

Historical messages should remain stable, and only messages produced by the new run should be reflected as stream-time updates.

### Root cause

The shared UI orchestrator initializes stream values from history with a shallow copy of the state object, but not of the `messages` array.

So the container object changes, while the `messages` array reference is still shared.

### Proposed fix

When seeding stream state from history/initial values, copy the `messages` array for the configured `messagesKey` before passing it into stream state.

A shallow array copy is enough here; the main problem is the shared array reference.

### Implementation

This patch:

- adds a small helper to copy the `messages` array for the configured `messagesKey`
- applies it in `StreamOrchestrator`
- applies the same protection in `CustomStreamOrchestrator`
- adds regression tests for both code paths

### Why this matters

Framework consumers often use `values/messages` as the source of truth for rendering streaming state.

If history and active stream state share the same `messages` array reference, UI layers can produce false positive updates on historical messages.

### DIFF
--- a/libs/sdk/src/ui/orchestrator-custom.test.ts
+++ b/libs/sdk/src/ui/orchestrator-custom.test.ts
@@ -363,6 +363,28 @@ describe("CustomStreamOrchestrator", () => {
 
       orch.dispose();
     });
+
+    it("does not reuse the initial messages array for stream values", async () => {
+      const historyMessage = { id: "m1", content: "hello", type: "human" };
+      const initialMessages = [historyMessage];
+      const transport = createMockTransport([]);
+
+      const orch = new CustomStreamOrchestrator<TestState>(
+        createOptions({
+          transport,
+          threadId: "t1",
+          initialValues: { messages: initialMessages },
+        })
+      );
+
+      await orch.submit({ messages: [] });
+
+      expect(orch.values.messages).toEqual(initialMessages);
+      expect(orch.values.messages).not.toBe(initialMessages);
+      expect(orch.values.messages[0]).toBe(historyMessage);
+
+      orch.dispose();
+    });
   });
 
   describe("subagents", () => {
--- a/libs/sdk/src/ui/orchestrator-custom.ts
+++ b/libs/sdk/src/ui/orchestrator-custom.ts
@@ -186,6 +186,15 @@ export class CustomStreamOrchestrator<
     return { ...current, [messagesKey]: messages };
   };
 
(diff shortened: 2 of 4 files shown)


christian-bromann:
Thanks for raising the PR @zhouchuan315-png 

We have published a new version of our Vue SDK that doesn't make use of the current orchestrator class anymore. We were not planning to make further updates to the v0.x Vue SDK.

Issue #2346, opened 2026-04-29
Insecure deserialization: JsonPlusSerializer calls @langchain/core load() on attacker-influenceable checkpoint state


etairl:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangGraph.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangGraph.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangGraph (or the specific integration package).

### Example Code

Reproducing the unsafe code path with the in-memory checkpointer (every other backend goes through the same default `JsonPlusSerializer`):

```typescript
import { MemorySaver } from "@langchain/langgraph-checkpoint";
import { v4 as uuid } from "uuid";

const saver = new MemorySaver();
const config = { configurable: { thread_id: uuid(), checkpoint_ns: "" } };

// Simulate untrusted input flowing into graph state. Anything that survives
// JSON serialization on the way *into* the checkpoint store reaches the
// reviver on the way *out*.
const attackerInfluencedState = {
  user_field: {
    lc: 1,
    type: "constructor",
    id: ["langchain_core", "messages", "AIMessage"], // any registered class
    kwargs: { content: "rehydrated as an instance, not plain data" },
  },
};

const checkpoint = {
  v: 4,
  id: uuid(),
  ts: new Date().toISOString(),
  channel_values: { state: attackerInfluencedState },
  channel_versions: {},
  versions_seen: {},
};

await saver.put(config, checkpoint as any, { source: "input", step: 0, parents: {} } as any, {});
const tuple = await saver.getTuple(config);

// On restore, JsonPlusSerializer._loads' reviver matches the {lc:1, ...}
// envelope and calls @langchain/core/load(), instantiating the referenced
// class with attacker-controlled kwargs. With a class that has constructor
// side effects (tools, retrievers, loaders, language models), this is an
// insecure-deserialization sink.
console.log(tuple?.checkpoint.channel_values.state.user_field.constructor.name);
// → "AIMessage" (instance) instead of plain Object
```

The vulnerable line is `libs/checkpoint/src/serde/jsonplus.ts` ~L78:

```ts
} else if (isLangChainSerializedObject(revivedObj)) {
  return load(JSON.stringify(revivedObj));   // no allowlist on revivedObj.id
}
```

## Summary

[...]

PR #2349, opened 2026-04-30
fix(langgraph-checkpoint-mongodb): block NoSQL operator injection via top-level identifiers

Nagendhra-web:
## Summary

Closes a NoSQL operator-injection sink (CWE-943) in `MongoDBSaver.{getTuple, list, put, putWrites, deleteThread}`. Without this guard, a caller able to shape `thread_id`, `checkpoint_ns`, `checkpoint_id`, or `task_id` (multi-tenant SDK deployments where the `RunnableConfig` originates from request input, or webhook payloads that flow into a persisted thread) can pass an object such as `{ $ne: null }` and have MongoDB return, overwrite, or delete checkpoints belonging to other tenants.

The audit posted in #2346 (cc @etairl) listed *NoSQL injection in `MongoDBSaver.{getTuple, list}`* among the unfiled findings from the same security-review pass that produced #2337. This PR confirms the finding, extends its scope to the three additional sinks the audit did not enumerate (`put`, `putWrites`, `deleteThread`), and provides the fix.

## Vulnerable sinks

| Method | Fields embedded directly into the query |
|---|---|
| `getTuple` | `thread_id`, `checkpoint_ns`, `checkpoint_id` |
| `list` | `thread_id`, `checkpoint_ns`, `before.configurable.checkpoint_id` |
| `put` | `thread_id`, `checkpoint_ns`, `checkpoint_id`, parent checkpoint id |
| `putWrites` | `thread_id`, `checkpoint_ns`, `checkpoint_id`, `taskId` |
| `deleteThread` | `thread_id` |

The pre-existing `metadata.*` filter guard in `list` already enforces primitive-only values for one query position. This PR extends the same defense to the top-level identifier keys that drive every primary-key query.

## Proof of concept

```ts
import { MongoClient } from "mongodb";
import { MongoDBSaver } from "@langchain/langgraph-checkpoint-mongodb";

const client = new MongoClient(process.env.MONGO_URL!);
await client.connect();
const saver = new MongoDBSaver({ client });

await saver.put(
  { configurable: { thread_id: "tenant-a", checkpoint_ns: "" } },
  { id: "cp-1", v: 4, ts: new Date().toISOString(),
    channel_values: { secret: "tenant-a-only" },
    channel_versions: {}, versions_seen: {} } as any,
  { source: "input", step: 0, parents: {} } as any,
  {}
);

const stolen = await saver.getTuple({
  configurable: {
    thread_id: { \$ne: null } as any, // operator injection
    checkpoint_ns: "",
  },
});
console.log(stolen?.checkpoint.channel_values.secret); // "tenant-a-only"
```

`{ \$ne: null }`, `{ \$gt: \"\" }`, `{ \$regex: \".*\" }`, and `{ \$exists: true }` all reach MongoDB unmodified. `deleteThread({ \$ne: null } as any)` deletes every checkpoint and every pending write in the database.

## Severity

Proposed CVSS v3.1 of 7.7 (High): \`AV:N/AC:L/PR:L/UI:N/S:U/C:H/I:H/A:H\`. Network-reachable, low-complexity, requires only the privilege the SDK already grants to a caller, no user interaction, scope unchanged, high impact across confidentiality (cross-tenant disclosure), integrity (cross-tenant overwrite), and availability (wholesale deletion).

## Fix

A single private static \`assertSafeIdentifier\` guard on \`MongoDBSaver\`, mirroring the existing primitive-only enforcement applied to \`metadata.*\` filter keys, applied at every query-building site (15 call sites across 5 methods).

\`\`\`ts
private static assertSafeIdentifier(
  field: string,
  value: unknown
): asserts value is string | undefined {
  if (value === undefined) return;
  if (typeof value !== \"string\") {
    const observed =
      value === null ? \"null\"
      : Array.isArray(value) ? \"array\"
      : typeof value;
    throw new Error(
      \`Invalid configurable value for key \"\${field}\": expected a string identifier (got \${observed}). This guard protects the MongoDB query from NoSQL operator injection.\`
    );
  }
}
\`\`\`

The guard is intentionally a TypeScript \`asserts\` predicate so call-sites get type narrowing for free and the compiler enforces that no later code path uses an unvalidated identifier.

## Why this design (and not the alternatives)

* Not query rewriting. Coercing operator-shaped values into strings (\`String(value)\`) would silently shift saver behavior for legitimate non-string callers and would also legitimize the attack surface. Rejecting at the boundary surfaces the type error to the caller instead.
* Not driver-level escaping. The MongoDB Node driver intentionally treats objects as operator expressions. There is no safe-string mode to opt into per-call. Validation must happen above the driver.
* Not a Zod or runtime-schema dependency. A short predicate has the same correctness with no new dependency, no bundle bloat for downstream consumers, and matches the existing \`Object.entries(filter).forEach(...)\` style.
* Not a per-method guard. Centralising in one static method makes it impossible for a future call-site to forget validation, and the \`asserts\` annotation surfaces missed sites at compile time.

## Test plan

* [x] 16 new tests in \`libs/checkpoint-mongodb/src/tests/checkpoints.test.ts\` under \`describe(\"identifier validation (NoSQL injection guard)\")\`, covering every method by every controllable identifier by four shapes of bad input (operator object, plain object, array, number, null), plus a happy-path roundtrip and a regression test for the pre-existing \`filter\` guard.
* [x] Existing test suite green (\`pnpm --filter @langchain/langgraph-checkpoint-mongodb test\`, 43 of 43).
* [x] Build, attw, and publint pass (\`pnpm --filter @langchain/langgraph-checkpoint-mongodb build\`).
* [x] No new dependencies, no public API changes, no behavior change for valid string identifiers.

## Related and follow-ups

The audit in #2346 lists six further unfiled findings of the same shape across other checkpoint backends:

* Redis key/glob injection in \`RedisSaver\` and \`ShallowRedisSaver\`
* RediSearch query injection in \`RedisStore\`
* \`__proto__\` prototype pollution in \`MemorySaver\`
* namespace-colon collisions in the in-memory \`Store\`
* SSE event/data injection via unsanitized namespace in \`pregel/stream\`
* \`maxConcurrency\` bypass via dynamic \`call()\`

Each warrants its own scoped PR. Happy to follow up with the same guard pattern (a shared \`assertSafeIdentifier\` exported from \`@langchain/langgraph-checkpoint\`) if the maintainers prefer to consolidate.

## Disclosure

Original finding credited to @etairl (audit posted in #2346). This PR was prepared for coordinated public disclosure since the audit list is already public. Happy to coordinate timing with a private GHSA if the maintainers prefer.

### DIFF
--- a/.changeset/secure-mongodb-checkpoint-injection.md
+++ b/.changeset/secure-mongodb-checkpoint-injection.md
@@ -1,0 +1,21 @@
+---
+"@langchain/langgraph-checkpoint-mongodb": patch
+---
+
+fix(checkpoint-mongodb): block NoSQL operator injection via top-level identifiers
+
+`MongoDBSaver.{getTuple,list,put,putWrites,deleteThread}` previously embedded
+`thread_id`, `checkpoint_ns`, `checkpoint_id`, and `task_id` directly into
+MongoDB queries with no type validation. A caller that can shape those
+fields, for example a multi-tenant SDK deployment where the
+`RunnableConfig` originates from request input, or a webhook body that
+flows into a persisted thread, could pass an object such as `{ $ne: null }`
+and have MongoDB return or overwrite checkpoints belonging to other
+tenants (CWE-943).
+
+This patch adds a single static `assertSafeIdentifier` guard that mirrors
+the existing primitive-only enforcement applied to `metadata.*` filter
+keys, and applies it at every query-building site. Behaviour for valid
+string identifiers is unchanged. Invalid types now raise a precise error
+with the field name, observed type, and a pointer to the security
+rationale.
--- a/libs/checkpoint-mongodb/src/checkpoint.ts
+++ b/libs/checkpoint-mongodb/src/checkpoint.ts
@@ -43,6 +43,49 @@ export class MongoDBSaver extends BaseCheckpointSaver {
       : {};
   }
 
+  /**
+   * Asserts that a value sourced from {@link RunnableConfig.configurable} (or
+   * any other caller-influenced position) is safe to embed directly in a
+   * MongoDB query: it must be a primitive string and not, for instance, an
+   * object such as `{ $ne: null }` that the driver would interpret as a
+   * query operator.
+   *
+   * Without this guard a caller that can shape `thread_id`, `checkpoint_ns`,
+   * `checkpoint_id`, or `taskId` (multi-tenant SDK deployments where the
+   * config originates from request input, webhook bodies that flow into a
+   * persisted thread, etc.) can promote a string field into an operator
+   * expression and cause MongoDB to return or overwrite checkpoints
+   * belonging to other tenants. CWE-943: Improper Neutralization of Special
+   * Elements in Data Query Logic (NoSQL Injection).
+   *
+   * Mirrors the existing primitive-only enforcement applied to `filter`
+   * values in {@link list} (added to defend the `metadata.*` query keys),
+   * and extends it to the top-level identifier keys those callers also
+   * control.
+   *
+   * @param field Name of the configurable field, used in the error message.
+   * @param value Value to validate. `undefined` is accepted so this guard
+   *              composes with the existing required-field checks each
+   *              method already performs.
+   */
+  private static assertSafeIdentifier(
+    field: string,
+    value: unknown
+  ): asserts value is string | undefined {
+    if (value === undefined) return;
+    if (typeof value !== "string") {
+      const observed =
+        value === null
+          ? "null"
+          : Array.isArray(value)
+          ? "array"
+          : typeof value;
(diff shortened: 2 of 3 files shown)


Nagendhra-web:
Fixes: GHSA-98xf-r82g-9mhx

Coordinated disclosure filed under the langgraphjs Security Advisories at https://github.com/langchain-ai/langgraphjs/security/advisories/GHSA-98xf-r82g-9mhx (currently in triage). CVSS 3.1 score: 8.8 (High), vector `CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:U/C:H/I:H/A:H`, CWE-943 (NoSQL Injection).

hntrl:
Closing in favor of #2397

PR #2353, opened 2026-04-30
fix(langgraph-checkpoint-mongodb): reject non-primitive thread_id/checkpoint_ns/checkpoint_id

### LINKED ISSUE #2351: Security: NoSQL injection in MongoDBSaver.getTuple permits cross-tenant checkpoint disclosure
### Checked other resources

### Example Code

```typescript
// Minimal reproducer — assumes a benign tenant has previously written a
// checkpoint under thread_id "victim-thread", and the application accepts
// a `thread_id` from a JSON request body without string-coercion / validation.

import { MongoClient } from "mongodb";
import { MongoDBSaver } from "@langchain/langgraph-checkpoint-mongodb";

const saver = new MongoDBSaver({
  client: new MongoClient(process.env.MONGO_URI!),
  dbName: "checkpoints",
});

// What an attacker would send as the JSON body of a typical /chat-style endpoint:
const attackerBody = {
  thread_id: { $gt: "" },
  checkpoint_ns: { $ne: null },
};

// What the (vulnerable but otherwise normal) server forwards into config:
const config = { configurable: attackerBody };

const tuple = await saver.getTuple(config);

// `tuple.checkpoint` is the latest checkpoint in the entire collection,
// even though the attacker is not authorized for that thread.
console.log(tuple?.checkpoint);
console.log(tuple?.metadata); // victim's metadata
console.log(tuple?.pendingWrites); // victim's pending writes
```

_No response_

## Summary

`MongoDBSaver.getTuple` (`libs/checkpoint-mongodb/src/checkpoint.ts:75-145`) forwards `thread_id`, `checkpoint_ns`, and `checkpoint_id` from `config.configurable` straight into a MongoDB `find()` call with **no type validation**. A caller who can shape those values — e.g. by passing an HTTP request body's `thread_id` into config without coercion — can submit Mongo operator objects (`{ $gt: "" }`, `{ $ne: null }`) to bypass thread scoping and read other tenants' checkpoints.

The same file already defends against this exact attack class inside `list()` (lines 170–180):

```ts
if (filter) {
  Object.entries(filter).forEach(([key, value]) => {
    // Prevent MongoDB operator injection - only allow primitive values
    if (value !== null && typeof value === "object") {
      throw new Error(
        `Invalid filter value for key "${key}": filter values must be primitives ...`
      );
    }
    query[`metadata.${key}`] = value;
  });
}
```

…so the maintainers clearly understand the risk class. `getTuple` and the second `find(configurableValues)` lookup in checkpoint_writes (line 112) just lack the analogous guard.

## Vulnerable code

```ts
// libs/checkpoint-mongodb/src/checkpoint.ts:75-96
async getTuple(config: RunnableConfig): Promise<CheckpointTuple | undefined> {
  const { thread_id, checkpoint_ns = "", checkpoint_id } = config.configurable ?? {};
  let query;
  if (checkpoint_id) {
    query = { thread_id, checkpoint_ns, checkpoint_id };
  } else {
    query = { thread_id, checkpoint_ns };
  }
  const result = await this.db
    .collection(this.checkpointCollectionName)
    .find(query)
    .sort("checkpoint_id", -1)
    .limit(1)
    .toArray();
  ...
}
```

## Attack path

1. Web service uses `MongoDBSaver` and sources `thread_id` (or the entire `configurable`) from a JSON request body — a common pattern for DIY backends that are not running on top of LangGraph Platform's URL-based routing. Express's default `qs` query parser is also vulnerable: `?thread_id[$gt]=` parses into `{ $gt: "" }`.
2. Attacker sends:
   ```json
   { "thread_id": { "$gt": "" }, "checkpoint_ns": { "$ne": null } }
   ```
3. Server constructs `config.configurable = req.body` (or merges fields without `String(...)` / `z.string()` coercion) and either calls `saver.getTuple(config)` directly **or** invokes the graph — `app.invoke(input, config)` calls `getTuple` internally when a saver is configured.
4. Query becomes `find({ thread_id: { $gt: "" }, checkpoint_ns: { $ne: null } })` → matches every checkpoint, sorted descending → returns the latest checkpoint in the entire collection.
5. Returned tuple's `checkpoint`, `metadata`, and `pendingWrites` are the victim's data, fully deserialized. The follow-up `find(configurableValues)` at line 112 also exfiltrates the victim's pending writes.

## Impact

- **Cross-tenant disclosure** of checkpoints: conversation state, message history, tool call inputs/outputs, custom metadata.
- **Horizontal privilege escalation** when `thread_id` is the application's tenancy boundary.
- Triggered through any code path that calls `app.invoke()` / `app.stream()` with a configured saver — not just direct `getTuple` callers.

## Not exploitable when

- `thread_id` is server-issued (session/JWT-derived) — already a string the attacker cannot make into an object.
- `thread_id` comes from a URL path param (e.g. `/threads/:thread_id`) — always a string.
- App validates with `typeof thread_id === "string"` / Zod / Joi.

## Recommended fix

Mirror the guard already present in `list()`:

```ts
const { thread_id, checkpoint_ns = "", checkpoint_id } = config.configurable ?? {};
for (const [name, v] of Object.entries({ thread_id, checkpoint_ns, checkpoint_id })) {
  if (v !== undefined && (typeof v === "object" || typeof v === "function")) {
    throw new Error(`Invalid configurable.${name}: must be a primitive`);
  }
}
```

A stricter version requires `typeof v === "string"`. The same guard must also apply before the `find(configurableValues)` lookup at line 112, since `configurableValues` inherits the unvalidated `thread_id` / `checkpoint_ns`.

## Notes

- I'm aware the bug-report template directs security issues to the private security policy. I'm filing this publicly per maintainer guidance on a related (public) report (#2346 / PR #2337). Happy to move to the private flow if preferred.
- A sister class of injection issues exists in `@langchain/langgraph-checkpoint-redis` (Redis key/glob/RediSearch query injection) — separate report on the way.

### System Info

- Package: `@langchain/langgraph-checkpoint-mongodb` (current `main`, all published versions)
- File: `libs/checkpoint-mongodb/src/checkpoint.ts:75-145`
- Platform: N/A — vulnerability is platform-independent and reproduces against any MongoDB backend reachable by `MongoDBSaver`.
- Node: any supported version.

etairl:
## Description

`MongoDBSaver.getTuple`, `list`, `put`, and `putWrites` forwarded `thread_id`, `checkpoint_ns`, and `checkpoint_id` from `config.configurable` straight into MongoDB queries with no type validation. A caller able to shape those values (for example, by passing an HTTP request body's `thread_id` straight into config without coercion) could submit Mongo operator objects (`{ $gt: "" }`, `{ $ne: null }`) to bypass thread scoping and read other tenants' checkpoints.

`getTuple` was the most exposed entry point because it's invoked automatically by `app.invoke()` whenever a saver is configured, so any chat handler that takes `thread_id` from a request body without `String(...)` / `z.string()` coercion was reachable.

## Fix

Adds a private `MongoDBSaver.assertConfigurableScalars` helper that rejects object/function values for these fields, mirroring the existing guard in `list()` for the `filter` argument:

```ts
private static assertConfigurableScalars(values: {
  thread_id?: unknown;
  checkpoint_ns?: unknown;
  checkpoint_id?: unknown;
}): void {
  for (const [name, value] of Object.entries(values)) {
    if (value === undefined || value === null) continue;
    if (typeof value === "object" || typeof value === "function") {
      throw new Error(
        `Invalid configurable.${name}: must be a primitive (string, number, or boolean), got ${typeof value}.`
      );
    }
  }
}
```

It is called at the top of every method that consumes those fields:

- `getTuple` — guards both the initial `find()` and the follow-up `find(configurableValues)` against `checkpoint_writes`.
- `list` — also guards `before.configurable.checkpoint_id` (which previously fed `{ $lt: ... }` unchecked).
- `put` and `putWrites` — guard the upsert filters.

## Tests

Adds 7 unit tests in `libs/checkpoint-mongodb/src/tests/checkpoints.test.ts` covering each entry point with a Mongo-operator payload, plus a happy-path assertion that primitive values still pass. All 13 unit tests in the suite pass:

```
✓ |unit| src/tests/checkpoints.test.ts (13 tests)
```

## Backward compatibility

Strictly additive validation — primitive `thread_id` / `checkpoint_ns` / `checkpoint_id` (the only documented and intended shape) keep working. Callers that were relying on undocumented object-shaped identifiers will get a clear error pointing at the offending field.

## Notes

- Same class of guard is already present (and tested) in `list()` for the `filter` argument; this PR extends it to the configurable identifiers.
- Sister injection issues likely exist in `@langchain/langgraph-checkpoint-redis` (key/glob/RediSearch query injection); separate report on the way.

Fixes #2351

### DIFF
--- a/.changeset/mongodb-saver-operator-injection.md
+++ b/.changeset/mongodb-saver-operator-injection.md
@@ -1,0 +1,5 @@
+---
+"@langchain/langgraph-checkpoint-mongodb": patch
+---
+
+fix: validate `thread_id`, `checkpoint_ns`, and `checkpoint_id` in `MongoDBSaver` to prevent NoSQL operator injection. Object/function values for these fields are now rejected before they reach a MongoDB query, mirroring the existing guard on the `filter` argument of `list()`. Applies to `getTuple`, `list`, `put`, and `putWrites`.
--- a/libs/checkpoint-mongodb/src/checkpoint.ts
+++ b/libs/checkpoint-mongodb/src/checkpoint.ts
@@ -43,6 +43,29 @@ export class MongoDBSaver extends BaseCheckpointSaver {
       : {};
   }
 
+  /**
+   * Reject non-primitive values for `thread_id`, `checkpoint_ns`, and
+   * `checkpoint_id` before they reach a MongoDB query. Mongo treats nested
+   * objects as operator expressions (e.g. `{ $gt: "" }`), so allowing
+   * caller-shaped objects here would let an attacker bypass thread scoping
+   * and read other tenants' checkpoints. Mirrors the existing guard in
+   * `list()` for the `filter` argument.
+   */
+  private static assertConfigurableScalars(values: {
+    thread_id?: unknown;
+    checkpoint_ns?: unknown;
+    checkpoint_id?: unknown;
+  }): void {
+    for (const [name, value] of Object.entries(values)) {
+      if (value === undefined || value === null) continue;
+      if (typeof value === "object" || typeof value === "function") {
+        throw new Error(
+          `Invalid configurable.${name}: must be a primitive (string, number, or boolean), got ${typeof value}.`
+        );
+      }
+    }
+  }
+
   constructor(
     {
       client,
@@ -78,6 +101,11 @@ export class MongoDBSaver extends BaseCheckpointSaver {
       checkpoint_ns = "",
       checkpoint_id,
     } = config.configurable ?? {};
+    MongoDBSaver.assertConfigurableScalars({
+      thread_id,
+      checkpoint_ns,
+      checkpoint_id,
+    });
     let query;
     if (checkpoint_id) {
       query = {
@@ -154,6 +182,11 @@ export class MongoDBSaver extends BaseCheckpointSaver {
     options?: CheckpointListOptions
   ): AsyncGenerator<CheckpointTuple> {
     const { limit, before, filter } = options ?? {};
+    MongoDBSaver.assertConfigurableScalars({
+      thread_id: config?.configurable?.thread_id,
+      checkpoint_ns: config?.configurable?.checkpoint_ns,
+      checkpoint_id: before?.configurable?.checkpoint_id,
+    });
     const query: Record<string, unknown> = {};
 
     if (config?.configurable?.thread_id) {
@@ -242,6 +275,11 @@ export class MongoDBSaver extends BaseCheckpointSaver {
         `The provided config must contain a configurable field with a "thread_id" field.`
       );
     }
+    MongoDBSaver.assertConfigurableScalars({
+      thread_id,
+      checkpoint_ns,
+      checkpoint_id: config.configurable?.checkpoint_id,
+    });
     const [
       [checkpointType, serializedCheckpoint],
       [metadataType, serializedMetadata],
@@ -301,6 +339,11 @@ export class MongoDBSaver extends BaseCheckpointSaver {
         `The provided config must contain a configurable field with "thread_id", "checkpoint_ns" and "checkpoint_id" fields.`
       );
     }
+    MongoDBSaver.assertConfigurableScalars({
+      thread_id,
+      checkpoint_ns,
+      checkpoint_id,
+    });
 
     const operations = await Promise.all(
       writes.map(async ([channel, value], idx) => {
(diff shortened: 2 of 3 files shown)


hntrl (commented):
Closing in favor of https://github.com/langchain-ai/langgraphjs/pull/2353. Will attach you as a co-author on the commit + advisory, but generally I'm not a fan of adding internal helpers as class properties

PR #2363, opened 2026-05-04
feat(sdk): add returnMinimal to threads update

cwlbraa:
Adds a first-class `returnMinimal` option to JS SDK `threads.update`. When enabled, the SDK sends `Prefer: return=minimal` and resolves with `undefined` for the 204 response. Includes a patch changeset and focused fetch test coverage.

### DIFF
--- a/libs/sdk/src/client.ts
+++ b/libs/sdk/src/client.ts
@@ -1174,7 +1174,7 @@ export class ThreadsClient<
         ? { ttl: payload.ttl, strategy: "delete" as const }
         : payload?.ttl;
 
-    return this.fetch<Thread>(`/threads/${threadId}`, {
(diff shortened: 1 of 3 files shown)


libs/sdk/src/client.ts
```diff
@@ -1090,19 +1090,95 @@ export class ThreadsClient<
...
+       */
+      signal?: AbortSignal;
+    }
+  ): Promise<Thread | void> {
     const ttlPayload =
       typeof payload?.ttl === "number"
         ? { ttl: payload.ttl, strategy: "delete" as const }
         : payload?.ttl;
 
     return this.fetch<Thread>(`/threads/${threadId}`, {
```
pjrule: nit: should this be `this.fetch<Thread | void>`?

pjrule (dismissed):

PR #2380, opened 2026-05-11
fix(sdk): re-encode finalized tool_call blocks as chunks during streaming

nick-hollon-lc:
When a streaming AI message has both finalized and still-streaming
tool calls — common with parallel tool calls, where blocks land
sequentially within a single message — the langchain core
\`AIMessageChunk\` constructor drops the caller-supplied \`tool_calls\`
field and rebuilds it from \`tool_call_chunks\` via
\`collapseToolCallChunks\`. Without re-encoding, every finished tool call
disappears from \`message.tool_calls\` until \`message-finish\` flips the
message to a plain \`AIMessage\`.

UI symptom: only one subagent card visible during streaming, then four
at the end (or the equivalent for any parallel-tool-call UX).

Fix in \`extractToolCallChunks\`: when a finalized \`tool_call\` block is
encountered, re-emit it as a \`tool_call_chunk\` with args
JSON-stringified back to a string, so the constructor's chunk-collapse
path includes it from the moment it arrives.

## Release Note

None

### DIFF
--- a/libs/sdk/src/stream/assembled-to-message.test.ts
+++ b/libs/sdk/src/stream/assembled-to-message.test.ts
@@ -1,0 +1,186 @@
+import { describe, expect, it, vi } from "vitest";
+import { AIMessage, AIMessageChunk } from "@langchain/core/messages";
+import type { ContentBlock } from "@langchain/protocol";
+import { assembledToBaseMessage } from "./assembled-to-message.js";
+
+describe("assembledToBaseMessage — tool_call / tool_call_chunk handling", () => {
+  // These tests pin the WORKAROUND for langchain-core's
+  // ``AIMessageChunk`` constructor, which discards caller-supplied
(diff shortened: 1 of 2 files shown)


nick-hollon-lc:
Closing — verified that smith-frontend doesn't actually need this re-encoding workaround. The bug described in the original commit message ("only one Subagent Task card during streaming, then four at the end") doesn't reproduce in practice with parallel `task` dispatches. Smith-frontend's per-message subagent rendering reads from `message.tool_calls` and still receives finalized tool calls correctly during streaming. The workaround was a real fix for pure-langchain consumers who only read `message.tool_calls` and were affected by `AIMessageChunk.collapseToolCallChunks` dropping finalized blocks — but with no in-tree consumer hitting that path, the simplest move is to drop it rather than maintain a workaround for a hypothetical caller.

PR #2381, opened 2026-05-11
fix(sdk): subscription plumbing — dedicated SSE streams, run.start ordering, WS pre-register

nick-hollon-lc:
Three subscription-lifecycle correctness fixes; all share
\`client/stream/index.ts\` and are thematically the same surface.

### 1. \`subscribeDedicated\` + dedicated SSE event stream for narrow projections

Selector hook consumers (\`useMessages\`, \`useToolCalls\`) scoped to a
subagent namespace would otherwise widen the shared content pump's
\`(channels, namespaces)\` union and pull every channel for every active
namespace onto the wire just to satisfy a single narrow consumer. SSE
transports get a fresh \`openEventStream\`; WS falls back to the shared
command stream (no union-widening cost on WS since all events flow on
one connection).

### 2. \`run.start\` lifecycle ordering

The lifecycle watcher and the values projection both open subscriptions
on \`/threads/{id}/stream/events\`; if either lands before \`run.start\`
commits the thread server-side, the server emits a
\`404: Thread not found\` protocol error and the client waits forever
for terminal events. Reorder so \`run.start\` awaits first, then
watchers attach. Also adds \`config\` + \`metadata\` to \`respondInput\`
signature so resume submits can carry per-run configuration overrides
(the v2 server already forwards both).

### 3. Pre-register subscription under a placeholder id before sending \`subscription.subscribe\`

The WebSocket server replays buffered events through
\`install_subscription_with_replay\` before returning the success
response that carries the \`subscription_id\` — without pre-registration,
those replayed events arrive at \`#handleIncoming\` while no matching
subscription is registered and never reach the iterator. UI symptom:
\`useMessages\` renders empty after click-to-expand on WebSocket.

## Release Note

None


libs/sdk/src/client/stream/index.ts
```diff
@@ -740,6 +767,49 @@ export class ThreadStream<
...
+      resolveGate = resolve;
+      rejectGate = reject;
+    });
+    this.#runStartReady = gate;
+    // Swallow unhandled rejection: gate-awaiters that surface the
+    // rejection via their own error path don't need a second log line,
+    // and consumers that never await it (the typical case once
+    // `run.start` resolves successfully) shouldn't trigger
+    // `unhandledrejection` either.
+    gate.catch(() => undefined);
```
hntrl, line 793: JS promises are declarative, this technically doesn't catch anything. Think you could just collapse this into not propagating `rejectGate`

hntrl: disregard, I just realized what the start gate was doing

hntrl (dismissed):

hntrl (approved):

PR #2408, opened 2026-05-16
fix(langgraph-checkpoint-sqlite): store channel deltas

### LINKED ISSUE #594: `SqliteSaver` ignores `newVersions` argument on `put` - not storing deltas

pragnyanramtha:
## Summary

- respect `newVersions` in `SqliteSaver.put` so SQLite checkpoint rows only store changed `channel_values`
- keep legacy full checkpoint writes when callers omit `newVersions`
- enable the existing checkpoint-validation delta storage conformance test for SQLite
- add a patch changeset for `@langchain/langgraph-checkpoint-sqlite`

Fixes #594.

## Validation

- `pnpm --filter @langchain/langgraph-checkpoint build:internal`
- `pnpm --filter @langchain/langgraph-checkpoint-sqlite build:internal`
- `pnpm --filter @langchain/langgraph-checkpoint-sqlite test`
- `pnpm --filter @langchain/langgraph-checkpoint-validation exec vitest run src/tests/sqlite.spec.ts`
- `pnpm exec oxfmt --check libs/checkpoint-sqlite/src/index.ts libs/checkpoint-validation/src/spec/put.ts .changeset/swift-sqlite-deltas.md`
- `pnpm exec oxlint libs/checkpoint-sqlite/src/index.ts libs/checkpoint-validation/src/spec/put.ts`
- `git diff --check`


pragnyanramtha:
After the base refresh, the only red check I see is `Browser Tests / sdk-vue`.

I checked the job log: the Vue test suite itself passed (`103 passed`), then Vitest reported an unhandled browser infrastructure error: `Failed to connect to the browser session ... [chrome] within the timeout.` The same browser suite passed on the adjacent refreshed LangGraphJS PRs, so this looks like a flaky Chrome/session startup failure rather than a SQLite checkpoint regression.

I tried to rerun the failed job/run, but GitHub rejected reruns without repository admin rights. The branch is current with `main`; all other checks are passing.

christian-bromann (changes requested):
`SqliteSaver` serializes the whole checkpoint (including `channel_values`) into the single `checkpoint` BLOB and `getTuple`/`list` return it verbatim with no parent merge. Filtering `channel_values` by `newVersions` at write time, without a version-keyed blob table to reconstruct from on read, means reading checkpoint *N* returns only the channels that changed in step *N* — every unchanged channel from prior steps is lost.

This breaks real graph state: `channelsFromCheckpoint` rebuilds regular (non-Delta) channels purely from `checkpoint.channel_values` (`emptyChannels`), so a dropped channel becomes `fromCheckpoint(undefined)` → empty channel on resume/time-travel/`getState`.

The correct pattern is already in `checkpoint-postgres`: `_dumpCheckpoint` strips `channel_values`, blobs are stored keyed by `(thread_id, ns, channel, version)`, and `_loadCheckpoint`/`_loadBlobs` reconstruct the *full* `channel_values` on read via `channel_versions`. SQLite has none of this.

CI is green because the un-skipped test only stores independent checkpoints and asserts each one's own delta — it never chains checkpoints and reads a later one back to verify full reconstructed state.

Suggested fix: add a version-keyed `checkpoint_blobs` table, strip `channel_values` from the checkpoint blob, upsert changed channels in `put`, reconstruct in `getTuple`/`list` from `channel_versions`, and handle migration of existing inline-`channel_values` rows. Also extend the conformance test to chain checkpoints and assert the full state on read. As a stopgap, revert to storing full `channel_values` and keep the test skipped rather than ship lossy persistence.

(Minor: `newVersions?` is optional here but required in the base class / Postgres.)

Issue #1142, opened 2025-04-30
Send API serialize/deserialize issue


brettshollenberger:
Hey all, just noticed an issue with serializing/deserializing objects that are wrapped with some custom class when using the `Send` API, for example `BaseMessage`, `HumanMessage`, etc. 

* For contrast, return a `BaseMessage` from a regular node, and it will stay wrapped as a `BaseMessage`
* But if you queue the state to separate workers using `Send`, the workers will receive an unwrapped (plain object) with the data. 
* This is easy enough to re-wrap as `BaseMessage`, but it causes some unexpected interactions with things like `createReactAgent`, which expect the `MessageState` interface to be applied (which fails when not wrapped as `BaseMessage`).

Minimal example:

```typescript
export interface GraphState {
    userRequest: BaseMessage; 
    chatHistory: BaseMessage[];
    task: Task;
    queue: Task[];
    completedTasks: Task[];
}

const queueTasks = async(state: GraphState) => {
    // in here, state.userRequest and state.chatHistory are fine
    if (!state.queue) {
        return [];
    }
    const queue = state.queue;
    return queue.map(task => new Send("executeTasks", { ...state, task }));
}

const planTasks = async(state: GraphState): Promise<Partial<GraphState>> => {
  //   in here, state.userRequest and state.chatHistory are fine  
  return { queue: tasks }
}

const executeTask = async(state: GraphState): Promise<Partial<GraphState>> => {
  // in here, state.userRequest and state.chatHistory are unwrapped/plain objects
  const agent =  createReactAgent(... stuff);
    const agentState = {
    messages: [
      ...(state.chatHistory.map((message) => { return new BaseMessage(message); },
      new HumanMessage(state.userRequest.content) // If we don't do this, createReactAgent will complain
    ]
  }
  agent.invoke(agentState); // This would explode if we don't re-wrap our messages after Send API
}

export const graph = new StateGraph(GraphAnnotation, ConfigurationAnnotation)
    .addNode("planTasks", planTasks)
    .addNode("executeTask", executeTask)
    .addEdge(START, "planTasks")
    .addConditionalEdges("planTasks", queueTasks)
    .addEdge("executeTasks", END)
```

Just logging the issue, thank you!

brettshollenberger:
This seems to be related, but when `Send`ing arrays of messages, I can't get them to serialize properly. When sending a single message, I can send successfully and then unwrap, but when sending `messages: []`, it blows up when trying to send. 

```tsx
[...]

PR #2409, opened 2026-05-17
fix: preserve non-plain Send args

pragnyanramtha:
## Summary
- preserve non-plain object instances while deserializing `Send`/`Command` argument trees
- keep recursive reconstruction for arrays, plain object records, serialized `Command`, and serialized `Send` payloads
- add a regression test covering `Set`, `Map`, `Date`, and a custom class instance passed through `Send`

Fixes part of #1142.

## Test plan
- `pnpm install --frozen-lockfile`
- `pnpm --filter @langchain/langgraph exec vitest run src/tests/constants.test.ts --testNamePattern "preserves non-plain objects"`
- `pnpm --filter @langchain/langgraph exec vitest run src/tests/constants.test.ts`
- `pnpm exec oxfmt --check libs/langgraph-core/src/constants.ts libs/langgraph-core/src/tests/constants.test.ts`
- `pnpm exec oxlint libs/langgraph-core/src/constants.ts libs/langgraph-core/src/tests/constants.test.ts`
- `git diff --check`
- `pnpm --filter @langchain/langgraph build`


christian-bromann (approved):
LGTM 👍

PR #2434, opened 2026-05-26
fix(sdk): avoid eager stream getter evaluation

hntrl:
## Summary

fix(sdk): avoid eager stream getter evaluation during spread

This fixes a React `useStream` development-mode failure where passing the stream handle through components that clone or rest-spread props could accidentally read lazy getters. The guarded `history` getter still throws when explicitly accessed with `fetchStateHistory: false`, but object spread no longer trips that path or widens `streamMode` by touching optional accessors.

## Changes

`@langchain/langgraph-sdk`

- Marks optional `useStream` accessors (`history`, `experimental_branchTree`, `toolProgress`, `subagents`, `activeSubagents`) as non-enumerable on the returned stream handle.
- Preserves explicit access behavior for those accessors, including the existing `history` guard and stream mode opt-in for `toolProgress`/`subagents`.
- Adds React hook regression coverage for object spread, explicit getter access, and stream mode inference.

### DIFF
--- a/.changeset/silent-stream-getters.md
+++ b/.changeset/silent-stream-getters.md
@@ -1,0 +1,7 @@
+---
+"@langchain/langgraph-sdk": patch
+---
+
+fix(react): avoid eager stream getter evaluation during object spread
+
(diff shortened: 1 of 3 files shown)


christian-bromann (approved):
LGTM 👍

PR #2450, opened 2026-05-28
feat(langgraph): add node-level timeouts (timeout on addNode, task, entrypoint, Send)

christian-bromann:
A `timeout` option is now supported on `StateGraph.addNode`, the functional API (`task`/`entrypoint`), and the `Send` constructor. Pass a number of milliseconds for a hard wall-clock cap, or a `TimeoutPolicy` for finer control:

```ts
import { TimeoutPolicy } from "@langchain/langgraph";

// hard wall-clock cap on each attempt
builder.addNode("agent", agentFn, { timeout: 60_000 });

// full control
builder.addNode("agent", agentFn, {
  timeout: {
    runTimeout: 60_000, // hard wall-clock cap, never refreshed
    idleTimeout: 10_000, // cap on time without observable progress
    refreshOn: "auto", // "auto" | "heartbeat"
  },
});

// per-task override
new Send("agent", state, { idleTimeout: 5_000 });
```

When a timeout fires, a `NodeTimeoutError` (carrying `node`, `kind` (`"run"`/`"idle"`), `timeout`, `elapsed`, `runTimeout`, `idleTimeout`) is raised, the attempt's buffered writes are dropped, and the node's `AbortSignal` is aborted. `idleTimeout` is refreshed by observable progress (writes, custom stream-writer calls, child-task scheduling, callback events) or an explicit `runtime.heartbeat()` call. The timer resets per retry attempt, and `NodeTimeoutError` is retryable under the default retry policy.

Ports langchain-ai/langgraph `#7599`, `#7646`, and `#7659`.

### DIFF
--- a/libs/langgraph-core/src/constants.ts
+++ b/libs/langgraph-core/src/constants.ts
@@ -130,6 +130,15 @@ export interface SendInterface<Node extends string = string, Args = any> {
   timeout?: TimeoutPolicy;
 }
 
+/** Keyword options for {@link Send}. */
+export type SendOptions = {
+  /**
+   * Per-task timeout policy overriding the target node's timeout for this
+   * pushed task. A bare number is treated as `runTimeout` (milliseconds).
+   */
+  timeout?: number | TimeoutPolicy;
+};
+
 export function _isSendInterface(x: unknown): x is SendInterface {
   const operation = x as SendInterface;
   return (
@@ -172,11 +181,11 @@ export function _isSendInterface(x: unknown): x is SendInterface {
  * };
  *
  * @remarks
- * A per-task timeout can be supplied as the third argument to override the
- * target node's configured timeout for this specific pushed task:
+ * A per-task timeout can be supplied via the third argument's `timeout` option
+ * to override the target node's configured timeout for this specific pushed task:
  *
  * ```typescript
- * new Send("generate_joke", { subjects: [subject] }, { idleTimeout: 5000 });
+ * new Send("generate_joke", { subjects: [subject] }, { timeout: { idleTimeout: 5000 } });
  * ```
  *
  * const graph = new StateGraph(ChainState)
@@ -212,10 +221,10 @@ export class Send<
    */
   public timeout?: TimeoutPolicy;
 
-  constructor(node: Node, args: Args, timeout?: number | TimeoutPolicy) {
+  constructor(node: Node, args: Args, options?: SendOptions) {
     this.node = node;
     this.args = _deserializeCommandSendObjectGraph(args) as Args;
-    this.timeout = coerceTimeoutPolicy(timeout);
+    this.timeout = coerceTimeoutPolicy(options?.timeout);
   }
 
   toJSON() {
@@ -407,9 +416,9 @@ export type CommandParams<
    *   - sequence of `Send` objects
    */
   goto?:
-  | Nodes
-  | SendInterface<Nodes> // eslint-disable-line @typescript-eslint/no-explicit-any
-  | (Nodes | SendInterface<Nodes>)[]; // eslint-disable-line @typescript-eslint/no-explicit-any
+    | Nodes
+    | SendInterface<Nodes> // eslint-disable-line @typescript-eslint/no-explicit-any
+    | (Nodes | SendInterface<Nodes>)[]; // eslint-disable-line @typescript-eslint/no-explicit-any
 };
 
 /**
@@ -651,7 +660,11 @@ export function _deserializeCommandSendObjectGraph(
       result = new Command(x);
       seen.set(x, result);
     } else if (_isSendInterface(x)) {
-      result = new Send(x.node, x.args, x.timeout);
+      result = new Send(
+        x.node,
+        x.args,
+        x.timeout !== undefined ? { timeout: x.timeout } : undefined
+      );
       seen.set(x, result);
     } else if ("lc_serializable" in x && x.lc_serializable) {
       result = x;
--- a/libs/langgraph-core/src/constants.ts
+++ b/libs/langgraph-core/src/constants.ts
@@ -1,4 +1,8 @@
 import { PendingWrite } from "@langchain/langgraph-checkpoint";
+import {
+  coerceTimeoutPolicy,
+  type TimeoutPolicy,
+} from "./pregel/utils/timeout.js";
 
 /** Special reserved node name denoting the start of a graph. */
 export const START = "__start__";
@@ -119,6 +123,11 @@ export class CommandInstance<
 export interface SendInterface<Node extends string = string, Args = any> {
   node: Node;
   args: Args;
+  /**
+   * Optional per-task timeout policy that overrides the target node's timeout
+   * for this specific pushed task.
+   */
+  timeout?: TimeoutPolicy;
 }
 
 export function _isSendInterface(x: unknown): x is SendInterface {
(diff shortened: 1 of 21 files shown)


libs/langgraph-core/src/pregel/algo.ts
```diff
@@ -881,6 +882,8 @@ export function _prepareSingleTask<
           id: taskId,
           path: taskPath,
           writers: proc.getWriters(),
+          // a per-Send timeout overrides the target node's configured timeout
+          timeout: (packet as Send).timeout ?? proc.timeout,
```
hntrl: weird type assign

libs/langgraph-core/src/pregel/timeout.ts
```diff
@@ -0,0 +1,304 @@
...
+  if (outcome.type === "err") {
+    dispose?.();
+    throw outcome.error;
+  }
+
+  // A watchdog fired: close the scope (drop late writes/calls), discard the
+  // attempt's buffered writes, abort the node, and surface a NodeTimeoutError.
+  const elapsed = Date.now() - start;
+  scope.close();
+  task.writes.splice(0, task.writes.length);
```
hntrl, line 313: ?

libs/langgraph-core/src/pregel/timeout.ts
```diff
@@ -0,0 +1,304 @@
...
+    config.signal,
+    timeoutController.signal
+  );
+
+  const scopedConfig = wrapConfig(
+    { ...config, signal: composedSignal },
+    scope,
+    policy,
+    String(task.name)
+  );
```
hntrl, line 227: the wrapConfig indirection does not spark joy

hntrl: I don't have concrete feedback here, I guess this is more nit than anything

libs/langgraph-core/src/pregel/timeout.ts
```diff
@@ -0,0 +1,304 @@
...
+  // A watchdog fired: close the scope (drop late writes/calls), discard the
+  // attempt's buffered writes, abort the node, and surface a NodeTimeoutError.
+  const elapsed = Date.now() - start;
+  scope.close();
+  task.writes.splice(0, task.writes.length);
+  // Abort BEFORE disposing the combined signal so the abort actually
+  // propagates to the node's signal (dispose removes the relay listeners).
+  timeoutController.abort();
+  // Prevent an unhandled rejection from the abandoned background task.
+  bg.catch(() => undefined);
```
hntrl: this doesn't do anything

libs/langgraph-core/src/tests/timeout.test.ts
```diff
@@ -0,0 +1,439 @@
...
+    const { result, error } = await _runWithRetry(taskObj);
+    expect(error).toBeUndefined();
+    expect(result).toBe("ok");
+    expect(calls).toBe(2);
+  });
+});
+
+const TimeoutState = Annotation.Root({
+  x: Annotation<number>,
+});
```
hntrl: Annotation.Root boo 🍅

libs/langgraph-core/src/constants.ts
```diff
@@ -188,13 +205,26 @@ export class Send<
...
 
-  constructor(node: Node, args: Args) {
+  /**
+   * Optional per-task timeout policy that overrides the target node's timeout
+   * for this specific pushed task. A bare number is treated as a hard
+   * `runTimeout` (in milliseconds).
+   */
+  public timeout?: TimeoutPolicy;
+
+  constructor(node: Node, args: Args, timeout?: number | TimeoutPolicy) {
```
hntrl: what would you say to making `timeout` a concern of args? positional kwargs for something like this seems bad

libs/langgraph-core/src/pregel/timeout.ts
christian-bromann: Same as in Python, see https://github.com/langchain-ai/langgraph/blob/f25a0f4f4c389e9e6ed66b998a23756404b7ebdb/libs/langgraph/langgraph/pregel/_retry.py#L492-L493

libs/langgraph-core/src/pregel/timeout.ts
christian-bromann: It attaches a handler whose only job is to swallow that late rejection. The return value is ignored on purpose.

christian-bromann: Will add a test to verify this behavior and better justify this line.

libs/langgraph-core/src/constants.ts
christian-bromann: Wouldn't fold into 2 parameters but will extend the 3rd one as `options: { timeout?: number | TimeoutPolicy }`

libs/langgraph-core/src/pregel/timeout.ts
hntrl: oh this splice line isn't actually clearing the array then
```
> const x = [1,2,3]
undefined
> x.splice(0,x.length)
[ 1, 2, 3 ]
```

libs/langgraph-core/src/pregel/timeout.ts
hntrl: right but because this promise chaining is declarative doesn't that make this an unhandled promise?

libs/langgraph-core/src/pregel/timeout.ts
```diff
@@ -0,0 +1,304 @@
...
+
+  const scopedConfig = wrapConfig(
+    { ...config, signal: composedSignal },
+    scope,
+    policy,
+    String(task.name)
+  );
+
+  const start = Date.now();
+  const bg = invoke(scopedConfig);
```
hntrl, line 230: do we just sacrifice bg to be run in the background? is there any way we can pin this to the `outcome` promise so its not unhandled?

libs/langgraph-core/src/errors.ts
```diff
@@ -117,6 +117,92 @@ export function isGraphInterrupt(e?: unknown): e is GraphInterrupt {
...
+  }
+}
+
+export function isNodeTimeoutError(e?: unknown): e is NodeTimeoutError {
+  return (
+    e !== undefined &&
+    (e as NodeTimeoutError).name === NodeTimeoutError.unminifiable_name
+  );
+}
+
```
hntrl, line 205: are we just using unminifiable_name for the instance check? what would you say to trending to using something like this instead: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/hasInstance

libs/langgraph-core/src/pregel/timeout.ts
christian-bromann: It, would:

```ts
> const x = [1,2,3]
undefined
> x.splice(0,x.length)
(3) [1, 2, 3]
> x
[]
```

libs/langgraph-core/src/pregel/timeout.ts
christian-bromann: ah gotcha, will add a `void`

libs/langgraph-core/src/pregel/timeout.ts
christian-bromann: Will use a Promise.race to make it more explicit.

libs/langgraph-core/src/errors.ts
christian-bromann: I don't prefer this implementation either tho it aligns with existing [error implementation](https://github.com/langchain-ai/langgraphjs/blob/4269e69a6db1ce15f40d30d85c3308c89e434330/libs/langgraph-core/src/errors.ts). I would defer an optimization here to another PR where we clean this up.

hntrl (dismissed):

PR #2451, opened 2026-05-28
feat(langgraph): add node-level error handlers

christian-bromann:
`StateGraph.addNode(name, fn, { errorHandler })` now accepts a first-class node-level error handler. The handler runs ONLY after the failing node's `retryPolicy` is exhausted, so retry and handling stay decoupled. It receives a typed `NodeError { node, error }`, can return a state update, and can route to a recovery branch via `new Command({ goto })` (saga / compensation flows).

Failure provenance is checkpointed (via a reserved `ERROR_SOURCE_NODE` write) so handlers observe the same context after a checkpoint resume. Uncaught node errors without a handler still abort the run as before, and `GraphBubbleUp` errors (such as `interrupt()`) are never swallowed by a handler.

Ports the Python feature from langchain-ai/langgraph#7233.

### DIFF
--- a/libs/langgraph-core/src/graph/graph.ts
+++ b/libs/langgraph-core/src/graph/graph.ts
@@ -40,6 +40,7 @@ import {
 } from "../utils.js";
 import {
   InvalidUpdateError,
+  NodeError,
   NodeInterrupt,
   UnreachableNodeError,
 } from "../errors.js";
@@ -223,7 +224,27 @@ export type NodeSpec<RunInput, RunOutput> = {
   subgraphs?: Pregel<any, any>[];
   ends?: string[];
   defer?: boolean;
+  /** Whether this node is an auto-generated node-level error handler. */
+  isErrorHandler?: boolean;
+  /** Name of the auto-generated error handler node to run on failure. */
+  errorHandlerNode?: string;
 };
+
+/**
+ * A node-level error handler callable.
+ *
+ * Invoked with the node input state, a {@link NodeError} describing the failed
+ * node and thrown error, and the runnable config. The handler runs ONLY after
+ * the failing node's {@link RetryPolicy} is exhausted. It may return a state
+ * update or a `Command` (to route via `goto`).
+ */
+export type NodeErrorHandler = (
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  state: any,
+  error: NodeError,
+  config?: LangGraphRunnableConfig
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+) => any;
 
 export type AddNodeOptions<Nodes extends string = string> = {
   metadata?: Record<string, unknown>;
@@ -231,6 +252,12 @@ export type AddNodeOptions<Nodes extends string = string> = {
   subgraphs?: Pregel<any, any>[];
   ends?: Nodes[];
   defer?: boolean;
+  /**
+   * Optional node-level error handler. Runs only after this node's
+   * {@link RetryPolicy} is exhausted. Receives a {@link NodeError} with the
+   * failed node name and error, and may return a state update or `Command`.
+   */
+  errorHandler?: NodeErrorHandler;
 };
 
 export class Graph<
@@ -596,8 +623,26 @@ export class Graph<
         allTargets.add(target);
       }
     }
+    // Node-level error handlers can route to any node via `Command({ goto })`
+    // (saga / compensation flows), so treat them like an open-ended branch:
+    // any node may be a recovery target reachable from a handler.
+    const hasErrorHandler = Object.values<NodeSpecType>(this.nodes).some(
+      (node) => (node as NodeSpec<unknown, unknown>).isErrorHandler
+    );
+    if (hasErrorHandler) {
+      for (const node of Object.keys(this.nodes)) {
+        allTargets.add(node);
+      }
+    }
     // validate targets
     for (const node of Object.keys(this.nodes)) {
+      // auto-generated error handler nodes are reachable only on failure of
+      // their source node, so they are exempt from the reachability check.
+      if (
+        (this.nodes[node as N] as NodeSpec<unknown, unknown>).isErrorHandler
+      ) {
+        continue;
+      }
       if (!allTargets.has(node)) {
         throw new UnreachableNodeError(
           [
--- a/libs/langgraph-core/src/graph/graph.ts
+++ b/libs/langgraph-core/src/graph/graph.ts
@@ -47,6 +47,7 @@ import {
 import { StateDefinition, StateType } from "./annotation.js";
 import { isPregelLike } from "../pregel/utils/subgraph.js";
 import type { StreamTransformer } from "../stream/types.js";
+import type { GraphNodeReturnValue } from "./types.js";
 
 export interface BranchOptions<
   IO,
@@ -231,6 +232,19 @@ export type NodeSpec<RunInput, RunOutput> = {
 };
 
 /**
+ * Return value type for node-level error handlers.
+ *
+ * Handlers may return a partial state update, a `Command`, or a Promise of either.
+ *
+ * @template Update - The update type (what fields can be returned)
+ * @template Nodes - Union of valid node names for Command.goto
+ */
+export type NodeErrorHandlerReturnValue<
+  Update,
+  Nodes extends string = string,
+> = GraphNodeReturnValue<Update, Nodes>;
+
+/**
  * A node-level error handler callable.
  *
  * Invoked with the node input state, a {@link NodeError} describing the failed
@@ -238,11 +252,15 @@ export type NodeSpec<RunInput, RunOutput> = {
  * the failing node's {@link RetryPolicy} is exhausted. It may return a state
  * update or a `Command` (to route via `goto`).
  */
-export type NodeErrorHandler<TState = unknown> = (
+export type NodeErrorHandler<
+  TState = unknown,
+  TUpdate = Partial<TState>,
+  Nodes extends string = string,
+> = (
   state: TState,
   error: NodeError,
   config?: LangGraphRunnableConfig
-) => unknown;
+) => NodeErrorHandlerReturnValue<TUpdate, Nodes>;
 
 export type AddNodeOptions<Nodes extends string = string> = {
   metadata?: Record<string, unknown>;
--- a/libs/langgraph-core/src/graph/graph.ts
+++ b/libs/langgraph-core/src/graph/graph.ts
@@ -238,13 +238,11 @@ export type NodeSpec<RunInput, RunOutput> = {
  * the failing node's {@link RetryPolicy} is exhausted. It may return a state
  * update or a `Command` (to route via `goto`).
  */
-export type NodeErrorHandler = (
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  state: any,
+export type NodeErrorHandler<TState = unknown> = (
+  state: TState,
   error: NodeError,
   config?: LangGraphRunnableConfig
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-) => any;
+) => unknown;
 
 export type AddNodeOptions<Nodes extends string = string> = {
   metadata?: Record<string, unknown>;
@@ -252,12 +250,6 @@ export type AddNodeOptions<Nodes extends string = string> = {
   subgraphs?: Pregel<any, any>[];
   ends?: Nodes[];
   defer?: boolean;
-  /**
-   * Optional node-level error handler. Runs only after this node's
-   * {@link RetryPolicy} is exhausted. Receives a {@link NodeError} with the
-   * failed node name and error, and may return a state update or `Command`.
-   */
-  errorHandler?: NodeErrorHandler;
 };
 
 export class Graph<
--- a/libs/langgraph-core/src/graph/graph.ts
+++ b/libs/langgraph-core/src/graph/graph.ts
@@ -79,16 +79,16 @@ type CompiledGraphTypeStreamTransformers<Spec> = Spec extends {
   streamTransformers: infer Transformers;
 }
   ? Transformers extends ReadonlyArray<
-      () => StreamTransformer<any> // eslint-disable-line @typescript-eslint/no-explicit-any
-    >
-    ? Transformers
-    : Transformers extends ReadonlyArray<
-          StreamTransformer<any> // eslint-disable-line @typescript-eslint/no-explicit-any
-        >
-      ? { readonly [K in keyof Transformers]: () => Transformers[K] }
-      : Transformers extends StreamTransformer<any> // eslint-disable-line @typescript-eslint/no-explicit-any
-        ? readonly [() => Transformers]
-        : []
+    () => StreamTransformer<any> // eslint-disable-line @typescript-eslint/no-explicit-any
+  >
+  ? Transformers
+  : Transformers extends ReadonlyArray<
+    StreamTransformer<any> // eslint-disable-line @typescript-eslint/no-explicit-any
+  >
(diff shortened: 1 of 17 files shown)


libs/langgraph-core/src/pregel/runner.ts
```diff
@@ -337,7 +348,43 @@ export class PregelRunner {
...
+          settledPregelTask,
+          settledError
+        );
+        if (handlerTask !== undefined) {
+          settled.errorHandled = true;
+          executingTasksMap[handlerTask.id] = _runWithRetry(
+            handlerTask,
+            retryPolicy,
+            { [CONFIG_KEY_CALL]: call?.bind(thisCall, this, handlerTask) },
+            signals?.composedAbortSignal
```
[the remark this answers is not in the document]
christian-bromann: This is ok, Python has the same inline scheduling behavior and I don't want to diverge from that.

libs/langgraph-core/src/pregel/retry.ts
```diff
@@ -60,6 +60,12 @@ export type SettledPregelTask = {
   task: PregelExecutableTask<any, any>;
   error: Error;
   signalAborted?: boolean;
+  /**
+   * Set when this task's error was routed to a node-level error handler. When
+   * true, the runner records failure provenance but does not treat the error
+   * as fatal to the run.
+   */
+  errorHandled?: boolean;
 };
```
hntrl: why should a task care if it has an error that's been handled or not? if an error gets handled, shouldn't the behavior explicitly be not to exit?

libs/langgraph-core/src/pregel/runner.ts
```diff
@@ -371,6 +422,18 @@ export class PregelRunner {
         }
       } else if (isGraphBubbleUp(error) && task.writes.length) {
         this.loop.putWrites(task.id, task.writes);
+      } else if (errorHandled) {
+        // The error was routed to a node-level error handler. Record failure
+        // provenance (ERROR + ERROR_SOURCE_NODE) so the originating task counts
+        // as "done" for this superstep and handlers see the same context after
+        // a checkpoint resume. The handler task (scheduled separately) carries
+        // the recovery state update / routing.
+        task.writes.splice(0, task.writes.length);
```
hntrl: ?

libs/langgraph-core/src/graph/graph.ts
```diff
@@ -596,8 +615,26 @@ export class Graph<
         allTargets.add(target);
       }
     }
+    // Node-level error handlers can route to any node via `Command({ goto })`
+    // (saga / compensation flows), so treat them like an open-ended branch:
+    // any node may be a recovery target reachable from a handler.
+    const hasErrorHandler = Object.values<NodeSpecType>(this.nodes).some(
+      (node) => (node as NodeSpec<unknown, unknown>).isErrorHandler
```
hntrl: this might be wishful but why do we have to do weird type assertions here

libs/langgraph-core/src/tests/node_error_handler.test.ts
```diff
@@ -0,0 +1,315 @@
...
+      return 0 as unknown as ReturnType<typeof setTimeout>;
+    }) as unknown as typeof setTimeout);
+  });
+
+  afterAll(() => {
+    vi.restoreAllMocks();
+  });
+
+  it("runs the error handler only after the retry policy is exhausted", async () => {
+    const State = Annotation.Root({ foo: Annotation<string> });
```
hntrl: use StateSchema instead for all these tests. (we just need to deprecate Annotation.Root)

libs/langgraph-core/src/tests/node_error_handler.test.ts
```diff
@@ -0,0 +1,315 @@
...
+  beforeAll(() => {
+    vi.spyOn(global, "setTimeout").mockImplementation(((fn: () => void) => {
+      fn();
+      return 0 as unknown as ReturnType<typeof setTimeout>;
+    }) as unknown as typeof setTimeout);
+  });
+
+  afterAll(() => {
+    vi.restoreAllMocks();
+  });
```
hntrl: whats this for?

christian-bromann: So the retries within the tests don't take forever, overwrite `setTimeout` and change all timeout parameter to `0`.

libs/langgraph-core/src/pregel/runner.ts
christian-bromann: I've aligned `_commit` with Python’s `commit()` in `_runner.py` (append `ERROR` / `ERROR_SOURCE_NODE` to existing task.writes when a handler is configured, instead of clearing via splice); retry already clears writes before each attempt, matching `_retry.py`.

libs/langgraph-core/src/pregel/retry.ts
christian-bromann: Good call, I dropped `errorHandled` and track routed exceptions on `PregelRunner` via a `WeakSet`, which matches Python’s `_handled_exception_ids` more closer.

libs/langgraph-core/src/graph/graph.ts
```diff
@@ -223,8 +224,26 @@ export type NodeSpec<RunInput, RunOutput> = {
...
+ * Invoked with the node input state, a {@link NodeError} describing the failed
+ * node and thrown error, and the runnable config. The handler runs ONLY after
+ * the failing node's {@link RetryPolicy} is exhausted. It may return a state
+ * update or a `Command` (to route via `goto`).
+ */
+export type NodeErrorHandler<TState = unknown> = (
+  state: TState,
+  error: NodeError,
+  config?: LangGraphRunnableConfig
+) => unknown;
```
hntrl: don't we want this to have the same return signature as any other node since it is in effect a new node?

libs/langgraph-core/src/graph/graph.ts
```diff
@@ -596,8 +615,24 @@ export class Graph<
         allTargets.add(target);
       }
     }
+    // Node-level error handlers can route to any node via `Command({ goto })`
+    // (saga / compensation flows), so treat them like an open-ended branch:
+    // any node may be a recovery target reachable from a handler.
+    const hasErrorHandler = Object.values<NodeSpecType>(this.nodes).some(
+      (node) => node.isErrorHandler
+    );
```
hntrl, line 641: so if any node has an error handler we're invalidating reach-ability check?

libs/langgraph-core/src/graph/state.ts
```diff
@@ -119,18 +123,33 @@ export type StateGraphNodeSpec<RunInput, RunOutput> = NodeSpec<
...
+ * @template HandlerState - State type passed to the node-level error handler
  */
 export type StateGraphAddNodeOptions<
   Nodes extends string = string,
   InputSchema extends StateDefinitionInit | undefined =
     | StateDefinitionInit
     | undefined,
+  HandlerState = InputSchema extends StateDefinitionInit
+    ? ExtractStateType<InputSchema>
+    : unknown,
```
hntrl, line 218: depending on how you ask this might be nit, but considering that HandlerState is type constrained can we have the conditional check inside of the type definition? Or why would you need to have HandlerState separate?

libs/langgraph-core/src/pregel/retry.ts
```diff
@@ -156,9 +157,10 @@ export async function _runWithRetry<
         resolvedRetryPolicy.maxInterval ?? DEFAULT_MAX_INTERVAL,
         interval * (resolvedRetryPolicy.backoffFactor ?? DEFAULT_BACKOFF_FACTOR)
       );
-      const intervalWithJitter = resolvedRetryPolicy.jitter
-        ? Math.floor(interval + Math.random() * 1000)
-        : interval;
+      const intervalWithJitter =
+        (resolvedRetryPolicy.jitter ?? DEFAULT_JITTER)
+          ? Math.floor(interval + Math.random() * 1000)
+          : interval;
```
hntrl: this is what we do in python too? This feels like its going to lead to some very awkward debugging trails

libs/langgraph-core/src/graph/state.ts
```diff
@@ -119,18 +123,33 @@ export type StateGraphNodeSpec<RunInput, RunOutput> = NodeSpec<
...
+   * {@link RetryPolicy} is exhausted. Receives a {@link NodeError} with the
+   * failed node name and error, and may return a state update or `Command`.
+   */
+  errorHandler?: NodeErrorHandler<HandlerState>;
 } & AddNodeOptions<Nodes>;
 
+type StateGraphAddNodeOptionsWithNodeInput<
+  Nodes extends string,
+  NodeInput,
+> = StateGraphAddNodeOptions<Nodes, StateDefinitionInit | undefined, NodeInput>;
```
hntrl: We just have this alias so we don't have to duplicate `StateDefinitionInit | undefined everywhere` yah? ooc why don't we bring SD and just use StateGraphAddNodeOptions with stronger type args?

libs/langgraph-core/src/graph/state.ts
```diff
@@ -907,6 +929,44 @@ export class StateGraph<
...
+        });
+        const handlerSpec: StateGraphNodeSpec<S, U> = {
+          runnable: handlerRunnable as unknown as Runnable<S, U>,
+          metadata: undefined,
+          input: inputSpec ?? this._schemaDefinition,
+          retryPolicy: undefined,
+          cachePolicy: undefined,
+          isErrorHandler: true,
+        };
+        this.nodes[errorHandlerNode as N] = handlerSpec;
```
hntrl, line 1213: is this going to make studio render super weird?

libs/langgraph-core/src/graph/state.ts
```diff
@@ -907,6 +929,44 @@ export class StateGraph<
...
 
+      // If an error handler is provided, register an auto-generated handler
+      // node that runs only after this node's retry policy is exhausted.
+      let errorHandlerNode: string | undefined;
+      if (options?.errorHandler !== undefined) {
+        errorHandlerNode = `__error_handler__${key}`;
+        if (errorHandlerNode in this.nodes) {
+          throw new Error(
+            `Auto-generated error handler node \`${errorHandlerNode}\` already exists.`
+          );
```
hntrl, line 1186: this seems like an internal error, is there a way we can be more explicit? Like when should this ever happen?

libs/langgraph-core/src/tests/node_error_handler.test.ts
hntrl: ok I'll take your word for it. If we have to monkey patch setTimeout like this I think its almost always indicative of some other obscure dynamics

libs/langgraph-core/src/tests/types.test-d.ts
```diff
@@ -606,6 +607,54 @@ describe("GraphNode", () => {
...
 });
 
+describe("StateGraph.addNode errorHandler", () => {
+  it("accepts handlers with the typed graph state", () => {
+    const State = new StateSchema({
+      count: z.number().default(0),
+      name: z.string(),
+    });
+
+    const typedHandler = (state: typeof State.State, _error: NodeError) => {
```
hntrl, line 617: will these types be inferred if I attach errorHandler directly?

libs/langgraph-core/src/tests/node_error_handler.test.ts
christian-bromann: Actually we could just use fake timers.

libs/langgraph-core/src/graph/graph.ts
christian-bromann: Good catch, should behave like other nodes too

libs/langgraph-core/src/graph/graph.ts
christian-bromann: Yes! If any node registers an `errorHandler`, we treat handler `Command({ goto })` like an open-ended branch and add every node to `allTargets`, which disables the unreachable-node check for the whole graph (not the other validation passes) because we can't statically know recovery targets at compile time.

christian-bromann: For example:
```ts
const graph = new StateGraph(State)
  .addNode("alwaysFailing", fn, {
    errorHandler: () => new Command({ goto: "nextNode" }),
  })
  .addNode("nextNode", recoveryFn)   // no edge from START — OK with handler
  .addNode("orphan", neverUsedFn)    // also OK once any errorHandler exists
  .addEdge(START, "alwaysFailing")
  .compile();
```

Without the `hasErrorHandler` widen, `nextNode` would throw `UnreachableNodeError` even though it is a valid recovery target. With it, orphan would also pass, so it is a deliberate graph-wide relaxation.

libs/langgraph-core/src/graph/state.ts
christian-bromann: `HandlerState` has to stay a separate type parameter because the default `addNode` overload types the handler from the node action’s `NodeInput` (via `StateGraphAddNodeOptionsWithNodeInput`), not from `InputSchema`. Inlining the conditional on `InputSchema` would collapse that to `unknown` when no `options.input` is provided.

libs/langgraph-core/src/pregel/retry.ts
christian-bromann: Yes, this matches Python’s `interval + random.uniform(0, 1)` (seconds there, ms here). I fixed the backoff formula, which was the bigger source of confusing retry timing, and added tests so the sleep delays are explicit.

libs/langgraph-core/src/graph/graph.ts
hntrl: Hmm I mean the same thinking could be applied for if I specify `new Command` in any normal node

libs/langgraph-core/src/pregel/retry.ts
hntrl: Sorry I guess I meant the default being that we do jitter

libs/langgraph-core/src/graph/state.ts
christian-bromann: Originally this looked weird in JS:

Screenshot 2026-06-03 at 11 51 52 AM

Compared to Python:

Screenshot 2026-06-03 at 11 51 23 AM

Which arguably doesn't look right either, but better.

I made some changes so that after static edges are collected, JS now adds implicit `node → __end__` edges for terminal nodes: nodes that are targets of a non-conditional edge and never appear as a source. Auto-generated error-handler nodes are skipped. With that we get a better graph structure:

Screenshot 2026-06-03 at 12 08 26 PM

I'ld suggest revisit this later to optimize graph structure for error handlers. I posted in our internal langgraph channel for feedback.

libs/langgraph-core/src/tests/types.test-d.ts
christian-bromann: yep, it is written this way to verify that this actually works.

libs/langgraph-core/src/pregel/retry.ts
christian-bromann: We do jitter in Python. I agree this could make debugging more complex but also would like to see this play out in an actual issue first.

libs/langgraph-core/src/graph/state.ts
christian-bromann: Updated error message.

libs/langgraph-core/src/graph/graph.ts
christian-bromann: For normal nodes we declare allowed destinations via ends (and/or edges), and we only widen allTargets for those declared targets, we don’t skip reachability for the whole graph. Error-handler nodes are auto-generated without an ends option, so recovery targets like recover in the demo would fail `UnreachableNodeError` unless we treat handlers like an open-ended branch; tightening would be adding ends on errorHandler rather than graph-wide disable.

hntrl (dismissed):

PR #2452, opened 2026-05-29
feat(core): DeltaChannel + writes-history saver API

christian-bromann:
`DeltaChannel` is a reducer channel that stores only a sentinel in checkpoint blobs instead of the full accumulated value, reconstructing state on read by replaying ancestor writes through a batch reducer. This avoids re-serializing the entire accumulated value at every step (e.g. long message histories).

- `DeltaChannel(reducer, { snapshotFrequency })` in `@langchain/langgraph`, count-based snapshot cadence (default `snapshotFrequency=1000`) plus a system bound `DELTA_MAX_SUPERSTEPS_SINCE_SNAPSHOT` (default 5000, env `LANGGRAPH_DELTA_MAX_SUPERSTEPS_SINCE_SNAPSHOT`).
- `messagesDeltaReducer` — a batching-invariant messages reducer that coerces raw object/string writes, for use with `DeltaChannel`.
- `BaseCheckpointSaver.getDeltaChannelHistory({ config, channels })` (beta), walks the parent chain returning per-channel `{ writes, seed? }`, with a direct-storage override in `MemorySaver`.
- `counters_since_delta_snapshot` added to `CheckpointMetadata`; `DeltaSnapshot` serialization support in the JSON+ serializer.

Reconstruction is wired through the Pregel read/execution paths (initialization, `getState`, `updateState`, local reads) and `exit` durability accumulates and anchors delta writes so threads remain reconstructible without forcing snapshots.


libs/langgraph-core/src/graph/messages_reducer.ts
```diff
@@ -129,3 +129,79 @@ export function messagesStateReducer(
...
+ * ```typescript
+ * import { DeltaChannel, messagesDeltaReducer } from "@langchain/langgraph";
+ *
+ * const channel = new DeltaChannel(messagesDeltaReducer);
+ * ```
+ */
+export function messagesDeltaReducer(
+  state: BaseMessage[],
+  writes: Messages[]
+): BaseMessage[] {
```
hntrl, line 162: would it make sense to refactor `messagesStateReducer` to take in a value that only has one write? Basically trying to protect against diverging paths later on

hntrl (dismissed):

libs/langgraph-core/src/graph/messages_reducer.ts
christian-bromann: I would suggest to keep as it for the following reason:

- Python parity. DeltaChannel's `reducer(state, [writes])` batch contract and the separate messagesDeltaReducer mirror Python's DeltaChannel / _messages_delta_reducer. Folding a single-write reducer would diverge from Python's public API.
- Performance. The batch reducer builds the id index once; folding messagesStateReducer re-coerces all of left and rebuilds the map per write → `O(N·M)`, on the replay path (~5000 writes) the channel is meant to keep cheap.

PR #2455, opened 2026-06-01
fix(sdk): prefer completed task's direct mapping over pending checkpoint's positional guess in fetchSubagentHistory

### LINKED ISSUE #2454: fetchSubagentHistory (SDK) restores a subagent from the wrong subgraph namespace when the latest checkpoint has a pending task
## Description

When restoring subagent conversations after a page refresh / thread reload, `StreamManager.fetchSubagentHistory` (`libs/sdk/src/ui/manager.ts`) resolves each subagent `tool_call_id` to its subgraph `checkpoint_ns` by scanning the parent thread's history. Under a specific — but common — condition it maps a subagent to the **wrong** namespace, so the subagent's messages are reconstructed from another task's checkpoint (or not at all).

## Conditions to reproduce

- A run uses subagent / `task` tool calls (LangGraph v2 PUSH tasks), **and**
- the most recent checkpoint in the parent thread is a **still-pending** PUSH task — e.g. the run is stopped at an interrupt, or reloaded mid-flight — while the AI message that triggered an **already-completed** subagent still lives in that head checkpoint's accumulated `values.messages`.

## Root cause

`fetchSubagentHistory` scans history newest-first and, per checkpoint, tries:

1. a **direct mapping** from a completed task's result `ToolMessage.tool_call_id` (the code comments call this *"more robust than positional alignment ... preferred"*), then
2. a **positional fallback** aligning push tasks to the AI message's subagent tool calls by Send index (`task.path[1]`).

It then **breaks on the first checkpoint that yields _any_ mapping** (direct *or* positional).

When the head checkpoint holds a pending task, its direct map is empty, so the positional fallback runs against the head checkpoint instead, latches onto the stale (already-completed) AI tool call, aligns it to the head's unrelated pending task, and breaks — before the loop ever reaches the older checkpoint whose task result holds the **correct** direct mapping.

Net effect: a newer, lower-confidence positional guess shadows an older, authoritative direct mapping — the opposite of the documented intent.

## Expected vs actual

- **Expected:** the subagent is restored from its own subgraph checkpoint (the namespace derived from its completed task result).
- **Actual:** it is restored from the head checkpoint's unrelated pending task's namespace → wrong messages, or empty.

## Note

No backend change is required — `getHistory` already serializes the correct material (the completed task result with the matching `tool_call_id`); only the client-side resolution order is wrong. I'll open a PR with a fix (two-phase resolution: collect all direct mappings across the full history first, then apply the positional fallback only to still-unmapped tool calls) plus a regression test.

## Environment

- `@langchain/langgraph-sdk`, affects current `main` (`libs/sdk/src/ui/manager.ts`).

JHSeo-git:
Fixes #2454

## Problem

`StreamManager.fetchSubagentHistory` (`libs/sdk/src/ui/manager.ts`) restores each subagent's conversation from its subgraph checkpoint by first resolving every subagent `tool_call_id` to its subgraph `checkpoint_ns`. It scans the parent thread's history **newest-first** and, per checkpoint, tries two strategies:

1. **Direct mapping** — read the `tool_call_id` straight off a completed PUSH task's result `ToolMessage` (`task.name + ":" + task.id`). The code comments correctly call this *"more robust than positional alignment ... preferred"*.
2. **Positional fallback** — when task results aren't populated yet, align push tasks to the AI message's subagent tool calls by Send index (`task.path[1]`).

The loop **breaks on the first checkpoint that yields any mapping** (direct *or* positional). That break is the bug.

When the most recent checkpoint is a still-pending PUSH task — e.g. a run stopped at an interrupt — its task has no result, so the direct map is empty and the **positional fallback runs against the head checkpoint instead**. The fallback's backward scan of `values.messages` then latches onto a *stale* AI message whose subagent actually **completed in an older checkpoint**, positionally aligns it to the head's unrelated pending task, and breaks — before the loop ever reaches the older checkpoint whose task result holds the **correct** direct mapping.

Net effect: the subagent's history is reconstructed from the **wrong subgraph namespace** (or not at all). A newer, lower-confidence positional guess silently shadows an older, authoritative direct mapping — the opposite of the code's stated intent.

## Fix

Split namespace resolution into two phases:

1. **Phase 1 — direct mapping across the _entire_ history first.** Collect all `tool_call_id → namespace` mappings from completed task results before attempting any fallback. These are unambiguous, so no pending head checkpoint can pre-empt them.
2. **Phase 2 — positional fallback only for tool calls still unmapped** after phase 1 (the genuinely live/pending case). This preserves the existing behavior for in-flight runs while guaranteeing a correct direct mapping is never overwritten by a positional guess.

No backend/protocol change is required — the server already serializes the correct material (the completed task result with the matching `tool_call_id`); only the client-side resolution order was wrong.

## Testing

- Added a regression test (`fetchSubagentHistory namespace resolution`): a pending head checkpoint sitting in front of an older checkpoint that holds the correct completed mapping. Verified it **fails on the pre-fix code** (`expected [ 'Stale pending result' ] to include 'Correct research result'`) and **passes with this change**.
- `vitest run` (SDK): **595 passed**, no type errors.
- `oxlint`: 0 warnings / 0 errors · `oxfmt --check`: clean · `tsc -p libs/sdk/tsconfig.json --noEmit`: clean.

### DIFF
--- a/.changeset/brown-mice-listen.md
+++ b/.changeset/brown-mice-listen.md
@@ -1,0 +1,5 @@
+---
+"@langchain/langgraph-sdk": patch
+---
+
+fix(sdk): prefer completed task's direct mapping over pending checkpoint's positional guess in fetchSubagentHistory
--- a/libs/sdk/src/ui/manager.test.ts
+++ b/libs/sdk/src/ui/manager.test.ts
@@ -1495,4 +1495,132 @@ describe("StreamManager", () => {
       expect(mgr.getSubagentsByMessage(streamingId)).toHaveLength(0);
     });
   });
+
+  describe("fetchSubagentHistory namespace resolution", () => {
+    it("prefers a completed task's direct mapping over a newer pending checkpoint's positional guess", async () => {
+      /**
+       * Regression test: the most recent checkpoint is a still-pending (e.g.
+       * interrupted) PUSH task. That head checkpoint has no task result, so it
+       * cannot produce a direct tool_call_id mapping — but its accumulated
+       * `values.messages` still contains the AI message whose subagent tool
+       * call already completed in an OLDER checkpoint.
+       *
+       * The positional fallback, run against the head checkpoint, would align
+       * that stale AI tool call to the head's unrelated pending push task and
+       * stop — never reaching the older checkpoint whose task result holds the
+       * correct namespace. Direct mappings must take precedence across the
+       * whole history, not just the first checkpoint that yields any mapping.
+       */
+      const sm = new StreamManager<TestState>(messageManager, {
+        throttle: false,
+        subagentToolNames: ["task"],
+      });
+
+      // AI message whose subagent tool call ("call-real") completed in an
+      // OLDER checkpoint. Registers a pending subagent with empty messages.
+      const aiMessage = {
+        type: "ai",
+        id: "ai-completed",
+        content: "",
+        tool_calls: [
+          {
+            id: "call-real",
+            name: "task",
+            args: {
+              subagent_type: "researcher",
+              description: "Research the real topic",
+            },
+          },
+        ],
+      };
+
+      sm.reconstructSubagents(
+        [aiMessage] as unknown as Parameters<typeof sm.reconstructSubagents>[0]
+      );
+      expect(sm.getSubagent("call-real")?.messages).toHaveLength(0);
+
+      // Main history is newest-first:
+      //   [0] head    — pending push task, no result; stale AI message in values
+      //   [1] older   — completed push task; result maps call-real -> task:uuid-real
+      const mainHistory = [
+        {
+          values: { messages: [aiMessage] },
+          tasks: [
+            {
+              id: "uuid-pending",
+              name: "task",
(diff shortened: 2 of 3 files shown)


christian-bromann (dismissed):
Thanks @JHSeo-git — great catch and a clean fix.

For new work, `@langchain/react` v1 avoids this entirely: subagent content comes from namespace-scoped selector hooks (`useMessages(stream, subagent)`) instead of client-side `getHistory` checkpoint guessing. Migration guide: [v1-migration.md §7](https://github.com/langchain-ai/langgraphjs/blob/main/libs/sdk-react/docs/v1-migration.md#7-subagents--subgraphs). See also [subagents.md](https://github.com/langchain-ai/langgraphjs/blob/main/libs/sdk-react/docs/subagents.md), [frontend overview](https://docs.langchain.com/oss/javascript/langchain/frontend/overview), [streaming cookbook](https://github.com/langchain-ai/streaming-cookbook), and [examples/ui-react](https://github.com/langchain-ai/langgraphjs/tree/main/examples/ui-react).

PR #2417, opened 2026-05-20
feat: blacklist compromised axios versions 1.14.1 and 0.30.4 (#2257)

neuralmanacle:
# Description
Addresses a supply-chain vulnerability by explicitly blacklisting axios versions 1.14.1 and 0.30.4 across the dependency tree using explicit logical version constraints.

# Changes
* **package.json**: Updated `pnpm.overrides.axios` to exclude both versions while maintaining the `>=1.15.0` safety threshold.
* **pnpm-lock.yaml**: Regenerated to lock in the new constraints and verify graph integrity.

PR #2456, opened 2026-06-01
chore: blacklist axios versions 1.14.1 and 0.30.4

neuralmanacle:
Axios >=1.15.0

Fixes # (issue)


christian-bromann:
As mentioned in https://github.com/langchain-ai/langgraphjs/pull/2417#issuecomment-4589455556 I don't think we need should expand the axios range beyoned >=1.15.0.

PR #2461, opened 2026-06-01
feat(langgraph): add `StateGraph.setNodeDefaults()` for graph-wide node policy defaults

christian-bromann:
## Summary
- Add `StateGraph.setNodeDefaults({ retryPolicy?, cachePolicy? })` for graph-wide node policy defaults, resolved at `compile()` so call order does not matter; per-node `addNode` options always take precedence.
- Defaults are not inherited by subgraphs (parity with Python #7747).
- Consolidate duplicate `retryPolicy` / `cachePolicy` fields into exported `NodePolicyOptions` with shared TSDoc; `NodeDefaults` and `StateGraphAddNodeOptions` both use it; internal specs use `ResolvedNodePolicies` after boolean `cachePolicy` normalization.

### DIFF
--- a/libs/langgraph-core/src/graph/state.ts
+++ b/libs/langgraph-core/src/graph/state.ts
@@ -110,8 +110,9 @@ export interface StateGraphArgs<Channels extends object | unknown> {
  *
  * Use with {@link StateGraph.addNode} for per-node overrides, or with
  * {@link StateGraph.setNodeDefaults} for graph-wide defaults applied at
- * `compile()` time. Per-node values always take precedence over defaults,
- * regardless of call order.
+ * `compile()` time. Per-node values always take precedence over defaults.
+ * `setNodeDefaults` may be called before or after `addNode`, including as the
+ * last step before `compile()`.
  */
 export type NodePolicyOptions = {
   /**
@@ -143,9 +144,8 @@ export type NodePolicyOptions = {
  */
 export type NodePolicies = {
   retryPolicy?: RetryPolicy;
-  cachePolicy?: CachePolicy;
-  /** Set when `addNode(..., { cachePolicy: false })` opts out of graph defaults. */
-  cachePolicyOptOut?: true;
+  /** `false` opts out of graph defaults set via {@link StateGraph.setNodeDefaults}. */
+  cachePolicy?: CachePolicy | false;
 };
 
 export type StateGraphNodeSpec<RunInput, RunOutput> = NodeSpec<
@@ -171,17 +171,6 @@ export type StateGraphAddNodeOptions<
   input?: InputSchema;
 } & NodePolicyOptions &
   AddNodeOptions<Nodes>;
-
-/**
- * Graph-wide default node policies set via {@link StateGraph.setNodeDefaults}.
- *
- * These are applied to every node at `compile()` time. Per-node values passed
- * to `addNode(...)` always take precedence, regardless of the order in which
- * `setNodeDefaults` and `addNode` are called.
- *
- * Defaults are **not** inherited by subgraphs.
- */
-export type NodeDefaults = NodePolicyOptions;
 
 export type StateGraphArgsWithStateSchema<
   SD extends StateDefinition,
@@ -613,17 +602,17 @@ export class StateGraph<
    * graph.
    *
    * Per-node values passed to {@link addNode} always take precedence over these
-   * defaults. Defaults are resolved at {@link compile} time, so this method can
-   * be called before or after `addNode` with the same result. Calling it
-   * multiple times merges the provided fields, with later calls overriding
-   * earlier ones on a per-field basis.
+   * defaults. Defaults are resolved at {@link compile} time, so call order does
+   * not matter — you may call this before or after `addNode`, including as the
+   * last step before `compile()`. Calling it multiple times merges the provided
+   * fields, with later calls overriding earlier ones on a per-field basis.
    *
    * Policies set here are **not** inherited by subgraphs.
    *
    * @param defaults - The default node policies to apply.
    * @returns The builder instance, for chaining.
    *
-   * @example
+   * @example Call before `addNode`
    * ```ts
    * const graph = new StateGraph(State)
    *   .setNodeDefaults({
@@ -635,8 +624,21 @@ export class StateGraph<
    *   .addEdge(START, "a")
    *   .compile();
    * ```
+   *
+   * @example Call after `addNode`, immediately before `compile()`
+   * ```ts
+   * const graph = new StateGraph(State)
+   *   .addNode("a", nodeA)
+   *   .addNode("b", nodeB, { retryPolicy: { maxAttempts: 5 } }) // overrides default
+   *   .addEdge(START, "a")
+   *   .setNodeDefaults({
+   *     retryPolicy: { maxAttempts: 3 },
+   *     cachePolicy: { ttl: 60 },
+   *   })
+   *   .compile();
+   * ```
    */
-  setNodeDefaults(defaults: NodeDefaults): this {
+  setNodeDefaults(defaults: NodePolicyOptions): this {
     if (defaults.retryPolicy !== undefined) {
       this._nodeDefaults.retryPolicy = defaults.retryPolicy;
     }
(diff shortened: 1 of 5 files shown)


libs/langgraph-core/src/graph/state.ts
```diff
@@ -105,15 +105,55 @@ export interface StateGraphArgs<Channels extends object | unknown> {
...
+   *
+   * - Pass a {@link CachePolicy} object for fine-grained control (e.g. custom
+   *   `keyFunc`, `ttl`).
+   * - Pass `true` to enable caching with default settings.
+   * - Pass `false` to disable caching.
+   *
+   * @see {@link CachePolicy}
+   */
+  cachePolicy?: CachePolicy | boolean;
+};
```
hntrl, line 137: are we adding timeout policy? follow-on item after other PRs are merged?

libs/langgraph-core/src/graph/state.ts
```diff
@@ -105,15 +105,55 @@ export interface StateGraphArgs<Channels extends object | unknown> {
...
+
+/**
+ * Resolved retry and cache policies stored on a node after boolean
+ * `cachePolicy` shorthand is normalized.
+ *
+ * @internal
+ */
+export type ResolvedNodePolicies = {
```
hntrl: nit: can we just call these NodePolicies?

libs/langgraph-core/src/graph/state.ts
```diff
@@ -550,6 +606,49 @@ export class StateGraph<
...
+   */
+  setNodeDefaults(defaults: NodeDefaults): this {
+    if (defaults.retryPolicy !== undefined) {
+      this._nodeDefaults.retryPolicy = defaults.retryPolicy;
+    }
+    if (defaults.cachePolicy !== undefined) {
+      this._nodeDefaults.cachePolicy =
+        typeof defaults.cachePolicy === "boolean"
+          ? defaults.cachePolicy
+            ? {}
```
hntrl, line 649: whats the behavior of a blank cache policy?

libs/langgraph-core/src/graph/state.ts
```diff
@@ -1168,10 +1267,24 @@ export class StateGraph<
...
+      nodeDefaults.cachePolicy !== undefined;
     for (const [key, node] of Object.entries<StateGraphNodeSpec<S, U>>(
       this.nodes
     )) {
-      compiled.attachNode(key as N, node);
+      const resolvedNode = hasNodeDefaults
+        ? {
+            ...node,
+            retryPolicy: node.retryPolicy ?? nodeDefaults.retryPolicy,
+            cachePolicy: node.cachePolicy ?? nodeDefaults.cachePolicy,
```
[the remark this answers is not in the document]
hntrl: open-swe seems valid

libs/langgraph-core/src/graph/state.ts
christian-bromann: Yep

libs/langgraph-core/src/graph/state.ts
christian-bromann: correct: `setNodeDefaults({ cachePolicy: true })` yields `nodes.a.cachePolicy === {}`

libs/langgraph-core/src/graph/state.ts
christian-bromann: fixed

christian-bromann:
@hntrl thanks for the review, applied feedback

libs/langgraph-core/src/graph/state.ts
hntrl: right but what does a blank cache policy do? Is it something worth making more explicit?

libs/langgraph-core/src/graph/state.ts
```diff
@@ -105,15 +105,57 @@ export interface StateGraphArgs<Channels extends object | unknown> {
...
+ * Resolved retry and cache policies stored on a node after boolean
+ * `cachePolicy` shorthand is normalized.
+ *
+ * @internal
+ */
+export type NodePolicies = {
   retryPolicy?: RetryPolicy;
   cachePolicy?: CachePolicy;
+  /** Set when `addNode(..., { cachePolicy: false })` opts out of graph defaults. */
+  cachePolicyOptOut?: true;
```
hntrl: maybe just make cachePolicy be `CachePolicy | false`? This seems very strange

libs/langgraph-core/src/graph/state.ts
```diff
@@ -126,10 +168,20 @@ export type StateGraphAddNodeOptions<
...
+/**
+ * Graph-wide default node policies set via {@link StateGraph.setNodeDefaults}.
+ *
+ * These are applied to every node at `compile()` time. Per-node values passed
+ * to `addNode(...)` always take precedence, regardless of the order in which
+ * `setNodeDefaults` and `addNode` are called.
+ *
+ * Defaults are **not** inherited by subgraphs.
+ */
+export type NodeDefaults = NodePolicyOptions;
```
hntrl: why do we alias here?

libs/langgraph-core/src/graph/state.ts
```diff
@@ -903,14 +1004,21 @@ export class StateGraph<
...
       let cachePolicy = options?.cachePolicy;
+      let cachePolicyOptOut: true | undefined;
       if (typeof cachePolicy === "boolean") {
-        cachePolicy = cachePolicy ? {} : undefined;
+        if (cachePolicy) {
+          cachePolicy = {};
+        } else {
+          cachePolicy = undefined;
+          cachePolicyOptOut = true;
+        }
```
hntrl: amended if we make updates to NodePolicy right?

libs/langgraph-core/src/graph/state.ts
```diff
@@ -1168,10 +1276,26 @@ export class StateGraph<
...
-      compiled.attachNode(key as N, node);
+      const resolvedNode = hasNodeDefaults
+        ? {
+            ...node,
+            retryPolicy: node.retryPolicy ?? nodeDefaults.retryPolicy,
+            cachePolicy: node.cachePolicyOptOut
+              ? undefined
+              : (node.cachePolicy ?? nodeDefaults.cachePolicy),
+          }
+        : node;
```
hntrl, line 1299: the hasNodeDefaults indirection seems kinda weird?

```suggestion
      const resolvedNode = {
        ...node,
        retryPolicy: node.retryPolicy ?? nodeDefaults.retryPolicy,
        cachePolicy: node.cachePolicy ?? nodeDefaults.cachePolicy,
      };
```

libs/langgraph-core/src/tests/set_node_defaults.test.ts
hntrl, line 1: Can you add a test where setNodeDefaults is at the end of the builder chain? I get a hunch there might be some weird behavior since we handle graph construction in a declarative way.

```ts
new StateGraph(schema)
  .addNode()
  .setNodeDefaults()
  .compile()
```

libs/langgraph-core/src/graph/state.ts
christian-bromann: In the docs we say:

> Retries are opt-in. A node retries only when it has a `retryPolicy` configured, either directly or through graph defaults with [`setNodeDefaults`](#graph-defaults). An empty policy (`{}`) is enough. Without a policy, the first failure ends the attempt and LangGraph does not call `retryOn`.

If retries are enabled, we use default retry policies (also documented).

libs/langgraph-core/src/graph/state.ts
christian-bromann: Originally to be more explicit in the documentation bc one was used in `addNode` the other in `setNodeDefaultOptions` but can merge

libs/langgraph-core/src/graph/state.ts
christian-bromann: We have to check whether `cachePolicy` is explicitly set to `false` to opt out of the default policy.

libs/langgraph-core/src/tests/set_node_defaults.test.ts
christian-bromann: > Can you add a test where setNodeDefaults is at the end of the builder chain?

We had such a test but will make the docs more explicit to highlight that it doesn't matter where the `setNodeDefaults` function is called in the chain.

libs/langgraph-core/src/graph/state.ts
christian-bromann: applied fix

hntrl (approved):

PR #2464, opened 2026-06-02
docs: import messagesStateReducer in low_level.md example

emmayusufu:
The `messagesStateReducer` example on the Low Level concepts page uses `messagesStateReducer` as its reducer, but the import only brings in `Annotation` and the `Messages` type. Copy-pasting it as-is fails to compile with `Cannot find name 'messagesStateReducer'`. This adds it to the import. The equivalent example just below already imports it the same way.


christian-bromann (approved):
Thank you 👍 

Please note that these docs are outdated. We recommend to check out our new docs at https://docs.langchain.com/oss/javascript/langgraph/overview

PR #2491, opened 2026-06-04
test: add managed-runtime integration suite + per-PR workflow

nick-hollon-lc:
## What

Ports the `langgraph-api` JS integration framework (the suite under `api/langgraph_api/js/tests`) into this repo at `integration/`, and wires it into per-PR CI via `.github/workflows/test-integration-managed.yml`.

It runs **JS graphs on the real managed `langgraph-api` server** (Go core + Python orchestrator + Node subprocess), driven by `@langchain/langgraph-sdk` over HTTP, using the **in-repo** `@langchain/langgraph` (packed from the PR, not npm).

## Why

The existing JS tests only exercise the in-repo JS server (`@langchain/langgraph-api`), which strips the raw `StreamChunk` 4th element before the wire. The managed runtime does not: it unpacks the raw tuple strictly (`stream.py:324`, `ns, mode, chunk = event["data"]["chunk"]`). That gap let PR #2314 (which widened `StreamChunk` from a 3-tuple to a 4-tuple) ship green and break streaming for JS graphs on the managed runtime with `ValueError: too many values to unpack (expected 3)`. This suite closes that gap from inside the JS repo.

## How it runs

`pnpm pack` the in-repo `@langchain/langgraph` into the graphs deployment, `langgraph build` a service image on the published `langchain/langgraphjs-api:20` base, `docker compose up` (Postgres + Redis + server on `:9123`), then `vitest`.

## Gates (intentional)

- **Path-filtered** to `libs/langgraph-core/**`, `libs/langgraph-api/**`, `libs/sdk/**`, `integration/**` (it is a few-minute job: pulls a large base image and builds).
- **Skipped on fork PRs**: repo secrets are not exposed to forks, so it runs only for same-repo branches. It does **not** use `pull_request_target`.

## Required before this check can pass

Add a repository secret `LANGSMITH_API_KEY` (or `LANGGRAPH_CLOUD_LICENSE_KEY`); the managed image needs it to start. After the first green run, add the check to branch protection.

## Sequencing note

With current `main`, the `subgraphs` streaming tests in this suite **reproduce #2314 and fail by design** (that is the proof the guard works). They go green once the `StreamChunk` 3-tuple fix lands (separate PR, already in progress). The fast in-process arity guard (`pregel.stream.test.ts`) ships with that fix and runs on every PR.


integration/graphs/agent_simple.ts
```diff
@@ -0,0 +1,113 @@
...
+  const cached: Record<string, FakeListChatModel> = {};
+  return (threadId: string) => {
+    cached[threadId] ??= new FakeListChatModel({
+      responses: ["begin", "end\u2028"],
+    });
+    return cached[threadId];
+  };
+})();
+
+const AgentState = Annotation.Root({
```
christian-bromann: Let's use `StateSchema` where possible

integration/graphs/configure-langgraph.py
```diff
@@ -0,0 +1,73 @@
...
+            "agent_simple": "./agent_simple.mts:graph",
+            "agent_simple_factory": "./agent_simple.mts:graphFactory",
+            "factory_store": "./factory_store.mts:graph",
+        },
+        "env": ".env",
+        "http": {
+            "app": "./http.mts:app",
+            "configurable_headers": {"includes": ["x-configurable-header"]},
+        },
+        "ui": {"agent-alias": "./agent.ui.tsx"},
```
christian-bromann, line 55: it doesn't seem like we need this for the tests?

integration/graphs/agent.ui.tsx
```diff
@@ -0,0 +1,10 @@
+import "./agent.css";
+import React from "react";
+
+export function StockPrice() {
```
christian-bromann, line 4: including this file

integration/README.md
```diff
@@ -0,0 +1,67 @@
...
+Ported from `langgraph-api/api/langgraph_api/js/tests`. This is the same
+integration framework the Python `langgraph-api` repo uses to run **JS graphs on
+the real managed server** (Go core + Python orchestrator + Node subprocess via
+`RemotePregel`), driven by the `@langchain/langgraph-sdk` client over HTTP.
+
+It exists so a JS change that breaks the managed runtime gets caught here —
+e.g. PR #2314 widened the raw `StreamChunk` to a 4-tuple, which the Python
+orchestrator unpacks strictly (`stream.py:324  ns, mode, chunk = ...`) and throws
+`too many values to unpack`. The in-repo JS server strips that 4th element, so
+only this managed-runtime path surfaces the regression.
```
christian-bromann: Let's have a common documentation rather than pointing to a specific issue/bug

integration/streaming.test.ts
```diff
@@ -0,0 +1,584 @@
...
+  await sql`DELETE FROM store`;
+  await sql`DELETE FROM checkpoints`;
+  await sql`DELETE FROM assistant WHERE metadata->>'created_by' is null OR metadata->>'created_by' != 'system'`;
+});
+
+/**
+ * Helper to parse SSE chunks from a raw ReadableStream.
+ * Returns an array of { event, data } objects.
+ */
+async function readSSEStream(
```
christian-bromann: let's move into `utils.ts`

integration/graphs/agent.css
```diff
@@ -0,0 +1 @@
+@import "tailwindcss";
```
christian-bromann, line 1: needed?

integration/background.test.ts
```diff
@@ -0,0 +1,287 @@
...
+};
+
+beforeAll(async () => {
+  await sql`DELETE FROM thread`;
+  await sql`DELETE FROM store`;
+  await sql`DELETE FROM checkpoints`;
+});
+
+describe("background run lifecycle", () => {
+  it.concurrent("create background run", { retry: 3 }, async () => {
```
christian-bromann: Are these retries needed? I wouldn't add them if not.
```suggestion
  it.concurrent("create background run", async () => {
```

integration/compose-postgres.yml
```diff
@@ -0,0 +1,52 @@
...
+      start_period: 10s
+    env_file: ./graphs/.env
+    environment:
+      REDIS_URI: redis://langgraph-redis:6379
+      DATABASE_URI: postgres://postgres:postgres@langgraph-postgres:5432/postgres?sslmode=disable
+      N_JOBS_PER_WORKER: "5"
+      LANGGRAPH_CLOUD_LICENSE_KEY: ${LANGGRAPH_CLOUD_LICENSE_KEY:-}
+      LANGSMITH_API_KEY: ${LANGSMITH_API_KEY:-}
+      LANGSMITH_TRACING: "false"
+      FF_JS_ZEROMQ_ENABLED: ${FF_JS_ZEROMQ_ENABLED:-}
```
christian-bromann: Is this needed for this test scenario?

integration/graphs/package.json
```diff
@@ -0,0 +1,14 @@
...
+    "@langchain/langgraph": "latest",
+    "@langchain/langgraph-sdk": "latest",
+    "zod": "^4.3.6",
+    "jose": "^6.2.2",
+    "hono": "^4.12.18"
+  },
+  "devDependencies": {
+    "tailwindcss": "^4.2.4"
+  }
+}
```
christian-bromann, line 15: I don't think we need this package.json, if so we should define these dependencies within the `/integration/package.json`

integration/package.json
```diff
@@ -0,0 +1,20 @@
...
+    "configure": "python3 graphs/configure-langgraph.py --node-version 20 --output graphs/langgraph.json",
+    "build": "cd graphs && uv run langgraph build --no-pull -c langgraph.json --tag langgraphjs-tests-api",
+    "up": "docker compose -f compose-postgres.yml up -V --force-recreate --wait",
+    "down": "docker compose -f compose-postgres.yml down -v",
+    "test": "vitest run --fileParallelism=false",
+    "test:streaming": "vitest run --fileParallelism=false streaming.test.mts",
+    "int": "pnpm configure && pnpm build && pnpm up && (pnpm test; ec=$?; pnpm down; exit $ec)"
+  },
+  "devDependencies": {
+    "@langchain/langgraph-sdk": "latest",
```
christian-bromann: We should refer to the langgraph version in this workspace no?
```suggestion
    "@langchain/langgraph-sdk": "workspace:*",
```

christian-bromann (commented):
Some comments.

Can we rename all `*.mts files to just `*.ts`?

nick-hollon-lc:
**Out of scope for this PR.** This PR only adds the `integration/` test suite and the managed-runtime workflow — it doesn't modify anything under `libs/`. The `libs/langgraph-api/src/stream.mts` checkpoints finding is in code that came from `main` (pulled in when `main` was merged into this branch, which the reviewer diffed against), as the "out-of-diff finding" label notes. Tracking the checkpoints-stream question separately; no change here.

integration/graphs/configure-langgraph.py
nick-hollon-lc: The ui registration is needed: api.test.ts's "generative ui" test (client["~ui"].getComponent("agent-alias", "weather-component")) asserts on agent-alias's entrypoint.css / entrypoint.js

integration/graphs/agent.ui.tsx
nick-hollon-lc: agent.ui.tsx backs the agent-alias UI component that api.test.ts's "generative ui" test exercises end to end (it fetches the built entrypoint.js and checks the bundle)

integration/graphs/agent.css
nick-hollon-lc: agent.css is imported by agent.ui.tsx so if we keep that we need this

integration/graphs/package.json
nick-hollon-lc: graphs/ is a separate deployment that langgraph build packs into the Docker image (langgraph.json has "dependencies": ["."]), so it needs its own manifest for the container-runtime deps (@langchain/langgraph, zod, hono, jose). those are distinct from the test-runner deps in integration/package.json (vitest, postgres, the SDK client), and the CI workflow rewrites this file independently to inject the in-repo @langchain/langgraph tarball.

PR #2503, opened 2026-06-07
fix(langgraph-checkpoint-validation): align node engine with vitest

jkennedyvz:
## Summary
- align checkpoint validation package Node engine with Vitest 4 runtime requirements
- document the CLI Node requirement
- add missing secret file patterns to .gitignore per commit safety preflight

## Testing
- Not run per request


christian-bromann (approved):
LGTM 👍 

A node.js bump usually requires a major release but I think we are okay to ship this as minor bump.

PR #2186, opened 2026-03-14
fix(langgraph-checkpoint-mongodb): MongoDB checkpointer metadata filter

jackjin1997:
## Summary
- The `list()` method's metadata filter queried `metadata.${key}` against a serialized binary blob, making it silently nonfunctional (dead code)
- Added a plain JSON `metadata_search` field alongside the serialized `metadata` in `put()`, and updated `list()` to query against it
- This is consistent with how Postgres (native JSONB `@>`) and SQLite (`jsonb(CAST(...))`) handle metadata filtering

## Test plan
- [x] Existing unit tests pass (6/6)
- [ ] Integration test with real MongoDB to verify filter works against `metadata_search`

Fixes #1591

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2504, opened 2026-06-08
fix(langgraph-checkpoint-sqlite): allow filter on arbitrary metadata keys in list()

jackjin1997:
## Summary

`SqliteSaver.list({}, { filter })` was silently dropping any filter key that wasn't `source`, `step`, or `parents` — returning the full unfiltered result set instead. The other checkpointers (MongoDB, Postgres, Redis) all accept arbitrary metadata keys, so SQLite was the only odd one out.

### Repro (failing on `main`)

```ts
await saver.put({ configurable: { thread_id: "1" } }, cp, { source: "update", step: 0, parents: {}, tenant_id: "acme" });
await saver.put({ configurable: { thread_id: "2" } }, cp, { source: "update", step: 0, parents: {}, tenant_id: "globex" });

// expected: 1 result (only thread 1)
// actual on main: 2 results (filter silently ignored)
for await (const t of saver.list({}, { filter: { tenant_id: "acme" } })) { ... }
```

### Fix

- Drop the `validCheckpointMetadataKeys` allow-list (`["source", "step", "parents"]`).
- Pass the JSON path as a bound parameter (`jsonb(...)->? = ?`) so arbitrary user keys flow through safely without SQL-injection risk. Validated `better-sqlite3` supports parameterized JSON path on the `->` operator.

### Tests

New test `should filter on arbitrary metadata keys, not just CheckpointMetadata keys` exercises three custom keys (`tenant_id`, `env`, missing) plus combined filters. **Reverse-verified**: with the fix reverted the new test fails with `expected 2, actual 3` (filter silently no-ops).

### Consistency parity

| Checkpointer | Arbitrary metadata filter | Before this PR |
|--------------|---------------------------|----------------|
| MongoDB | ✅ via `metadata_search.{key}` (#2186) | ✅ |
| Postgres | ✅ via `metadata @> $1` | ✅ |
| Redis | ✅ via RediSearch tags | ✅ |
| **SQLite** | ✅ via `jsonb(metadata)->? = ?` (this PR) | ❌ |

## AI Disclosure
This bug was identified through cross-checkpointer code review with AI assistance, and the fix was validated against the existing test suite plus a new regression test.

### DIFF
--- a/libs/checkpoint-sqlite/src/tests/checkpoints.test.ts
+++ b/libs/checkpoint-sqlite/src/tests/checkpoints.test.ts
@@ -183,7 +183,8 @@ describe("SqliteSaver", () => {
       return tuples;
     };
 
-    expect((await collect({ tenant_id: "acme" })).length).toBe(2);
+    const ids = (tuples) => tuples.map(t => t.config.configurable?.thread_id).sort();
+    expect(ids(await collect({ tenant_id: "acme" }))).toEqual(["1", "2"]);
     expect((await collect({ env: "prod" })).length).toBe(2);
     expect((await collect({ tenant_id: "acme", env: "prod" })).length).toBe(1);
     expect((await collect({ tenant_id: "missing" })).length).toBe(0);
--- a/libs/checkpoint-sqlite/src/tests/checkpoints.test.ts
+++ b/libs/checkpoint-sqlite/src/tests/checkpoints.test.ts
@@ -147,6 +147,48 @@ describe("SqliteSaver", () => {
     expect(checkpointTuple1.checkpoint.ts).toBe("2024-04-20T17:19:07.952Z");
   });
 
+  it("should filter on arbitrary metadata keys, not just CheckpointMetadata keys", async () => {
+    const sqliteSaver = SqliteSaver.fromConnString(":memory:");
+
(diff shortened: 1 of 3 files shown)


libs/checkpoint-sqlite/src/tests/checkpoints.test.ts
```diff
@@ -147,6 +147,48 @@ describe("SqliteSaver", () => {
...
+      filter: Record<string, unknown>
+    ): Promise<CheckpointTuple[]> => {
+      const tuples: CheckpointTuple[] = [];
+      for await (const t of sqliteSaver.list({} as RunnableConfig, { filter })) {
+        tuples.push(t);
+      }
+      return tuples;
+    };
+
+    expect((await collect({ tenant_id: "acme" })).length).toBe(2);
```
christian-bromann:
```suggestion
    const ids = (tuples) => tuples.map(t => t.config.configurable?.thread_id).sort();
    expect(ids(await collect({ tenant_id: "acme" }))).toEqual(["1", "2"]);
```

christian-bromann (dismissed):
Minor test improvement, otherwise LGTM 👍

christian-bromann (approved):
👍

PR #2512, opened 2026-06-09
fix(langgraph-checkpoint-postgres): reject SQL LIKE wildcards in PostgresStore namespace labels

jackjin1997:
## Summary

`BaseStore.search()` on `PostgresStore` matches namespaces via `namespace_path LIKE \${prefix}%` (bound parameter). The bind protects against SQL injection but **does not** keep Postgres from interpreting `%`, `_`, or `\` inside the label itself as `LIKE` wildcards / escape characters. So a caller-supplied namespace prefix containing `%` silently widens the prefix into a glob — data leaks across namespaces in any multi-tenant deployment that pipes configurable values into a `BaseStore` namespace.

`InMemoryStore` is unaffected — it compares prefixes via literal `String.startsWith()`.

### Repro (conceptually, against the old `validateNamespace`)

```ts
const store = PostgresStore.fromConnString(url);

// tenant A writes private data
await store.put(["acme", "users", "alice"], "preferences", { theme: "dark" });

// tenant B searches under their own namespace — fine
await store.search(["other-tenant"]);

// but B can also do this, and the search returns A's row:
await store.search(["%"]);            // matches every namespace
await store.search(["a%", "users"]);  // matches "acme:users", "any:users", ...
```

After this PR `validateNamespace` rejects `%` / `_` / `\` at all `search` / `get` / `put` / `delete` entrypoints, matching the existing handling of `.` and empty labels.

### Tests

New `src/store/modules/utils.test.ts` exercises `validateNamespace` directly (it's a pure function, so no live Postgres needed). 8 new cases cover each wildcard char in isolation, in combination with other letters, and at non-first positions; benign characters that *look* similar (`-`, `:`, digits, unicode) are explicitly allowed.

**Reverse-verified**: with this PR's `utils.ts` reverted, 7 of the 7 new LIKE-wildcard tests fail (`expected [Function] to throw an error / Received undefined`).

### Consistency

| Store | Namespace prefix match | Wildcard handling |
|-------|------------------------|--------------------|
| `InMemoryStore` | `String.startsWith()` (literal) | n/a — already safe |
| `PostgresStore` | `namespace_path LIKE \${prefix}%` | wildcards rejected at validation (this PR) |
| `RedisStore` (precedent) | `KEYS namespace:*:...` | already rejects `*`/`?`/`[`/`]`/`\` via `assertSafeKeyComponent` |

CWE-1336 (Improper Neutralization of Special Elements Used in a Template Engine).

## AI Disclosure

This issue was identified via cross-store consistency review with AI assistance, and the fix was validated against a focused unit-test suite plus reverse-verification (test suite fails without the fix).

### DIFF
--- a/.changeset/fix-postgres-store-namespace-like-wildcard.md
+++ b/.changeset/fix-postgres-store-namespace-like-wildcard.md
@@ -1,0 +1,5 @@
+---
+"@langchain/langgraph-checkpoint-postgres": patch
+---
+
+fix: reject SQL `LIKE` wildcards (`%`, `_`) and the backslash escape character in `PostgresStore` namespace labels. `BaseStore.search()` matches namespaces via `namespace_path LIKE ${prefix}%`, and these characters in caller-supplied namespace labels are interpreted as wildcards by Postgres even through a bound parameter — letting a namespace prefix of `["%"]` match every namespace in the store across tenants. `validateNamespace` now throws for these characters at all `search` / `get` / `put` entrypoints, keeping store-wide consistency. CWE-1336.
--- a/libs/checkpoint-postgres/src/store/modules/utils.test.ts
+++ b/libs/checkpoint-postgres/src/store/modules/utils.test.ts
@@ -1,0 +1,56 @@
+import { describe, it, expect } from "vitest";
+import { validateNamespace } from "./utils.js";
+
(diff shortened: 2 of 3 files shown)


christian-bromann (approved):
Thank you 👍

PR #2516, opened 2026-06-10
fix(langgraph-checkpoint-sqlite): honor WRITES_IDX_MAP for special channels in putWrites

jackjin1997:
## Summary

`SqliteSaver.putWrites` stored every write at the call-local ordinal `idx`, ignoring `WRITES_IDX_MAP`. So when an `INTERRUPT` (or `ERROR` / `SCHEDULED` / `RESUME`) followed a regular write for the same task, both landed at `idx=0` and the table PK `(thread_id, checkpoint_ns, checkpoint_id, task_id, idx)` collided — `INSERT OR REPLACE` silently dropped the regular write.

Other checkpointer implementations (Memory in `libs/checkpoint`, Postgres) and the **Python** SQLite checkpointer (`langgraph/checkpoint/sqlite/__init__.py`) all pin special channels to fixed negative indices via `WRITES_IDX_MAP` so they can't collide with per-step regular writes.

### Cross-impl parity before this PR

|---|---|---|
| Python SQLite | ✅ | `OR REPLACE` vs `OR IGNORE` based on `all special` |
| **TS SQLite** | ❌ | always `OR REPLACE` |

### Fix

1. `idx` resolves to `WRITES_IDX_MAP[channel] ?? idx`.
2. Statement switches between `INSERT OR REPLACE` and `INSERT OR IGNORE` depending on whether every write targets a special channel, matching `langgraph/checkpoint/sqlite/__init__.py`. INTERRUPT→RESUME state transitions overwrite (both special); concurrent tasks can't clobber each other's regular writes at the same idx.

### Tests

New regression test interleaves regular and special-channel writes across two tasks (`task_A` does regular writes → INTERRUPT → RESUME, `task_B` does a regular write). All five logical writes survive `getTuple().pendingWrites`.

**Reverse-verified**: without the fix, `task_A:foo` and `task_A:__interrupt__` both disappear from the round-trip — the test fails with `Set { task_A:__resume__, task_A:bar, task_B:baz }`.

## AI Disclosure

This bug was identified via cross-implementation review against the Python SQLite checkpointer and TS `MemorySaver` / `PostgresSaver`, with AI assistance. The fix mirrors the Python reference implementation.

### DIFF
--- a/.changeset/fix-sqlite-checkpoint-put-writes-special-channel-idx.md
+++ b/.changeset/fix-sqlite-checkpoint-put-writes-special-channel-idx.md
@@ -1,0 +1,5 @@
+---
+"@langchain/langgraph-checkpoint-sqlite": patch
+---
+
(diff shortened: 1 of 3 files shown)


christian-bromann (approved):
Thanks a lot 👍

PR #2517, opened 2026-06-10
fix(langgraph-checkpoint-mongodb): honor WRITES_IDX_MAP for special channels in putWrites

jackjin1997:
## Summary

Follow-up to #2516 (which fixed the same bug for `SqliteSaver`). `MongoDBSaver.putWrites` stored every write at the call-local ordinal `idx` and always used `$set`, ignoring `WRITES_IDX_MAP`. So:

- A follow-up `putWrites([[INTERRUPT, value]], taskId)` for the same checkpoint computed `idx=0` and collided with the task's first regular write — the unconditional `$set` upsert silently overwrote whichever row landed there first.
- A mixed call `putWrites([[foo, …], [bar, …], [INTERRUPT, …]], taskId)` shifted the INTERRUPT to idx=2, where a peer task storing a regular write at the same idx for the same task_id would clobber it.

The fix mirrors `BaseCheckpointSaver.put_writes` semantics and the Postgres / SQLite (TS and Python) implementations.

### Cross-impl parity after this PR

|---|---|---|
| SQLite (TS, #2516) | ✅ | `OR REPLACE` vs `OR IGNORE` based on `all special` |
| **MongoDB (TS, this PR)** | ✅ | `$set` vs `$setOnInsert` based on `all special` |
| Redis (TS) | ❌ | follow-up needed |

### Fix

1. `idx` resolves to `WRITES_IDX_MAP[channel] ?? idx`, so ERROR / SCHEDULED / INTERRUPT / RESUME land at fixed negative indices that can't collide with per-step regular writes.
2. The update operator switches between `$set` and `$setOnInsert` depending on whether every write targets a special channel. State transitions like INTERRUPT → RESUME still overwrite (both special); a regular write from one task can never silently overwrite another concurrent task's regular write at the same idx.

When timestamps are enabled, the `$setOnInsert` path also defers `upserted_at` to insert-only, so a no-op upsert against a peer task's existing row doesn't bump that row's "last modified" stamp.

### Tests

Mock-based unit test (no real MongoDB needed) captures `bulkWrite` operations and asserts:
- Mixed regular + INTERRUPT batch → indices `[-3, 0, 1]` (vs the pre-fix `[0, 1, 2]`) and every op uses `$setOnInsert`.
- Special-only batch (RESUME) → idx `-4` and uses `$set`.

**Reverse-verified**: with the fix reverted the test fails with `Received [0, 1, 2]`.

## AI Disclosure

Identified via cross-implementation review against the SQLite fix in #2516 — same root cause class, separate test surface.

### DIFF
--- a/.changeset/fix-mongodb-checkpoint-put-writes-special-channel-idx.md
+++ b/.changeset/fix-mongodb-checkpoint-put-writes-special-channel-idx.md
@@ -1,0 +1,5 @@
+---
+"@langchain/langgraph-checkpoint-mongodb": patch
+---
+
+fix: `MongoDBSaver.putWrites` now honors `WRITES_IDX_MAP`, pinning special channels (`__error__`, `__scheduled__`, `__interrupt__`, `__resume__`) to fixed negative indices instead of the call-local ordinal. Previously a mixed `putWrites([[...regular...], [INTERRUPT, …]], taskId)` placed the INTERRUPT at a positive idx that could collide with a regular write at the same `(task_id, idx)`, and the unconditional `$set` upsert silently overwrote whichever row landed there first. The conflict-resolution clause now matches the Postgres / SQLite (TS and Python) checkpointers: `$set` only when every channel is a special one, `$setOnInsert` otherwise.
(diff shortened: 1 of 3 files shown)


christian-bromann (approved):
LGTM 👍

PR #2518, opened 2026-06-10
fix(langgraph-checkpoint-redis): honor WRITES_IDX_MAP for special channels in putWrites

jackjin1997:
## Summary

Third in the series following #2516 (SQLite) and #2517 (MongoDB). `RedisSaver.putWrites` stored each write at the call-local ordinal `idx` — the key suffix in `checkpoint_write:<thread>:<ns>:<ckpt>:<task>:<idx>` was just the loop index — and always used unguarded `JSON.SET`, ignoring `WRITES_IDX_MAP`.

- A follow-up `putWrites([[INTERRUPT, value]], taskId)` for the same checkpoint computed `idx=0` and collided with the task's first regular write key; the unconditional `JSON.SET` silently overwrote whichever row landed there first.
- A mixed call like `[[foo, …], [bar, …], [INTERRUPT, …]]` shifted the INTERRUPT to idx=2, where a peer task storing a regular write at the same idx for the same task_id would clobber it.

### Cross-impl parity after this PR

|---|---|---|
| SQLite (TS, #2516) | ✅ | `OR REPLACE` vs `OR IGNORE` based on `all special` |
| MongoDB (TS, #2517) | ✅ | `$set` vs `$setOnInsert` based on `all special` |
| **Redis (TS, this PR)** | ✅ | `JSON.SET` vs `JSON.SET … NX` based on `all special` |

### Fix

1. Key suffix now resolves to `WRITES_IDX_MAP[channel] ?? idx`, so ERROR / SCHEDULED / INTERRUPT / RESUME land at fixed negative indices that can't collide with per-step regular writes.
2. The `JSON.SET` call switches between unguarded and the `NX` modifier depending on whether every write targets a special channel. State transitions like INTERRUPT → RESUME still overwrite (both special); a regular write from one task can never silently overwrite another concurrent task's regular write at the same idx.

### Tests

Mock-based unit test in a new `putWrites.test.ts` (no Redis needed) captures every `client.json.set` call site and asserts:

- Mixed regular + INTERRUPT batch → key tails `[-3, 0, 1]` (vs the pre-fix `[0, 1, 2]`) and every JSON.SET carries `{ NX: true }`.
- Special-only batch (RESUME) → key tail `-4` and no NX (so state transitions can overwrite).

**Reverse-verified**: with the fix reverted both new tests fail.

## AI Disclosure

Identified via cross-implementation review against #2516 and #2517 — same root cause class, different storage layer.

### DIFF
--- a/.changeset/fix-redis-checkpoint-put-writes-special-channel-idx.md
+++ b/.changeset/fix-redis-checkpoint-put-writes-special-channel-idx.md
@@ -1,0 +1,5 @@
+---
+"@langchain/langgraph-checkpoint-redis": patch
+---
+
+fix: `RedisSaver.putWrites` now honors `WRITES_IDX_MAP`, pinning special channels (`__error__`, `__scheduled__`, `__interrupt__`, `__resume__`) to fixed negative indices in their Redis key (`checkpoint_write:…:<idx>`) instead of the call-local ordinal. Previously a mixed `putWrites([[…regular…], [INTERRUPT, …]], taskId)` placed the INTERRUPT key at the positive idx of its position in the batch, where a peer task's regular write at the same idx would overwrite it via the unconditional `JSON.SET`. The conflict-resolution clause now matches Postgres / SQLite / MongoDB: unguarded `JSON.SET` when every write is a special channel, `JSON.SET … NX` (insert-or-ignore) otherwise.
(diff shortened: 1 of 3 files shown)


christian-bromann (approved):
Thanks a ton!

christian-bromann:
> [CI / Environment Tests / exports-esbuild (pull_request)](https://github.com/langchain-ai/langgraphjs/actions/runs/27256292990/job/80491554757?pr=2518)Failing after 57s

Will rerun once possible.

PR #2525, opened 2026-06-11
fix(langgraph-checkpoint-redis): handle undefined pending writes

lhlyu:
## Summary

Fix Redis checkpoint restore when a pending write document has no `value` field.

RedisJSON omits `undefined` values, but `loadPendingWrites` always tried to deserialize:

```ts
JSON.stringify(writeDoc.value)

When value is missing, this produces undefined and can fail with:

JSON Parse error: Unexpected EOF

This breaks interrupt/resume checkpoint loading.

## Changes

- Treat missing pending write value as undefined.
- Apply the fix to both RedisSaver and ShallowRedisSaver.
- Add unit tests for both.

### DIFF
--- a/.changeset/fix-redis-undefined-pending-write-values.md
+++ b/.changeset/fix-redis-undefined-pending-write-values.md
@@ -1,0 +1,5 @@
+---
+"@langchain/langgraph-checkpoint-redis": patch
+---
+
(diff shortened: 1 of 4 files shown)


christian-bromann (approved):
LGTM 👍

PR #2026, opened 2026-03-03
fix(checkpoint-redis): detect existing writes in put to preserve has_writes flag

pawel-twardziak:
Both `RedisSaver.put()` and `ShallowRedisSaver.put()` unconditionally set `has_writes: "false"` when storing a checkpoint document. This is correct for the normal flow (put then putWrites), but breaks the interrupt flow where `putWrites()` is called *before* `put()`.

 In the interrupt flow, `putWrites` stores write data in Redis under a sorted set key (`write_keys_zset:...`). When `put()` runs afterwards and sets `has_writes: "false"`, the writes still exist in Redis but are invisible to `getTuple()` / `list()` because the RediSearch query filters on `has_writes == "true"`. This causes `pendingWrites` (including `__interrupt__` writes) to silently disappear.

The fix checks whether the write keys sorted set already exists in Redis before setting the `has_writes` flag, so pre-existing writes are preserved.

### Changes

- **`RedisSaver.put()`** (`index.ts`): check `client.exists()` on the write keys zset before building the checkpoint document
- **`ShallowRedisSaver.put()`** (`shallow.ts`): same fix
- **Integration tests** (`checkpoint.int.test.ts`): 6 new tests (3 per saver) covering putWrites-before-put, putWrites-after-put, and put-putWrites-put (double-put) scenarios

Fixes #1996

Issue #1996, opened 2026-02-23
RedisSaver.put() overwrites has_writes flag after putWrites() sets it for interrupt checkpoints


jnchaba:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangGraph.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangGraph.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangGraph (or the specific integration package).

### Example Code

The following code requires you run redis-stack in a docker image to reproduce via:

```bash
docker run -d --name redis-repro -p 6379:6379 redis/redis-stack:latest
```

```typescript
/**
 * Minimal reproduction: RedisSaver loses interrupt payload
 *
 * Bug: RedisSaver.put() unconditionally sets has_writes: "false" via JSON.SET,
 * which overwrites the has_writes: "true" that putWrites() sets when persisting
 * __interrupt__ channel data. On subsequent getState(), loadCheckpointWithWrites()
 * skips loading pendingWrites because has_writes === "false", so
 * tasks[].interrupts comes back as [].
 *
 * Expected: getState().tasks[0].interrupts[0].value contains the interrupt payload
 * Actual:   getState().tasks[0].interrupts === []
 *
 * Requires: redis-stack running on localhost:6379
 *   docker run -d --name redis-repro -p 6379:6379 redis/redis-stack:latest
 */

import { RedisSaver } from "@langchain/langgraph-checkpoint-redis";
import { createReactAgent } from "@langchain/langgraph/prebuilt";
import { FakeStreamingChatModel } from "@langchain/core/utils/testing";
import { AIMessageChunk } from "@langchain/core/messages";
import { tool } from "@langchain/core/tools";
import { z } from "zod";
import { createClient } from "redis";

// A dummy tool the agent will try to call
const myTool = tool(async (input) => `result: ${input.query}`, {
  name: "search",
  description: "Search for information",
  schema: z.object({ query: z.string() }),
});

// FakeStreamingChatModel — must use `chunks` (not `responses`) for tool_calls
const fakeModel = new FakeStreamingChatModel({
  chunks: [
    new AIMessageChunk({
      content: "",
      tool_calls: [
        {
          id: "call_1",
          name: "search",
          args: { query: "test" },
        },
      ],
    }),
  ],
});

async function main() {
[...]

PR #2556, opened 2026-06-18
fix(langgraph-checkpoint-mongodb): no-op putWrites on empty writes

mohamedkhaled4053:
## Summary

`MongoDBSaver.putWrites` builds a `bulkWrite` operations array from `writes` and calls `collection.bulkWrite(operations)`. When `writes` is empty, `operations` is empty and the MongoDB driver throws:

```
MongoServerError: Invalid BulkOperation, Batch cannot be empty
```

This is reached in normal operation by **human-in-the-loop / `interrupt()`** flows: when a graph interrupts to wait for input, a task can complete producing **zero channel writes**, so LangGraph calls `putWrites(config, [], taskId)` and the saver throws.

The other first-party savers don't hit this — `checkpoint-postgres` and `checkpoint-sqlite` iterate over `writes`, so an empty array is a natural no-op. The sibling `store.ts` in this same package already guards its bulk write with `if (bulkOps.length > 0)`. `putWrites` was simply missing the equivalent guard.

## Fix

Skip the `bulkWrite` call when there are no operations, mirroring the existing guard in `store.ts`. The guard is at the call site (not an early return), so the existing `configurable` validation is preserved. This aligns the MongoDB saver's empty-writes behavior with the postgres and sqlite savers.

## Tests

- **Unit** (`checkpoints.test.ts`): asserts `putWrites(config, [], taskId)` resolves and that `bulkWrite` is **not** called. Verified to fail before the fix (`bulkWrite` was called with `[]`).
- **Integration** (`checkpoints.int.test.ts`): asserts `putWrites` with empty writes resolves and persists nothing — the truest repro against the real driver.

`pnpm build`, `pnpm lint`, `pnpm format:check`, and the unit suite all pass locally. A changeset is included (`@langchain/langgraph-checkpoint-mongodb` patch).

## Related

Same class of HITL / `interrupt()` + `putWrites` bug previously fixed for the Redis saver: #1996 / #2026.

### DIFF
--- a/.changeset/fix-mongodb-checkpoint-empty-writes.md
+++ b/.changeset/fix-mongodb-checkpoint-empty-writes.md
@@ -1,0 +1,5 @@
+---
+"@langchain/langgraph-checkpoint-mongodb": patch
+---
+
+Fix `MongoDBSaver.putWrites` throwing `MongoServerError: Invalid BulkOperation, Batch cannot be empty` when called with an empty `writes` array. This is reached by human-in-the-loop / `interrupt()` flows, where a task can complete producing zero channel writes and LangGraph calls `putWrites(config, [], taskId)`. `putWrites` now no-ops on empty writes, matching the behavior of the postgres and sqlite savers (which iterate and naturally skip empty batches).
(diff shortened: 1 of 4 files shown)


christian-bromann (approved):
LGTM 👍 

Thank you for the contribution!

PR #2569, opened 2026-06-23
fix(sdk): coalesce submit threadId when useStream hook is unbound

stescott:
## Summary

Aligns the LGP `useStream` / orchestrator `submit()` paths with the **custom transport** behavior: when the hook is mounted with `threadId: null`, a caller-provided `submitOptions.threadId` should be usable for `runs.stream()` without always going through `threads.create()`.

## Context / docs nuance

`SubmitOptions.threadId` is documented as:

> *"The ID to use when creating a new thread… when threadId is null or undefined."*

So today’s LGP behavior (pass it into `threads.create()`, then stream on the response `thread_id`) is **consistent with that wording**. The problems are:

1. **Inconsistency** — custom transport already does `this.#threadId ?? submitOptions?.threadId` and streams without a forced create; LGP React + orchestrator do not.
2. **Pre-provisioned threads** — when the host has already created the thread (upsert/`ifExists: do_nothing`), `submit()` should not need another `threads.create()` just because the hook prop is still `null` (bootstrap gate / render lag).
3. **Fragile create path** — the run uses `thread.thread_id` from the create response rather than the caller-supplied ID; any platform mismatch can split conversation state.

## Root cause (LGP)

In `stream.lgp.tsx` `submit()`:

```ts
let usableThreadId = threadId; // hook prop only

if (!usableThreadId) {
  const thread = await client.threads.create({
    threadId: submitOptions?.threadId,
    ...
  });
  usableThreadId = thread.thread_id;
}
```

When hook `threadId` is `null` and `submitOptions.threadId` is set, LGP **always** calls `threads.create()` instead of coalescing and streaming directly (custom transport pattern).

## Fix

- Coalesce `threadId ?? submitOptions?.threadId` before deciding to mint
- When adopting `submitOptions.threadId` while the hook is still unbound, sync internal refs / `onThreadId` without calling `threads.create()`
- Mirror the same coalesce in `orchestrator.ts` (LGP non-React path)

## Test plan

- [x] New regression test: `uses submitOptions.threadId when hook threadId is null (no threads.create)`
- [x] `pnpm test src/react/stream.lgp.test.tsx` in `libs/sdk`

## Related / consumer note

- @superagent/ui-kit HOZ-3555: consumer-side fix gates `useStream` until bootstrap completes, awaits upsert, and pins `threadId` on submit. That works on current SDK when create honors the hint; this PR makes the SDK path match custom transport and avoids redundant create for pre-provisioned threads.
- Happy to adjust if maintainers prefer documenting “create-only” semantics and adding an explicit opt-out (e.g. `ifExists`) instead.

### DIFF
--- a/.changeset/usestream-submit-threadid-idempotent.md
+++ b/.changeset/usestream-submit-threadid-idempotent.md
@@ -1,0 +1,5 @@
+---
+"@langchain/langgraph-sdk": patch
+---
+
+fix(sdk): honor submitOptions.threadId with idempotent thread create when useStream is unbound
--- a/libs/sdk/src/react/stream.lgp.test.tsx
+++ b/libs/sdk/src/react/stream.lgp.test.tsx
@@ -63,6 +63,28 @@ function mountThread(client: Client) {
   return thread;
 }
 
+function mountUnboundThread(client: Client) {
+  let thread:
+    | ReturnType<typeof useStream<Record<string, unknown>>>
+    | undefined;
+
+  function Probe() {
+    thread = useStream<Record<string, unknown>>({
+      assistantId: "test-assistant",
+      client,
+      threadId: null,
+      fetchStateHistory: false,
+    });
+    return null;
+  }
+
+  renderToString(<Probe />);
+  if (thread == null) {
+    throw new Error("Failed to mount stream handle.");
+  }
+  return thread;
+}
+
 it("does not evaluate guarded accessors during object spread", () => {
   const client = createMockClient();
   const thread = mountThread(client);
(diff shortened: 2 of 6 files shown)


christian-bromann (changes requested):
Nice catch on the redundant-create / orphan-thread issue. My concern is that skipping creation entirely breaks the documented contract: `SubmitOptions.threadId` is documented as the ID used to *create* a new thread (optimistic UI). After this change, an unbound hook streams straight to `/threads/${threadId}/runs/stream`, which 404s if the host hasn't pre-provisioned the thread.

Could we use the idempotent middle ground instead of skipping create?

```ts
let usableThreadId = threadId ?? submitOptions?.threadId;

if (!threadId) { // hook still unbound
  const thread = await client.threads.create({
    threadId: usableThreadId,   // honor caller-supplied hint
    ifExists: "do_nothing",     // idempotent for pre-provisioned threads
    metadata: submitOptions?.metadata,
    signal,
  });
  usableThreadId = usableThreadId ?? thread.thread_id;
  threadIdRef.current = usableThreadId;
  threadIdStreamingRef.current = usableThreadId;
  onThreadId(usableThreadId);
}
```

This hits all three goals: no redundant churn/orphan, streams on the caller ID, and still provisions for callers relying on the documented behavior.

Smaller notes:
- **Metadata drop:** the adopt branch skips create, so `submitOptions.metadata` is never applied at the thread level. The `ifExists` approach avoids this.
- **Tests:** only the React path is covered, not `orchestrator.ts`, and there's no test for the risky case (unbound hook + `threadId` for a thread that doesn't exist server-side).

I also want to point out that we recommend users to use [`@langchain/react`](https://www.npmjs.com/package/@langchain/react) for better streaming DevX and feature support. Eventually we will deprecate and remove the current React primitives in the SDK package.

PR #2593, opened 2026-07-08
fix(core): throw on duplicate edges in Graph and StateGraph

shahrryyar:
### Summary

Duplicate edges were silently accepted in `Graph` and `StateGraph`, diverging from Python LangGraph. This PR makes `addEdge` throw when the same edge is added twice.

This resolves the following `TODO` inside `libs/langgraph-core/src/graph/state.ts`:
`// TODO: this doesn't dedupe edges as in py, so worth fixing at some point`

### Problem

- In Python, edges are kept in sets of tuples (e.g. `set[tuple[str, str]]`). Tuples are hashable and compared by value, so duplicates collapse automatically.
- In JavaScript, edges are stored in a `Set` of arrays (`Set<[string, string]>`). Arrays compare by reference, so calling `.addEdge` twice with the same start/end produced two separate entries. That caused redundant edge evaluation during visualization, sorting, and validation.

### Solution

Per review feedback, the chosen contract is **throw**, not silent dedupe — adding the same edge twice is a bug in user code, and failing loudly surfaces it at graph-construction time rather than hiding it.

1. **`Graph.addEdge`**: checks `this.edges` for an existing `startKey -> endKey` pair and throws ``Edge from `X` to `Y` already exists.``
2. **`StateGraph.addEdge`**: element-wise comparison of waiting edges (fan-in start lists) against `this.waitingEdges`; throws ``Edge from [A, B] to `C` already exists.``
3. **Tests**: `graph_structure.test.ts` asserts both overloads throw on the second add, and that `edges.size` / `waitingEdges.size` are unchanged.

### Behavior change

This is a breaking change for code that calls `addEdge` idempotently. That pattern was already producing a malformed graph, so the throw converts a silent correctness bug into an explicit error. Released as a patch via `.changeset/strong-edges-throw.md` ("Throw when a graph contains a duplicate edge."); happy to retarget the changeset to `minor` or `major` if you prefer.


libs/langgraph-core/src/graph/state.ts
```diff
@@ -1268,6 +1268,17 @@ export class StateGraph<
...
 
+    const alreadyExists = Array.from(this.waitingEdges).some(
+      ([starts, end]) => {
+        if (end !== endKey) return false;
+        if (starts.length !== startKey.length) return false;
+        return starts.every((s, i) => s === startKey[i]);
+      }
+    );
+    if (alreadyExists) {
+      return this;
```
christian-bromann: Shouldn't we throw instead? I think we should let people know that their code is incorrect.

christian-bromann (changes requested):

libs/langgraph-core/src/graph/state.ts
shahrryyar: ohh, done I think all the way is fixed ill update the pr in next minutes. thanks.

christian-bromann (approved):
LGTM 👍

shahrryyar:
Addressed the contract mismatch Copilot flagged: the implementation throws (per @christian-bromann's review), and the title, description, and changeset now all say "throw" instead of "deduplicate". The behavior change is called out explicitly in the description.

CI is green across all checks. Anything else needed before merge?

PR #2602, opened 2026-07-14
fix(sdk): support clearing cron end_time via update({ endTime: null })

HugoDurand:
## What & why

`CronsClient.update` could not clear a cron's `end_time`. The field was typed `endTime?: string`, so a type-checked caller had no way to express "remove the end time" — even though, unlike the Python SDK, the JS request builder was already correct at runtime: it uses per-field `if (payload?.endTime !== undefined)` guards (not a `None`-strip), and `null !== undefined`, so a `null` would already be forwarded and `JSON.stringify` preserves it.

So the defect was purely a **type imprecision**, and the fix is at the source — the model — rather than a workaround in the request logic:

- `CronsUpdatePayload.endTime`: `string` → `string | null`. PATCH semantics are now the standard tri-state:
  - **omit `endTime`** → keep the existing value
  - **`endTime: null`** → clear it
  - **`endTime: "<iso>"`** → set it
- `CronsCreatePayload.endTime` is intentionally left `string` — create never needs to clear.

This mirrors the Python SDK change in langchain-ai/langgraph#8334 (`update(end_time=None)`), and requires an Agent Server that supports clearing `end_time` via `PATCH /runs/crons/{id}` (`end_time: null`). Against an older server, `update(id, { endTime: null })` is a harmless no-op for clearing (the server coalesces `null` back to the stored value), and a clear-only patch may 400.

## Also in this PR (unify `update()`)

While here, `update()`'s payload construction is unified with the existing `create()` / `createForThread()` style: the 16 repetitive `if (payload?.X !== undefined) json.Y = payload.X` guards become a single declarative object literal. `JSON.stringify` drops `undefined` keys and keeps `null`, so the wire output is **identical** — this is a behaviour-preserving refactor, pinned by the pre-existing "only sends defined fields" / "sends all updatable fields" tests. Net −44 lines in the method.

## Test plan

- [x] `vitest run` on `libs/sdk/src/client/crons/index.test.ts` — **9 passed** (8 existing + 1 new: `update(id, { endTime: null })` sends body `{ end_time: null }`), no type errors.
- [x] `pnpm build` (tsc compile + `attw` + `publint`) — clean.
- [x] `oxlint` — 0 warnings / 0 errors; `oxfmt --check` — clean.
- [x] Changeset added (`@langchain/langgraph-sdk`: patch).

### DIFF
--- a/libs/sdk/src/client/crons/index.ts
+++ b/libs/sdk/src/client/crons/index.ts
@@ -101,56 +101,25 @@ export class CronsClient extends BaseClient {
    * ```
    */
   async update(cronId: string, payload?: CronsUpdatePayload): Promise<Cron> {
-    const json: Record<string, unknown> = {};
-
-    if (payload?.schedule !== undefined) {
-      json.schedule = payload.schedule;
-    }
-    if (payload?.timezone !== undefined) {
-      json.timezone = payload.timezone;
-    }
-    if (payload?.endTime !== undefined) {
-      json.end_time = payload.endTime;
-    }
-    if (payload?.input !== undefined) {
-      json.input = payload.input;
-    }
-    if (payload?.metadata !== undefined) {
-      json.metadata = payload.metadata;
-    }
-    if (payload?.config !== undefined) {
-      json.config = payload.config;
-    }
-    if (payload?.context !== undefined) {
-      json.context = payload.context;
-    }
-    if (payload?.webhook !== undefined) {
-      json.webhook = payload.webhook;
-    }
-    if (payload?.interruptBefore !== undefined) {
-      json.interrupt_before = payload.interruptBefore;
-    }
-    if (payload?.interruptAfter !== undefined) {
-      json.interrupt_after = payload.interruptAfter;
-    }
-    if (payload?.onRunCompleted !== undefined) {
-      json.on_run_completed = payload.onRunCompleted;
-    }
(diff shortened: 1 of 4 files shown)


libs/sdk/src/client/crons/index.ts
```diff
@@ -101,56 +101,25 @@ export class CronsClient extends BaseClient {
...
-    if (payload?.streamSubgraphs !== undefined) {
-      json.stream_subgraphs = payload.streamSubgraphs;
-    }
-    if (payload?.streamResumable !== undefined) {
-      json.stream_resumable = payload.streamResumable;
-    }
-    if (payload?.durability !== undefined) {
-      json.durability = payload.durability;
-    }
+    const json: Record<string, unknown> = {
```
jdrogers940, line 104: nit: these values are now all present as undefined in the json - is that fine and intentional?

jdrogers940 (approved):

libs/sdk/src/client/crons/index.ts
HugoDurand: Yes Intentional
`JSON.stringify` drops the `undefined` keys before the request is sent, so the body is identical to the guard-based version.
The "only sends defined fields to API" test asserts exactly that.
Same pattern as `create()`/`createForThread()`. `end_time` is the one that matters: null is kept (clears it), omitted is dropped (leaves it unchanged)

PR #2608, opened 2026-07-15
feat(core): add versioned graph checkpoint migrations

shahrryyar:
## Summary

This PR adds versioned graph checkpoint migrations so persisted graphs can evolve safely when their state schema changes.

It also addresses the reviewer feedback on duplicate migration registration: registering the same migration more than once now throws instead of silently returning. This makes incorrect configuration visible immediately and avoids hiding a bug in application setup.

## What changed

- Added graph versioning and state migration options to graph compilation.
- Migrated checkpoints, pending writes, history, and manual state updates consistently.
- Added strict version validation, mutation isolation, checkpoint identity protection, and fail-closed handling for unsupported cases.
- Added tests, persistence documentation, and a changeset.

## Verification

- `pnpm lint` passed.
- `pnpm format:check` passed.
- `pnpm build` passed: 25/25 tasks.
- Exact CI unit command passed with `CI=1`.
- LangGraph core: 80 test files, 1,648 passed, 4 skipped, no type errors.
- API tests: 210 passed, 1 skipped.
- Checkpoint validation: 3,138 passed, 6 skipped.
- `git diff --check` passed.

Thank you for the review and for pointing out the need to fail fast on incorrect migration configuration.

### DIFF
--- a/.changeset/versioned-graph-migrations.md
+++ b/.changeset/versioned-graph-migrations.md
@@ -1,0 +1,6 @@
+---
+"@langchain/langgraph": minor
+"@langchain/langgraph-checkpoint": patch
+---
+
+feat(core): add versioned graph checkpoint migrations
--- a/docs/docs/concepts/persistence.md
+++ b/docs/docs/concepts/persistence.md
@@ -4,6 +4,45 @@ LangGraph has a built-in persistence layer, implemented through checkpointers. W
 
 ![Checkpoints](img/persistence/checkpoints.jpg)
 
+## Versioned graph deployments
+
+Long-lived threads can outlive the graph code that created them. Set
+`graphVersion` when compiling a graph and register migrations for checkpoint
+formats that are no longer compatible:
+
+```typescript
+const graph = workflow.compile({
+  checkpointer,
+  graphVersion: 2,
+  legacyGraphVersion: 1,
+  stateMigrations: [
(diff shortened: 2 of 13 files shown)


shahrryyar:
Rebased onto current `main` (clean, no conflicts) and dropped an unrelated change that had crept in: the commit was also switching `lint:eslint` in `libs/langgraph-core/package.json` from `eslint` to `oxlint src/`. That has nothing to do with checkpoint migrations and would have quietly changed what the package lints in CI, so it is reverted — the diff is now 12 files, all migration-related.

Verified locally after the rebase:

- `state_migrations.test.ts`: 12 passed
- full `langgraph-core` suite: 74 files, 1510 passed, 3 skipped

This one has not had a review pass yet. It is a larger change than #2593, so if the shape is wrong I would rather adjust early — in particular:

- `graph_version` is added to `CheckpointMetadata` as ordinary metadata so existing checkpointer implementations keep working without changes. Happy to move it somewhere more explicit if you would prefer it not to live in metadata.
- Migrations are registered on the graph and run in the Pregel loop when a resumed checkpoint's version is behind the current one. If you would rather this be an opt-in wrapper than something on the graph builder, say so and I will restructure it.

No rush on my end — mainly want to know whether the direction is something you want before investing more in it.

PR #2610, opened 2026-07-17
fix(langgraph): support falsy command resume values

xiaotianna:
## Summary

Allow falsy values such as `false`, `0`, and an empty string to be used as
`Command.resume` values.

`PregelLoop` determines whether a resume value is present using
`resume != null`, but `mapCommand` previously used a truthiness check. As a
result, falsy resume values produced no pending writes and caused:

```text
EmptyInputError: Received empty Command input
```

## Changes

- Treat only `null` and `undefined` as missing resume values
- Preserve `false`, `0`, and `""` as valid resume values
- Add regression tests for these falsy values

## Tests

```bash
pnpm exec vitest run src/pregel/io.mapCommand.test.ts --project unit --typecheck.enabled=false
```

Result: 14 tests passed.


xiaotianna:
This failure seems unrelated to this PR's code changes.

The failing workflow is `.github/workflows/dependabot-auto-merge.yml`, which uses:

`dependabot/fetch-metadata@25dd0e34f4fe68f24cc83900b1fe3fe149efef98`

The action is pinned to a full commit SHA, but `dependabot/fetch-metadata@*` is not included in the repository/enterprise allowed action patterns.

This PR only changes `libs/langgraph-core/src/pregel/io.ts`, its test file, and a changeset, so I don't think this CI failure is caused by my change.

Could a maintainer either add `dependabot/fetch-metadata@*` to the Actions allowlist, or update/disable the Dependabot auto-merge workflow?

PR #2647, opened 2026-08-03
fix(sdk): declare ESM scope for bundled dependency chunks

ostapondo:
Importing `@langchain/langgraph-sdk` crashes on Node 22.0–22.6 with `SyntaxError: Cannot use import statement outside a module` raised from the bundled `p-retry` chunk. Reported by users in langchain-ai/openwiki#442 against Windows, but it is not Windows-specific — reproduced on macOS with Node 22.2.0.

**Root cause.** `noExternal` dependencies are emitted as ESM `.js` chunks under a virtual `dist/node_modules` tree with no `package.json` of their own. Node stops its package scope lookup at a `node_modules` path segment, so the chunks never inherit the SDK's `"type": "module"` and are parsed as CommonJS on Node versions without module syntax detection (default since 22.7). On newer Node the same layout still fails under `--no-experimental-detect-module` (as `ERR_REQUIRE_CYCLE_MODULE`).

**Change.** After the build, `@langchain/build` writes a `{"type":"module"}` `package.json` into each bundled package directory under `dist/node_modules`. It is a no-op for packages that bundle nothing or are not `type: module` (their ESM chunks already get self-declaring `.mjs` extensions). I chose this over `fixedExtension: true`, which would rename every published `.js` artifact to `.mjs` across all workspace packages and invalidate the hand-written exports maps.

**Verification.**
- Packed the built SDK, installed it fresh, and ran `import`/`require` of the root and `/client` entrypoints on Node 22.2.0 and 24 — all six combinations pass; without the stubs the same commands crash.
- New test in `libs/sdk`: a static check that every directory holding a `.js` chunk sits in a scope declaring ESM (mirroring Node's lookup order, including the node_modules-boundary-first rule), plus a subprocess import of `dist/index.js` under `--no-experimental-detect-module`. Both fail on the unmodified build; both were mutation-tested (wrong stub type, step disabled, stubs deleted — red each time).
- Full workspace build 25/25, sdk suite 750 tests green, lint and format clean. publint/attw run before the stub step and are unaffected; `npm pack` includes the stubs in the tarball.

**Not covered.** Watch mode: tsdown's `build()` resolves before the watcher writes output, so stubs are not written during `--watch` sessions. Dev-only — publish and CI artifacts come from one-shot builds. The emit gap arguably belongs upstream in tsdown (`noExternal` chunks in a virtual `node_modules` carry no scope declaration); this fixes it where this repo builds.

### DIFF
--- a/.changeset/fix-bundled-chunk-esm-scope.md
+++ b/.changeset/fix-bundled-chunk-esm-scope.md
@@ -1,0 +1,7 @@
+---
+"@langchain/langgraph-sdk": patch
+---
+
+fix(sdk): declare ESM scope for bundled dependency chunks
+
+Emit a `{"type":"module"}` package.json into each bundled package directory under `dist/node_modules`, so the ESM chunks are no longer parsed as CommonJS on Node 22.0–22.6, which crashed consumers at import time with "Cannot use import statement outside a module".
--- a/internal/build/index.ts
+++ b/internal/build/index.ts
@@ -1,5 +1,6 @@
-import { resolve, extname } from "node:path";
+import { resolve, extname, basename, dirname, join } from "node:path";
 import { existsSync } from "node:fs";
+import { readdir, writeFile } from "node:fs/promises";
 import { fileURLToPath, pathToFileURL } from "node:url";
 
 import { build, type Format, type UserConfig } from "tsdown";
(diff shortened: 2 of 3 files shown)


ostapondo:
The `ubuntu-latest, 26.x` failure is unrelated to this diff: the same test fails 3/3 on unmodified `main` under Node 26.5.1 and passes under 26.5.0, and every CI run in this repo since Aug 1 is red the same way. Analysis and fix: #2648, #2649. This PR doesn't touch `langgraph-api`.

PR #2649, opened 2026-08-03
fix(api): widen the scheduled-run window in the in-flight cancel test

### LINKED ISSUE #2648: CI red on all PRs since Aug 1: Node 26.5.1 (undici 8.8+) dispatch stall breaks "update state while run in flight"
Every CI run since Aug 1 fails on `Unit Tests (ubuntu-latest, 26.x)` with the same single failure:

```
FAIL tests/api.test.mts > threads crud > update state while run in flight
AssertionError: expected 'error' to be 'interrupted'
```

This hits every branch, including dependabot PRs that only bump frontend deps. The last green run on `main` (Jul 29) used Node 26.5.0; runners now resolve `26.x` to 26.5.1. Nothing in the repo changed.

## Reproduction

On unmodified `main`, `CI=1 npx vitest run tests/api.test.mts` in `libs/langgraph-api`:

| Node | undici (bundled) | Result |
|---|---|---|
| 26.5.0 | 8.7.0 | pass (2/2) |
| 26.5.1 | 8.9.0 | fail (3/3, same assertion) |

The failure needs the full file's concurrent load; the test passes in isolation via `-t`.

## Root cause

The regression is in undici, which jumped 8.7.0 → 8.9.0 inside the Node patch release. Cross-checking by overriding the SDK's fetch with standalone undici builds:

| Node | fetch from | Result |
|---|---|---|
| 26.5.0 | undici@8.9.0 | fail |
| 26.5.0 | undici@8.8.0 | fail (2/2) |
| 26.5.1 | undici@8.7.0 | pass |

So the change landed in undici 8.8.0. A `diagnostics_channel` trace of the failing `POST /threads/:id/state` request shows where the time goes:

```
undici:request:create      +0ms
undici:client:sendHeaders  +1560ms   <- request sat in the dispatcher
undici:request:headers     +1563ms
```

The server handled the request in 1ms (hono log). Under the suite's load (67 concurrent tests sharing one client, several long-lived SSE joins) requests intermittently stall ~1.5–2.6s client-side before their headers are written, with delays quantizing around 500ms. A candidate is the idle-socket validation change in 8.8.0 (nodejs/undici#5499), but I haven't bisected undici commits and don't yet have a standalone repro to report upstream.

## Why this test in particular

It schedules a run with `afterSeconds: 2` and assumes create + 500ms sleep + update-state + cancel complete inside that window. The dispatch stall pushes the cancel past it: the queue picks the run up at the 2s mark, the cancel finds a running run instead of a pending one (`Attempted to cancel non-pending run` in the log), and the abort surfaces in the queue worker as an error, so the run ends `"error"`.

## Fixes

- Widening the test's window to 5s makes the sequence fit even with slow dispatch, and costs nothing since the happy path cancels the run before it starts. PR: #2649. Verified 3/3 green on 26.5.1 and green on 26.5.0.
- Alternatively the workflow could pin `node-version: 26.5.0`, but that freezes the whole matrix over one test and needs a manual unpin later.

One thing worth a maintainer's opinion: cancelling a run that has already started always ends it as `"error"` — `ops.runs.cancel` only sets `"interrupted"` while the run is still pending, and the queue worker classifies the cancel abort like any other failure. If `"interrupted"` is the intended status for a cancelled running run, I'm happy to follow up with that change separately.

ostapondo:
Fixes #2648.

`update state while run in flight` schedules a run 2 seconds out and assumes the create, update-state, and cancel round trips finish inside that window. Since the GitHub runners moved to Node 26.5.1 (undici 8.9.0), client-side fetch dispatch under the suite's concurrent load can stall ~1.5s, the cancel misses the window, and the run ends `"error"` instead of `"interrupted"` — every CI run since Aug 1 is red on 26.x. Full analysis with the undici A/B and `diagnostics_channel` trace is in the issue.

Scheduling the run 5 seconds out fits the sequence even with slow dispatch. The happy path cancels the run before it starts, so the test's duration is unchanged.

Verified locally on unmodified logic: 3/3 failing runs on Node 26.5.1 before the change, 3/3 green after, still green on 26.5.0.

Not covered: the underlying undici 8.8.0 dispatch regression (needs a standalone repro before an upstream report) and the question of whether cancelling an already-running run should end it as `"interrupted"` — both noted in the issue.

### DIFF
--- a/libs/langgraph-api/tests/api.test.mts
+++ b/libs/langgraph-api/tests/api.test.mts
@@ -548,7 +548,9 @@ describe("threads crud", () => {
     const thread = await client.threads.create();
     const run = await client.runs.create(thread.thread_id, "agent_simple", {
       input: { messages: [{ role: "human", content: "foo" }] },
-      afterSeconds: 2,
+      // wide enough that update + cancel land before the run starts, even
+      // when client-side dispatch is slow under the concurrent test load
+      afterSeconds: 5,
     });
 
     // make sure that the run has been created


christian-bromann (approved):
LGTM 👍

PR #2658, opened 2026-08-04
feat(sdk): export HTTPError with isInstance guard

### LINKED ISSUE #2633: No exported error type / type guard to identify LangGraph SDK HTTP errors
### Checked other resources

### Example Code

To reproduce:

``` typescript
// No public way to check: `err instanceof HTTPError` or `isLangGraphSDKError(err)`
```

_No response_

### Description

There's no way to reliably distinguish an error thrown by the SDK's HTTP layer from any other error, comparable to Axios's isAxiosError.

Looking at the source: an HTTPError class exists internally in [libs/sdk/src/utils/async_caller.ts](https://github.com/gopigment/monorepo/blob/master/libs/sdk/src/utils/async_caller.ts), but it is not re-exported from the package's public entry point ([libs/sdk/src/index.ts](https://github.com/gopigment/monorepo/blob/master/libs/sdk/src/index.ts)). The only publicly exported error-related symbols are ProtocolError, StreamError, and MediaAssemblyError, which relate to streaming, not to request/response failures.

Expected behavior: Export HTTPError (or an equivalent) from the package root, and/or provide a type guard like isLangGraphHTTPError(err) so consumers can safely narrow caught errors and access status/response with proper typing.

### System Info

Environment: @langchain/langgraph-sdk, current main as of 2026-07-24

Roovelrz:
## Summary

Expose the SDK HTTP error through the public package API so consumers can reliably identify and narrow HTTP request failures.

## Changes

* export the existing `HTTPError` class
* add a branded `HTTPError.isInstance()` runtime type guard
* re-export `HTTPError` from the SDK package root
* update the existing HTTP error test to assert the public error class
* add focused coverage for type narrowing, error metadata, retained response access, and unrelated values
* add a patch changeset for `@langchain/langgraph-sdk`

## Motivation

The SDK already throws a structured `HTTPError` containing `status`, `text`, and an optional `response`, but the class is currently private.

Consumers therefore cannot distinguish SDK HTTP failures from unrelated errors without parsing error messages or using unsafe casts.

This change makes the existing error type public without changing request execution, response retention, or retry behavior.

## Test plan

* [x] Run the focused `async_caller.test.ts` suite
* [x] Run the full `@langchain/langgraph-sdk` unit test suite
* [x] Build `@langchain/langgraph-sdk`
* [x] Run oxlint on the changed TypeScript files
* [x] Run oxfmt checks on the changed non-test source files
* [x] Verify ordinary errors, structurally similar objects, and null values are rejected by `HTTPError.isInstance()`

Fixes #2633

### DIFF
--- a/libs/sdk/src/utils/async_caller.ts
+++ b/libs/sdk/src/utils/async_caller.ts
@@ -67,10 +67,13 @@ function isResponse(x: unknown): x is Response {
   return "status" in x && "statusText" in x && "text" in x;
 }
 
+const HTTP_ERROR_SYMBOL = Symbol.for("langgraph.sdk.http_error");
+
 /**
  * Utility error to properly handle failed requests
  */
-class HTTPError extends Error {
+export class HTTPError extends Error {
+  protected readonly [HTTP_ERROR_SYMBOL] = true as const;
   status: number;
 
   text: string;
@@ -82,6 +85,15 @@ class HTTPError extends Error {
     this.status = status;
     this.text = message;
     this.response = response;
+  }
+
+    static isInstance(error: unknown): error is HTTPError {
+    return (
+      typeof error === "object" &&
+      error !== null &&
+      HTTP_ERROR_SYMBOL in error &&
+      (error as Record<symbol, unknown>)[HTTP_ERROR_SYMBOL] === true
+    );
   }
 
   static async fromResponse(
@@ -117,6 +129,8 @@ class HTTPError extends Error {
  * means that by default, each call will be retried up to 5 times, with an
  * exponential backoff between each attempt.
  */
+
+
 export class AsyncCaller {
   protected maxConcurrency: AsyncCallerParams["maxConcurrency"];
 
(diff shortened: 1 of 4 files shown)


libs/sdk/src/utils/async_caller.ts
```diff
@@ -117,6 +117,15 @@ class HTTPError extends Error {
  * means that by default, each call will be retried up to 5 times, with an
  * exponential backoff between each attempt.
  */
+
+/**
+ * Returns whether an unknown value is an HTTP error thrown by the SDK.
+ */
+export const isHTTPError = (error: unknown): error is HTTPError => {
```
christian-bromann: To match other primitives, e.g. `BaseMessage` from `langchain`, we should attach a static `isInstance` to the class that is branded.

christian-bromann (changes requested):

libs/sdk/src/utils/async_caller.ts
Roovelrz: Thanks, updated this to follow the branded primitive pattern. `HTTPError` now carries a `Symbol.for` brand and exposes `HTTPError.isInstance()` for runtime type narrowing, instead of relying on a standalone `instanceof`-based guard.

PR #2660, opened 2026-08-05
feat: Add Oracle AI Vector Database Integration

fileames:
Adds Oracle AI  Database persistence support for LangGraph.js through the new `@langchain/langgraph-checkpoint-oracledb` package.

  - Adds `OracleCheckpointSaver` for durable graph checkpoints, pending writes, checkpoint listing, child namespaces, and
  thread deletion.
  - Adds `OracleStore` for long-term memory, filtering, batched operations, and namespace management.
  - Supports optional Oracle VECTOR-powered semantic search with explicitly managed vector indexes.
  - Supports direct connections, existing pools, configurable table prefixes, automatic table setup, and cleanup APIs.

The code for this PR was primarily developed by @Devx228 during his internship at Oracle.


libs/checkpoint-oracledb/README.md
```diff
@@ -0,0 +1,255 @@
...
+const results = await store.search(["memories", "user-1"], {
+  query: "database memory",
+  limit: 5,
+});
+```
+
+VECTOR indexes are managed explicitly:
+
+```ts
+await store.createVectorIndex({
```
fede-kamel: These vector-index management APIs don't exist on `OracleStore` — the vector index is created automatically during `setup()` via `VECTOR_MIGRATIONS` (`src/store/migrations.ts`). Verified against the built package:

```js
typeof store.createVectorIndex  // "undefined"
typeof store.listVectorIndexes  // "undefined"
typeof store.dropVectorIndex    // "undefined"
```

Anyone following this section gets a runtime `TypeError`. Suggest replacing it with a description of the automatic index creation plus the real knobs (`index_type`, `index_name`, `accuracy`, `parallel`).

libs/checkpoint-oracledb/README.md
```diff
@@ -0,0 +1,255 @@
...
+const indexes = await store.listVectorIndexes();
+await store.dropVectorIndex({
+  name: "LG_MEMORY_IVF_IDX",
+  ifExists: true,
+});
+```
+
+## Tables and cleanup
+
+The checkpoint saver accepts `tablePrefix`, which is normalized to uppercase:
```
fede-kamel: The option is `tableSuffix`, not `tablePrefix`, and the names are suffixed, not prefixed: `getOracleCheckpointTables("FOO")` resolves to `CHECKPOINTS_FOO`, `CHECKPOINT_BLOBS_FOO`, `CHECKPOINT_WRITES_FOO`, `CHECKPOINT_MIGRATIONS_FOO`.

Worth noting the failure mode: `new OracleCheckpointSaver({ connection, tablePrefix: "MYPREFIX" })` is accepted and silently ignored (unknown options aren't validated), so a reader following this section ends up on the default unsuffixed tables with no error.

libs/checkpoint-oracledb/README.md
```diff
@@ -0,0 +1,255 @@
...
+STORE_MIGRATIONS_<SUFFIX>
+VECTOR_MIGRATIONS_<SUFFIX>
+```
+
+Without an index configuration the default suffix is `novec`. With an index
+configuration it is derived deterministically from dimensions and fields, as
+in Python. Specify the same suffix explicitly when applications in both
+languages must share one Store. Empty keys are rejected because Oracle treats
+an empty string as `NULL` and the shared key column is `NOT NULL`.
+JavaScript currently supports `COSINE` vector distance; setup rejects a shared
```
fede-kamel: Stale limitation: the package supports all three metrics. `ORACLE_VECTOR_DISTANCE_METRICS` is `["COSINE", "EUCLIDEAN", "DOT"]`, constructor validation accepts `EUCLIDEAN`/`DOT` (verified), and both the generated `CREATE VECTOR INDEX ... DISTANCE <metric>` and the `VECTOR_DISTANCE(..., <metric>)` search path honor them. Nothing rejects a shared Python store configured with a non-COSINE metric — `assertStoredIndexConfigMatches` only requires both sides to agree.

.changeset/five-glasses-post.md
```diff
@@ -0,0 +1,5 @@
+---
+"@langchain/langgraph-checkpoint-oracledb": major
```
fede-kamel: `major` on a package whose `package.json` is already at `1.0.0` means the first npm publish will be **2.0.0** (`changeset status` confirms the pending major bump). Other new packages in this repo started lower (sdk-angular at 0.1.0 via a minor changeset, checkpoint-redis at 0.0.x). Consider setting `version` to `0.0.0` and keeping this major changeset (first publish becomes 1.0.0), or downgrading the bump.

libs/checkpoint-oracledb/src/sql.ts
```diff
@@ -0,0 +1,589 @@
...
+      const predicates = [`${subject}.type() == "array"`];
+      value.forEach((item, index) => {
+        const itemPredicate = compileMetadataFilterValue(
+          compiler,
+          "@",
+          item,
+          `${path}[${index}]`,
+          ancestors
+        );
+        predicates.push(`exists(${subject}[*]?(${itemPredicate}))`);
```
fede-kamel, line 449: Array filter semantics diverge from the Python package this integration shares tables with. This compiles arrays to order-insensitive per-element containment, while Python's `_build_metadata_predicates` uses exact positional `JSON_EQUAL` (and its comments say so). Reproduced against the same checkpoint row (metadata `{"tags": ["a", "b"]}`) on 23ai:

| `list(filter=...)` | Python | JS |
|---|---|---|
| `{tags: ["a", "b"]}` | 1 | 1 |
| `{tags: ["b", "a"]}` | 0 | 1 |
| `{tags: ["a"]}` | 0 | 1 |

JS matches checkpoint-postgres's `@>` containment semantics, so Python is arguably the outlier — but each implementation's comments currently claim to match the other. Suggest documenting the divergence in the README's cross-language section (and possibly filing an alignment issue on the Python side).

libs/checkpoint-oracledb/src/store/index.ts
```diff
@@ -0,0 +1,2137 @@
...
+    return this.withConnection(async (connection) => {
+      const tables = {
+        store: this.tableName,
+        storeVectors: this.vectorTableName,
+        storeMigrations: this.migrationTableName,
+        vectorMigrations: this.vectorMigrationTableName,
+      };
+      const vectorRequired = this.indexConfig !== undefined;
+      const expectedTables = getExpectedStoreTables(tables, vectorRequired);
+      const expectedVersions = [0, 1, 2, 3, 4];
```
fede-kamel: Two related issues here:

1. The versions are hardcoded rather than derived from `STORE_MIGRATIONS.length` / `VECTOR_MIGRATIONS.length`, so they silently drift when a migration is appended.
2. Only the store migration table is inspected — vector migration state is never read, so a vector-configured store whose index migration (v1) was never applied still reports `ready`.

Repro on a live 23ai database: after `DROP INDEX <vector index>` + `DELETE FROM VECTOR_MIGRATIONS_<sfx> WHERE v = 1`, `getDiagnostics()` returns `{ status: "ready", issues: [] }` even though `vector.observedIndexes` confirms the index is gone.

libs/checkpoint-oracledb/vitest.config.js
```diff
@@ -0,0 +1,36 @@
...
+
+export default defineConfig((env) => {
+  /** @type {import("vitest/config").UserConfigExport} */
+  const common = {
+    test: {
+      hideSkippedTests: true,
+      globals: true,
+      testTimeout: 30_000,
+      exclude: ["**/*.int.test.ts", ...configDefaults.exclude],
+      passWithNoTests: true,
```
fede-kamel, line 46: nit (CI trap): `passWithNoTests: true` is inherited by int mode, and all four int files `describe.skip` when credentials are absent — so `pnpm test:int` with no env vars exits 0 having executed nothing (verified: `101 skipped`, exit code 0). If `ORACLE_*` creds ever disappear from a CI environment, integration coverage vanishes silently. Consider failing loudly (or at least warning) in int mode when `ORACLE_CONNECT_STRING` is unset.

fede-kamel (commented):
Reviewed the full package and ran it end-to-end against an Oracle Database 23ai (Free, container) instance. LGTM — the core is solid (I don't have approve rights on this repo, so leaving this as a comment review). What I verified:

- Build clean (attw + publint), unit suite 132/132.
- Full integration suite: **817 passed / 1 intentional skip** against a real database. One runbook note: on a small container (default `processes=200`) the int files must run serialized (`vitest run --mode int --no-file-parallelism`); run in parallel they exhaust DB processes and cascade-fail with `NJS-511`/`ORA-12516`.
- **Cross-language compatibility verified live** against the Python `langgraph-oracledb` package, both directions: checkpoints (msgpack `bytes` channel values, >1 MiB values routed to `CHECKPOINT_BLOBS`, nested metadata filters, pending writes incl. `__error__`), and the store — an identical index config derived the same table suffix from both languages, and each side vector-searched the other's embeddings with identical scores.
- A real `StateGraph` compiled with `OracleCheckpointSaver` resumes state across invocations, with full checkpoint history.
- Injection review: every interpolated identifier is validated, values are bound, JSON-path literals are escaped — I could not construct an injection through filter keys or index config values.

Inline comments below each include a reproduction. Two repo-level items without a good diff anchor:

1. **No `LICENSE` file** — the package ships only `LICENSE-3rd-party.txt`; every sibling lib includes the MIT text and `package.json` declares MIT.
2. **Stale lockfile importer entry** — `pnpm-lock.yaml` records `@langchain/core` (specifier `^1.2.1`) under this package's `dependencies`, but `package.json` declares it only as a peerDependency (`^1.1.44`). Looks generated from an earlier package.json revision; worth regenerating the lockfile. (It does not break `pnpm install --frozen-lockfile` on pnpm 9.15.9 — verified.)

Nice work overall — the byte-compatible `pythonJsonDumps` suffix hashing and the ORA-00001 retry paths are particularly carefully done.

libs/checkpoint-oracledb/src/saver.ts
```diff
@@ -0,0 +1,1401 @@
...
+function shouldUseBlob(value: unknown, jsonSizeThresholdMb: number): boolean {
+  if (
+    value === null ||
+    typeof value === "string" ||
+    typeof value === "boolean" ||
+    (typeof value === "number" && Number.isFinite(value))
+  ) {
+    return false;
+  }
+  if (!isPlainJsonValue(value)) return true;
```
fede-kamel, line 464: Worth a README note (shared with the Python saver, so not a JS regression): plain-JSON channel values under the size threshold bypass the serde entirely and are stored as readable JSON inside the `checkpoint` column. With a custom **encrypting** serde, `dumpsTyped` is never invoked for them.

Repro: `put()` with an XOR "encrypting" serde and a small `{ api_key: "..." }` channel value → serde call count 0, and the secret is readable via `JSON_SERIALIZE(checkpoint)`. A sentence in the serializer section of the README would prevent a bad surprise; routing everything through the serde when a custom one is configured could be a follow-up enhancement.

libs/checkpoint-oracledb/src/tests/store-index-config.test.ts
```diff
@@ -0,0 +1,731 @@
...
+  test("sorts keys by code point rather than locale", () => {
+    // localeCompare would place "a" before "A" in most locales; Python does not.
+    expect(pythonJsonDumps({ a: 1, A: 2, _z: 3 })).toBe(
+      '{"A": 2, "_z": 3, "a": 1}'
+    );
+  });
+
+  test("escapes non-ASCII the way ensure_ascii=True does", () => {
+    // CPython: json.dumps({"fields": ["tést"]}, sort_keys=True)
+    expect(pythonJsonDumps({ fields: ["tést"] })).toBe(
```
fede-kamel: nit: this string literal contains an invisible U+007F (DEL) between `é` and `st` — visible only via `od -c` (`t   é  ** 177   s   t`). It's deliberate (it exercises the `ensure_ascii` DEL escaping), but it's invisible in editors and easily destroyed by formatters or copy-paste. Suggest `"t\u00e9\u007fst"` with a brief comment instead.

libs/checkpoint-oracledb/src/diagnostics.ts
```diff
@@ -0,0 +1,868 @@
...
+  return "ready";
+};
+
+export const probeOracleVector = async (
+  connection: OracleDiagnosticsConnection,
+  dims: number
+): Promise<OracleStoreVectorDiagnostics["probe"]> => {
+  const safeDims =
+    typeof dims === "number" && Number.isFinite(dims) && dims > 0
+      ? Math.min(Math.floor(dims), ORACLE_VECTOR_MAX_DIMENSIONS)
```
fede-kamel: nit: dims are clamped to `ORACLE_VECTOR_MAX_DIMENSIONS` (65535), but the string literal built from the probe vector isn't checked against the 32767-byte VARCHAR2 bind limit that `vectorLiteral` in the store enforces. Above roughly 16k dims the probe bind exceeds that limit, and a perfectly healthy vector database gets reported as probe `unknown`.

libs/checkpoint-oracledb/src/sql.ts
fede-kamel: Follow-up: the Python side is being aligned to this (containment) semantics rather than documenting the divergence — filed as oracle/langchain-oracle#296 with the repro table above, fix in oracle/langchain-oracle#293. With that change both packages return identical rows for list filters on shared tables (verified live with a seven-case typed-element matrix: dict containment/miss, `null`, `true`/`false` discrimination, numbers, nested arrays). So no README caveat needed here once that lands — this branch's containment behavior becomes the cross-language contract, and the Python comment claiming "lists keep exact-match semantics ... matching the JS saver" gets corrected in the same PR.

libs/checkpoint-oracledb/README.md
fileames: Looks like README was stale after the latest changes, did the update.

.changeset/five-glasses-post.md
fileames: Done

fileames:
Thanks @fede-kamel for the detailed review :) I did the changes requested. Hopefully all looks good now.

fileames:
Hi @dqbd! I noticed you’ve reviewed a few similar PRs in the past, so I was hoping you might be able to help us out.

We’re looking to add Oracle as a checkpointer for LangGraph.js. We already have the corresponding [Python package](https://github.com/oracle/langchain-oracle/tree/main/libs/langgraph-oracledb), and we’d really appreciate your guidance or help in getting Oracle support into LangGraph.js as well. Thanks!

fede-kamel:
Quick status note from the Oracle side: this branch now shows as conflicting with `main`, so it will need a rebase before it can be merged. @fileames, happy to help with that if useful.

For any maintainer picking this up: the implementation was reviewed and exercised end to end against a real Oracle 26ai database in August (see the earlier review threads), and the requested changes were applied. It has been waiting on a maintainer look since then.

PR #2665, opened 2026-08-06
fix(langgraph): run all tasks when maxConcurrency is 1

### LINKED ISSUE #2656: Send fan-out tasks are silently dropped when maxConcurrency: 1
### Summary

Invoking a graph that fans out with `Send` while passing `maxConcurrency: 1` silently skips the dispatched tasks. No error is raised, no warning is logged, and the graph returns a normal-looking result — the only symptom is that the work never happened.

Any value >= 2 (and omitting the option entirely) behaves correctly.

### Environment

- `@langchain/langgraph` 1.4.8
- Node v24.18.1
- Windows 11 (also observed inside a larger graph on the same setup)

### Reproduction

```js
import { StateGraph, Annotation, Send, START, END } from '@langchain/langgraph';

const State = Annotation.Root({
  items: Annotation({ reducer: (a, b) => b ?? a, default: () => [] }),
  seen: Annotation({ reducer: (a, b) => (a ?? []).concat(b ?? []), default: () => [] }),
});

const build = () => new StateGraph(State)
  .addNode('start', () => ({}))
  .addNode('worker', state => ({ seen: [state.item] }))
  .addConditionalEdges('start', s => s.items.map(item => new Send('worker', { item })), ['worker'])
  .addEdge(START, 'start')
  .addEdge('worker', END)
  .compile();

const items = ['a', 'b', 'c', 'd', 'e', 'f'];
for (const maxConcurrency of [1, 2, 4, undefined]) {
  const out = await build().invoke({ items }, maxConcurrency ? { maxConcurrency } : {});
  console.log(`maxConcurrency=${maxConcurrency} sent=${items.length} ran=${(out.seen ?? []).length}`);
}
```

### Actual

```
maxConcurrency=1         sent=6  ran=0
maxConcurrency=2         sent=6  ran=6
maxConcurrency=4         sent=6  ran=6
maxConcurrency=undefined sent=6  ran=6
```

### Expected

`maxConcurrency: 1` should run all six tasks sequentially — concurrency is a rate limit, not a task budget. Failing that, it should raise rather than skip work silently.

### Impact

We hit this while bounding a fan-out of paid LLM calls so a per-run cost ceiling could be enforced between batches. `maxConcurrency: 1` looked like the safest possible setting and instead skipped the work entirely — a document with N entities produced results for none of them, with the run reporting success. Silent partial execution is difficult to detect downstream because the output is structurally valid.

In a larger graph we observed one task running rather than zero; either way the remainder are dropped without a signal.

### Workaround

Floor the value at 2 and pin the behaviour with a regression test, so a fix upstream surfaces as a test failure rather than going unnoticed.

pushkarverma3698:
Fixes #2656

Invoking a graph that fans out with `Send` under `maxConcurrency: 1` silently skips every task after the first. No error is raised and the output is structurally valid, so the lost work leaves no signal downstream.

## Root cause

The scheduling loop in `PregelRunner._executeTasksWithRetry` continues while:

```ts
(startedTasksCount === 0 || Object.keys(executingTasksMap).length > 0) && tasks.length
```

With a limit of 1, `executingTasksMap` holds a single task. When it settles and is deleted at the bottom of the loop, the map is empty and `startedTasksCount` is non-zero, so the loop exits with the remaining tasks never scheduled. Any limit >= 2 keeps at least one task in flight across a settle, which is why only `1` is affected.

The condition conflates "nothing in flight" with "nothing left to start".

## Fix

Continue while unstarted tasks remain **or** tasks are in flight. The inner top-up loop already caps in-flight work at `maxConcurrency`, so the limit itself is unchanged. Tasks injected mid-tick (error handlers, PUSH tasks via `call`) still keep the loop alive through the barrier exactly as before.

Reproduction from the issue:

| maxConcurrency | before | after |
|---|---|---|
| 1 | ran=0 | ran=6 |
| 2 | ran=6 | ran=6 |
| 4 | ran=6 | ran=6 |
| unset | ran=6 | ran=6 |

## Test

Regression test added beside the existing `test_max_concurrency` port: a 6-way `Send` fan-out at `maxConcurrency: 1` must complete all six tasks **and** observe a peak concurrency of exactly 1 — so a future change can't "fix" this by quietly widening the limit. Fails on `main` (0 of 6 tasks run), passes with this change.

`libs/langgraph-core`: 79 files, 1639 tests passed, 0 type errors. `oxlint` and `oxfmt --check` clean.

### DIFF
--- a/.changeset/fix-max-concurrency-one-drops-tasks.md
+++ b/.changeset/fix-max-concurrency-one-drops-tasks.md
@@ -1,0 +1,7 @@
+---
+"@langchain/langgraph": patch
+---
+
+fix(langgraph): run all tasks when maxConcurrency is 1
+
+The runner's scheduling loop exited as soon as its only in-flight task settled, silently dropping every remaining task in the superstep. Keep scheduling while unstarted tasks remain, so `maxConcurrency: 1` executes the full fan-out sequentially.
--- a/libs/langgraph-core/src/pregel/runner.ts
+++ b/libs/langgraph-core/src/pregel/runner.ts
@@ -320,7 +320,8 @@ export class PregelRunner {
       : undefined;
 
     while (
-      (startedTasksCount === 0 || Object.keys(executingTasksMap).length > 0) &&
+      (startedTasksCount < tasks.length ||
+        Object.keys(executingTasksMap).length > 0) &&
       tasks.length
     ) {
       for (
--- a/libs/langgraph-core/src/tests/python_port/graph_structure.test.ts
+++ b/libs/langgraph-core/src/tests/python_port/graph_structure.test.ts
@@ -1047,6 +1047,53 @@ describe("Graph Structure Tests (Python port)", () => {
   });
 
   /**
+   * Regression test for #2656: Send fan-out tasks were silently dropped when
+   * maxConcurrency was 1, because the runner's scheduling loop exited as soon
+   * as the single in-flight task settled.
+   */
+  it("should run all Send tasks sequentially when maxConcurrency is 1", async () => {
+    const StateAnnotation = Annotation.Root({
+      items: Annotation<unknown[]>({
+        reducer: (a, b) => a.concat(b),
+        default: () => [],
+      }),
+    });
+
+    let currently = 0;
+    let maxCurrently = 0;
+
+    const worker = async (state: unknown): Promise<{ items: unknown[] }> => {
+      currently += 1;
+      if (currently > maxCurrently) {
+        maxCurrently = currently;
+      }
+      await new Promise<void>((resolve) => {
+        setTimeout(resolve, 1);
+      });
+      currently -= 1;
+      return { items: [state] };
+    };
+
+    const sendToMany = (): Send[] => {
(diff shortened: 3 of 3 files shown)


t3s7r:
Independent confirmation of the bug and of this fix.

We hit #2656 on `@langchain/langgraph` 1.4.9 in a production graph that fans out with `Send` under `maxConcurrency: 1`. Reproduction on the unpatched dist: 8 `Send` tasks, `maxConcurrency: 1`, with a checkpointer. Result: 1 worker ran, `invoke` resolved with no error, the final state carried no results, and `getState().next` still listed 7 `worker` tasks. The same graph under `maxConcurrency: 2` ran all 8 and left `next` empty.

We have been running exactly this loop-guard change (`startedTasksCount < tasks.length || <in flight>`) as a pnpm patch on `dist/pregel/runner.{js,cjs}` since 2026-08-31. The full fan-out now executes sequentially, and we have seen no side effects across daily runs since. We also carry a port of the regression test from this PR, and it is green against the patched build.

#2659 is the same one-line fix and can be closed as a duplicate of this one.

@christian-bromann would you be able to take a look? CI is green here and the change is a single guard condition.

PR #2668, opened 2026-08-06
fix(sdk): always attach Response on HTTPError

### LINKED ISSUE #2632: error.response is only populated when callerOptions.onFailedResponseHook is configured
### Checked other resources

### Example Code

To reproduce:

``` typescript
const client = new Client({ apiUrl: "http://localhost:2024" });
try {
  await client.threads.get("nonexistent-thread");
} catch (err) {
  console.log(err.response); // undefined, even though the request failed with a real HTTP response
}
```
Environment: @langchain/langgraph-sdk, current main as of 2026-07-24

_No response_

### Description

When an API call made through Client fails, the resulting error's .response property is undefined unless the client was configured with callerOptions.onFailedResponseHook.

This is because in [libs/sdk/src/utils/async_caller.ts](https://github.com/gopigment/monorepo/blob/master/libs/sdk/src/utils/async_caller.ts), HTTPError.fromResponse is called with includeResponse: !!onFailedResponseHook — the raw response is only attached as a side effect of registering that hook, which is meant for something else (e.g. logging/telemetry on failure), not for consumers who just want to inspect the response on error.

Expected behavior: There should be an explicit, independent option (e.g. callerOptions.includeResponseOnError or simply always attaching error.response) so callers can access the failed response without having to register an unrelated hook.

### System Info

this is on latest right now

edenbuilds:
## Summary
- Always pass `includeResponse: true` when converting failed `Response` objects to `HTTPError`, so `error.response` is available without registering `onFailedResponseHook`.

## Test plan
- [x] New unit test `should attach error.response without onFailedResponseHook` passes with the fix
- [x] Same test fails with `response: undefined` when the fix is stashed (`git stash` A/B)
- [x] Changeset for `@langchain/langgraph-sdk`

Fixes #2632

### DIFF
--- a/.changeset/always-attach-http-error-response.md
+++ b/.changeset/always-attach-http-error-response.md
@@ -1,0 +1,5 @@
+---
+"@langchain/langgraph-sdk": patch
+---
+
+Always attach the underlying Response on HTTPError.
--- a/libs/sdk/src/utils/async_caller.test.ts
+++ b/libs/sdk/src/utils/async_caller.test.ts
@@ -173,6 +173,42 @@ describe("AsyncCaller", () => {
       );
     });
 
+    it("should attach error.response without onFailedResponseHook", async () => {
+      const caller = new AsyncCaller({ maxRetries: 0 });
+
+      const responseError = {
+        status: 404,
+        statusText: "Not Found",
+        text: () => Promise.resolve("missing"),
+      };
+
+      const failingCallable = vi.fn(() => Promise.reject(responseError));
+
+      await expect(caller.call(failingCallable)).rejects.toMatchObject({
+        status: 404,
+        response: responseError,
+      });
+    });
+
+    it("should not attach a response to failures that carry none", async () => {
+      const caller = new AsyncCaller({ maxRetries: 0 });
+
+      // Only a Response rejection becomes an HTTPError, so these keep no
+      // `response`: a generic failure is rethrown as-is, and a connection
+      // failure is replaced by ConnectionError.
+      for (const failure of [
+        new TypeError("boom"),
+        new TypeError("fetch failed"),
+      ]) {
+        const error = await caller
+          .call(vi.fn(() => Promise.reject(failure)))
+          .catch((caught) => caught);
+
+        expect(error).not.toHaveProperty("response");
+        expect(error).not.toHaveProperty("status");
+      }
+    });
+
     it("should retry connection errors and throw ConnectionError when retries are exhausted", async () => {
       const caller = new AsyncCaller({ maxRetries: 1 });
 
--- a/libs/sdk/src/utils/async_caller.ts
+++ b/libs/sdk/src/utils/async_caller.ts
@@ -160,8 +160,10 @@ export class AsyncCaller {
               if (error instanceof Error) {
                 throw error;
               } else if (isResponse(error)) {
+                // Always attach the failed Response so callers can inspect
+                // error.response without registering onFailedResponseHook.
                 throw await HTTPError.fromResponse(error, {
-                  includeResponse: !!onFailedResponseHook,
+                  includeResponse: true,
                 });
               } else {
                 throw new Error(error);


denyskorolkov:
Matches [the proposed fix in the issue](https://github.com/langchain-ai/langgraphjs/issues/2632#issuecomment-5200516372) (option 2 — always
attach `response`, decoupled from `onFailedResponseHook`).

Nit on the changeset: "even when onFailedResponseHook is unset" reads like the
hook is still relevant post-fix, but it plays no role anymore in whether
`response` gets attached - might read cleaner as just "Always attach the
underlying Response on HTTPError." (the historical coupling is useful context
in the PR description, doesn't need to be in the public changelog).

Also curious: is `response` still expected to be `undefined` in some paths
(e.g. a network failure with no Response object at all), or does this
guarantee a value now?

edenbuilds:
Thanks @denyskorolkov — both taken.

**Changeset.** You're right, the hook has no role post-fix and mentioning it in the changelog only implies otherwise. Now just:

> Always attach the underlying Response on HTTPError.

The historical coupling stays in the PR description.

**On `response` still being `undefined`:** not for `HTTPError`. It's constructed in exactly one place — the `isResponse(error)` branch of `AsyncCaller.call` — so with `includeResponse: true` every `HTTPError` the SDK throws now carries its `Response`.

A failure with no `Response` never becomes an `HTTPError` at all, which is the case you're pointing at:

- a generic rejection hits `if (error instanceof Error) throw error` and is rethrown untouched;
- a network failure whose message matches the connection checks (`fetch failed`, `ECONNREFUSED`, …) is replaced, once retries are exhausted, by a plain `Error` with `name === "ConnectionError"`.

Neither carries `response` or `status`, and neither did before. So the property stays optional on the class because the constructor parameter is optional, but there's no longer an `HTTPError` path that omits it.

I added a test pinning that boundary, since it's exactly the question a future reader will have:

```ts
it("should not attach a response to failures that carry none", async () => {
  // TypeError("boom") → rethrown as-is; TypeError("fetch failed") → ConnectionError
  expect(error).not.toHaveProperty("response");
  expect(error).not.toHaveProperty("status");
});
```

Writing it corrected my own assumption, incidentally — I expected `fetch failed` to surface as the original `TypeError`, and it comes back as `ConnectionError` because the message matches the connection branch even with `maxRetries: 0`.

17 caller tests pass, and the `error.response` regression still fails against the current `async_caller.ts`. Squashed to one commit.

denyskorolkov:
> Thanks Backend & AI Systems Engineer · Founder @ Prema Vision (@denyskorolkov) — both taken.
> 
> **Changeset.** You're right, the hook has no role post-fix and mentioning it in the changelog only implies otherwise. Now just:
> 
> > Always attach the underlying Response on HTTPError.
> 
> The historical coupling stays in the PR description.
> 
> **On `response` still being `undefined`:** not for `HTTPError`. It's constructed in exactly one place — the `isResponse(error)` branch of `AsyncCaller.call` — so with `includeResponse: true` every `HTTPError` the SDK throws now carries its `Response`.
> 
> A failure with no `Response` never becomes an `HTTPError` at all, which is the case you're pointing at:
> 
> * a generic rejection hits `if (error instanceof Error) throw error` and is rethrown untouched;
> * a network failure whose message matches the connection checks (`fetch failed`, `ECONNREFUSED`, …) is replaced, once retries are exhausted, by a plain `Error` with `name === "ConnectionError"`.
> 
> Neither carries `response` or `status`, and neither did before. So the property stays optional on the class because the constructor parameter is optional, but there's no longer an `HTTPError` path that omits it.
> 
> I added a test pinning that boundary, since it's exactly the question a future reader will have:
> 
> ```ts
> it("should not attach a response to failures that carry none", async () => {
>   // TypeError("boom") → rethrown as-is; TypeError("fetch failed") → ConnectionError
>   expect(error).not.toHaveProperty("response");
>   expect(error).not.toHaveProperty("status");
> });
> ```
> 
> Writing it corrected my own assumption, incidentally — I expected `fetch failed` to surface as the original `TypeError`, and it comes back as `ConnectionError` because the message matches the connection branch even with `maxRetries: 0`.
> 
> 17 caller tests pass, and the `error.response` regression still fails against the current `async_caller.ts`. Squashed to one commit.

Thanks for doing this and sharing the ConnectionError catch, glad it's pinned in a test now. Contribution looks great.

edenbuilds:
Thanks @denyskorolkov — appreciate the review and the green light. Happy to iterate if anything else comes up before merge.

christian-bromann (approved):
Thank you , LGTM 👍

PR #2692, opened 2026-08-13
fix(sdk): adopt server metadata on same-id optimistic message echo

christian-bromann:
## Summary

Optimistic humans kept winning over the values echo, so committed additional_kwargs (e.g. attachment paths) stayed off stream.messages until hydration. Prefer the server copy when content matches; streamed tokens still win when they have moved past the snapshot.

- Same-id `values` echoes of an optimistic human now take server-authored metadata (`additional_kwargs`, etc.) on `stream.messages` instead of keeping the plain optimistic copy until reload.
- In-flight AI token streaming is unchanged: streamed content still wins when it has moved past the snapshot.

### DIFF
--- a/.changeset/fix-optimistic-human-values-metadata.md
+++ b/.changeset/fix-optimistic-human-values-metadata.md
@@ -1,0 +1,12 @@
+---
+"@langchain/langgraph-sdk": patch
+---
+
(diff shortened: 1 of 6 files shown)


wario2k:
@christian-bromann any idea on when this could be reviewed?

PR #2706, opened 2026-08-19
fix(langgraph): dedupe merged callback handlers by identity

zduric-langchain:
This fixes a double-streaming bug in certain nested graph case scenarios.

`mergeCallbacks` concatenated `handlers` and `inheritableHandlers` while deduping tags, so a handler inherited by both the parent and nested config (in nested graphs) picked up an extra registration at every nesting level with tracing on. With tracing on, a nested streamMode: "messages" run delivered every token twice or more.


christian-bromann (dismissed):
LGTM 👍

PR #2708, opened 2026-08-19
fix(sdk): reconnect protocol SSE after clean EOF

### LINKED ISSUE #2701: [SDK] Thread-centric SSE stream does not reconnect after clean EOF
## Checked other resources

- [x] I added a descriptive title.
- [x] I searched the LangGraph.js documentation and API reference.
- [x] I searched existing GitHub issues and did not find a matching clean-EOF report.
- [x] The behavior originates in the LangGraph JavaScript SDK rather than application code.
- [x] The bug is still present in the latest stable release, `@langchain/langgraph-sdk@1.9.29`.

## Example code

This standalone reproduction uses the public SDK API and an injected `fetch`. It does not require credentials, a running agent server, or a UI framework.

```bash
mkdir langgraph-clean-eof-repro
cd langgraph-clean-eof-repro
npm init -y
npm install @langchain/langgraph-sdk@1.9.29 @langchain/core
```

Save the following as `repro.mjs`:

```javascript
import assert from "node:assert/strict"
import {
  ProtocolSseTransportAdapter,
} from "@langchain/langgraph-sdk"

const encoder = new TextEncoder()

const completedEvent = {
  type: "event",
  event_id: "event-1",
  seq: 1,
  method: "lifecycle",
  params: {
    namespace: [],
    timestamp: "2026-01-01T00:00:00Z",
    data: { event: "completed" },
  },
}

let streamRequests = 0
const reconnectAttempts = []

const fetchMock = async (_input, init = {}) => {
  streamRequests += 1

  if (streamRequests === 1) {
    return new Response(
      new ReadableStream({
        start(controller) {
          controller.enqueue(
            encoder.encode(
              `data: ${JSON.stringify(completedEvent)}\n\n`,
            ),
          )

          // The successful SSE response reaches clean EOF.
          controller.close()
        },
      }),
      {
        status: 200,
        headers: {
          "content-type": "text/event-stream",
        },
      },
    )
  }

  // Keep the replacement connection open if the SDK reconnects.
  return await new Promise((_resolve, reject) => {
    const signal = init.signal

    if (signal?.aborted) {
      reject(signal.reason)
      return
    }

    signal?.addEventListener(
      "abort",
      () => {
        reject(
          signal.reason ??
            new DOMException("Aborted", "AbortError"),
        )
      },
      { once: true },
    )
  })
}

const adapter = new ProtocolSseTransportAdapter({
  apiUrl: "http://agent-server.test",
  threadId: "thread-1",
  fetch: fetchMock,
  idleReconnect: 0,
  maxReconnectAttempts: 1,
  reconnectDelayMs: () => 0,
  onReconnect: ({ attempt }) => {
    reconnectAttempts.push(attempt)
  },
})

const handle = adapter.openEventStream({
  channels: ["lifecycle"],
})

try {
  await handle.ready
  await new Promise((resolve) => setTimeout(resolve, 100))

  console.log({
    streamRequests,
    reconnectAttempts,
  })

  assert.equal(
    streamRequests,
    2,
    "Active SSE handle did not reconnect after clean EOF",
  )

  assert.deepEqual(reconnectAttempts, [1])
} finally {
  handle.close()
  await adapter.close()
}
```

Run it:

```bash
node repro.mjs
```

Actual output:

```text
{ streamRequests: 1, reconnectAttempts: [] }

AssertionError [ERR_ASSERTION]: Active SSE handle did not reconnect after clean EOF
```

Expected output:

```text
{ streamRequests: 2, reconnectAttempts: [1] }
```

The corresponding user-visible sequence is:

1. Create one `ThreadStream` using the default SSE transport.
2. Start a run and consume it through completion.
3. The active SSE HTTP response reaches clean EOF while the client-side `ThreadStream` remains open.
4. Start a second run on the same `ThreadStream`.
5. The command request succeeds, but the SDK does not issue a new `POST /threads/{thread_id}/stream/events` request.
6. Recreate or rehydrate the client. The second run's persisted result appears, confirming that the run completed but its live events were not observed.

## Error message and stack trace

There is no SDK error or stack trace. The SSE response completes normally, and the SDK silently treats the subscription as permanently finished. The assertion in the reproduction makes that silent failure visible.

### Actual behavior

`ProtocolSseTransportAdapter.openEventStream()` reconnects only when stream processing throws. When the response body reaches clean EOF, the `for await` loop ends normally and the implementation closes the shared queue and returns:

```typescript
for await (const event of iterable) {
  if (ac.signal.aborted || this.closed) break
  if (isRecord(event.data)) streamQueue.push(event.data)
}
streamQueue.close()
return
```

Because no exception is thrown, `maxReconnectAttempts`, `onReconnect`, and `reconnectDelayMs` are bypassed.

The surrounding `ThreadStream` continues retaining the same subscription handle. Before another run, `#prepareForNextRun()` resumes paused subscriptions, but `resume()` only opens a local delivery gate; it cannot reopen an HTTP response or a queue that has already been closed. Consequently, the next command can succeed while the client has no live event transport.

### Expected behavior

If an SSE response reaches EOF while its handle is still active—that is, neither `handle.close()` nor transport shutdown initiated the end—the SDK should treat the EOF as an unexpected disconnect and reopen the event subscription through the existing reconnect path.

The retry should honor `maxReconnectAttempts`, `onReconnect`, and `reconnectDelayMs`. It should not occur after an explicit client abort or transport close. Replay and deduplication can continue following the SDK's existing transport contract; this report does not require a particular `since` value.

An alternative would be to mark the retained subscription handle as terminal and recreate it before the next command. Immediate reconnect is preferable because an active thread-level subscription may need to observe events not initiated by that client.

### Impact

- Later commands succeed but their streamed output is absent.
- Applications can remain stale indefinitely without an error signal.
- Reloading or rebuilding the client appears to fix the problem because persisted state is hydrated through a fresh connection.
- The behavior can affect any compliant server, proxy, or network path that terminates an SSE response cleanly.

### Suggested regression coverage

Add a transport test where the first successful SSE response emits an event and closes normally while the returned handle remains active. Verify that the adapter invokes the configured reconnect path, opens a replacement request, and continues delivering events through the original handle.

Also add a `ThreadStream` integration test covering two serial runs separated by a clean SSE EOF. Verify that the second run creates or uses a live event subscription and completes without rebuilding the client.

## System info

- Package when initially observed: `@langchain/langgraph-sdk@1.9.28`
- Latest stable release with the same code path: `@langchain/langgraph-sdk@1.9.29`
- Transport: thread-centric protocol over Fetch/SSE
- Endpoint: `POST /threads/{thread_id}/stream/events`
- Minimal reproduction verified on Linux x64 with Bun 1.3.13; the script uses Web-standard APIs available in Node.js 18 and newer.
- The issue is transport-level and does not depend on a specific UI framework.

original4422:
## Summary

- treat clean SSE EOF as an unexpected disconnect while the event-stream handle remains active
- route it through the existing reconnect callback, retry budget, and backoff path
- preserve terminal behavior for explicit handle or transport close
- add focused coverage for event delivery across reconnect, zero retry budget, and explicit close

## Root cause

`ProtocolSseTransportAdapter.openEventStream()` only reached its reconnect logic when stream consumption threw. A successful SSE response that ended normally closed the shared queue and returned, leaving the retained thread stream handle unable to observe later events.

The fix keeps explicit abort/shutdown terminal, but turns an otherwise clean EOF into the same existing unexpected-disconnect path. It does not introduce a second retry loop or change replay cursor behavior. PR #2689 modifies adjacent durable cursor handling and should compose with this narrow EOF-boundary change.

## Test plan

- `pnpm exec vitest run src/client/stream/transport/http.test.ts --reporter=verbose` (from `libs/sdk`): 20 passed, no type errors
- `pnpm --filter @langchain/langgraph-sdk test`: 53 files / 778 tests passed, no type errors
- `pnpm format:check`: passed (440 files)
- `pnpm lint`: passed (1,162 files, 0 warnings/errors)
- `pnpm --filter @langchain/langgraph-sdk build`: passed, including attw and publint
- `pnpm build`: 25/25 tasks passed

Fixes #2701

### DIFF
--- a/libs/sdk/src/client/stream/transport/http.test.ts
+++ b/libs/sdk/src/client/stream/transport/http.test.ts
@@ -499,7 +499,8 @@ describe("ProtocolSseTransportAdapter SSE reconnect with custom fetch", () => {
       attempt: 1,
       cause: expect.any(Error),
     });
-    expect(reconnectDelayMs).toHaveBeenCalledExactlyOnceWith(1);
+    expect(reconnectDelayMs).toHaveBeenCalledTimes(1);
+    expect(reconnectDelayMs).toHaveBeenCalledWith(1);
 
     await transport.close();
   });
--- a/libs/sdk/src/client/stream/transport/http.test.ts
+++ b/libs/sdk/src/client/stream/transport/http.test.ts
@@ -448,7 +448,7 @@ describe("ProtocolSseTransportAdapter SSE reconnect with custom fetch", () => {
   it("reconnects after clean EOF and continues on the original handle", async () => {
     let streamOpens = 0;
     const onReconnect = vi.fn();
-    const reconnectDelayMs = vi.fn(() => 0);
+    const reconnectDelayMs = vi.fn((_attempt: number) => 0);
     const fetchImpl = vi.fn(
       (input: URL | RequestInfo, init?: RequestInit) => {
         if (!String(input).includes("/stream/events")) {
@@ -502,6 +502,47 @@ describe("ProtocolSseTransportAdapter SSE reconnect with custom fetch", () => {
     expect(reconnectDelayMs).toHaveBeenCalledTimes(1);
     expect(reconnectDelayMs).toHaveBeenCalledWith(1);
 
+    await transport.close();
+  });
+
+  it("exhausts the consecutive-failure budget on empty 200 responses", async () => {
+    const onReconnect = vi.fn();
+    const reconnectDelayMs = vi.fn((_attempt: number) => 0);
+    const fetchImpl = vi.fn(() =>
+      Promise.resolve(
+        new Response(null, {
+          status: 200,
+          headers: { "content-type": "text/event-stream" },
+        })
+      )
+    ) as MockFetch;
+
+    const transport = new ProtocolSseTransportAdapter({
+      apiUrl: "http://localhost:8123",
+      threadId: THREAD_ID,
+      fetch: fetchImpl,
+      maxReconnectAttempts: 2,
+      reconnectDelayMs,
+      onReconnect,
+      idleReconnect: 0,
+    });
+
+    const handle = transport.openEventStream({ channels: ["values"] });
+    await handle.ready;
+    const iterator = handle.events[Symbol.asyncIterator]();
+
+    await expect(iterator.next()).rejects.toThrow(
+      "Protocol SSE stream ended unexpectedly."
+    );
+    expect(fetchImpl).toHaveBeenCalledTimes(3);
+    expect(onReconnect.mock.calls.map(([options]) => options.attempt)).toEqual([
+      1, 2,
+    ]);
+    expect(reconnectDelayMs.mock.calls.map(([attempt]) => attempt)).toEqual([
+      1, 2,
+    ]);
+
+    handle.close();
     await transport.close();
   });
 
--- a/libs/sdk/src/client/stream/transport/http.test.ts
+++ b/libs/sdk/src/client/stream/transport/http.test.ts
@@ -25,6 +25,32 @@ function streamEventBodies(fetchImpl: MockFetch): Record<string, unknown>[] {
     .filter((body): body is Record<string, unknown> => body != null);
 }
 
+function protocolEventResponse(
+  eventId: string,
+  seq: number,
+  options: {
+    keepOpen?: boolean;
+    onStart?: (
+      controller: ReadableStreamDefaultController<Uint8Array>
+    ) => void;
+  } = {}
+): Response {
+  const frame = `event: values\ndata: {"type":"event","method":"values","seq":${seq},"event_id":"${eventId}"}\n\n`;
+  return new Response(
+    new ReadableStream<Uint8Array>({
+      start(controller) {
+        controller.enqueue(new TextEncoder().encode(frame));
+        options.onStart?.(controller);
+        if (!options.keepOpen) controller.close();
+      },
+    }),
+    {
+      status: 200,
+      headers: { "content-type": "text/event-stream" },
+    }
+  );
+}
+
 describe("ProtocolSseTransportAdapter URL resolution", () => {
   it("preserves apiUrl path prefix for protocol commands", async () => {
     const { calls, fetch } = createFetchRecorder();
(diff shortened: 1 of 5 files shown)


libs/sdk/src/client/stream/transport/http.test.ts
```diff
@@ -419,6 +445,148 @@ describe("ProtocolSseTransportAdapter SSE reconnect with custom fetch", () => {
...
+    }
+
+    expect(received.map((message) => message.event_id)).toEqual(["e1", "e2"]);
+    expect(streamOpens).toBe(2);
+    expect(onReconnect).toHaveBeenCalledTimes(1);
+    expect(onReconnect).toHaveBeenCalledWith({
+      attempt: 1,
+      cause: expect.any(Error),
+    });
+    expect(reconnectDelayMs).toHaveBeenCalledExactlyOnceWith(1);
```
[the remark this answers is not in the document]
original4422: Thanks for flagging this. Vitest 4.1.10 did accept the matcher locally, but the standard call-count plus argument assertions are clearer and avoid compatibility concerns. Updated in f74e3d28; the focused 20-test file, typecheck, repository format check, and lint all pass.

libs/sdk/src/client/stream/transport/http.ts
```diff
@@ -330,8 +330,11 @@ export class ProtocolSseTransportAdapter implements TransportAdapter {
               streamQueue.push(event.data as Message);
             }
           }
-          streamQueue.close();
-          return;
+          if (ac.signal.aborted || this.closed) {
+            streamQueue.close();
+            return;
+          }
+          throw new Error("Protocol SSE stream ended unexpectedly.");
```
rrajpuro, line 342: Could we treat this as a consecutive-failure budget by resetting the reconnect counter only after the stream has demonstrated liveness—for example, after receiving an event/heartbeat or remaining connected for a minimum period? Resetting merely after a successful HTTP response could create an unbounded `200 → immediate EOF → reconnect` loop, and reconnect backoff should still apply.

`attempt` is scoped to the lifetime of this handle and only increments. A server, reverse proxy, load balancer, CDN, or firewall can terminate otherwise healthy SSE responses with clean EOF more than once over the lifetime of a long-lived thread. Since this transport defaults to five reconnect attempts, the same active handle eventually becomes permanently closed despite successful intermediate reconnections, recreating #2701 for a later run. Alternatively, marking and recreating a terminal event stream before the next run would keep the thread usable.

libs/sdk/src/client/stream/transport/http.test.ts
```diff
@@ -419,6 +445,149 @@ describe("ProtocolSseTransportAdapter SSE reconnect with custom fetch", () => {
     await transport.close();
   });
 
+  it("reconnects after clean EOF and continues on the original handle", async () => {
```
rrajpuro, line 448: This proves one transport-level reconnect, but could we also add the `ThreadStream` regression described in #2701? The test should keep one `ThreadStream`, complete run 1, let its underlying event response reach clean EOF, start run 2, and assert that a new `/stream/events` request is opened and the later event (ideally `input.requested`) is delivered through the same thread handle. It would also be useful to cover repeated clean EOFs beyond the default retry count before the later event; this test keeps the second response open, so it cannot catch lifetime retry-budget exhaustion.

rrajpuro: Could we extend this to a `StreamController`-level regression that exercises both SDK-managed SSE handles? Start the normal root content pump and the dedicated wildcard `["lifecycle", "input"]` watcher, let both responses end with clean EOF, then have run 2 deliver a root `values`/message event plus a nested `input.requested` event. Assert that the content is visible, `rootStore.interrupts` contains the nested interrupt, and both handles opened replacement requests. A single-handle `ThreadStream` test would not verify that the wildcard watcher reconnects.

original4422: Addressed in c5605761 with the requested StreamController-level regression. It uses the real ProtocolSseTransportAdapter and one ThreadStream, exercises both the root content pump and the wildcard lifecycle/input watcher through six event-bearing clean EOFs (beyond the configured default-sized budget of five), then submits run 2 and verifies a replacement request for each handle, visible root values/messages, and a nested input.requested entry in rootStore.interrupts. The separate empty-200 test also proves non-live 200→EOF loops remain bounded. Validation: focused controller+transport suite 77/77 with typecheck; full SDK 53 files/780 tests with typecheck; repository format and lint clean; SDK build and full 25-task monorepo build passed.

libs/sdk/src/client/stream/transport/http.ts
original4422: Addressed in c5605761. The retry counter now resets only after a decoded SSE event demonstrates liveness, so clean EOFs after healthy event delivery use a consecutive-failure budget across the handle lifetime. A 200 response alone does not reset it: the new empty-200 regression proves immediate EOF still exhausts maxReconnectAttempts and preserves attempt-based backoff/callback values. Validation: focused controller+transport suite 77/77 with typecheck; full SDK 53 files/780 tests with typecheck; repository format and lint clean; SDK build and full 25-task monorepo build passed.

original4422:
Hi @christian-bromann, when you have a chance, would you mind reviewing this PR and letting me know if any changes are needed? Thanks!

original4422:
Hi @christian-bromann, following up since the last review request—when you have a chance, could you please take another look and let me know whether this is ready to merge or needs changes? Thanks!

original4422:
Hi @christian-bromann, when you have a chance, could you please take another look and let me know whether this is ready to merge or needs any changes? Thank you.

original4422:
Hi @hntrl, when you have a chance, could you please take a look and let me know whether this PR is ready to merge or needs any changes? Thank you.

simon-lighthouse:
+1 for this issue; this is a blocker for our adoption of v3 streaming as we are losing the rendering of live messages in our implementation as a result. @hntrl / @christian-bromann

original4422:
Hi @hntrl, could you please take a look at this fix when you have a chance? I identified the current main-branch conflict in the controller-test imports; the branch still needs updating and revalidation before it can merge. Thank you.

original4422:
Hi @hntrl, when you have a chance, could you please check this PR and advise on the next steps? It is currently conflicting with `main`, and native/runtime verification plus full disk, RSS, and swap proof remain incomplete, so I’m not claiming it is ready to merge. Thank you.

rrajpuro:
### Related fix released and reproduction verified

The clean-EOF failure from #2701 is addressed by [#2812](https://github.com/langchain-ai/langgraphjs/pull/2812), merged on September 9, 2026 ([merge commit](https://github.com/langchain-ai/langgraphjs/commit/db4bdad61ddfc6c1113269b131ac2efde3eecf69)). The reconnect behavior is present in the published `@langchain/langgraph-sdk@1.11.0`.

The unchanged standalone reproduction from #2701 was rerun:

| SDK version | Stream requests | Reconnect attempts | Result |
| --- | --- | --- | --- |
| `1.9.29` | `1` | `[]` | Fails |
| `1.11.0` | `2` | `[1]` | Passes both assertions |

With `maxReconnectAttempts: 1` and `idleReconnect: 0`, SDK 1.11.0 reconnects after the first successful SSE response closes normally. This confirms that the original reported case is fixed in a released version.

For reviewing the overlap with this PR: #2812 also resets the retry budget after receiving an event, and deliberately preserves terminal clean-EOF behavior when `maxReconnectAttempts: 0`. Additional behavior and regression coverage in this PR can be assessed separately.

Verification here covers the standalone transport reproduction; it does not include end-to-end execution of two real graph runs.

PR #2717, opened 2026-08-20
docs: fix contributor setup and test commands

To3akaRin:
## Summary

- fix the npm package link and repository directory used in the contributor setup
- replace the removed `test:single` instructions with the current pnpm workspace and Vitest commands
- point integration-test setup to the existing `libs/langgraph-core/.env.example`

## Testing

- `pnpm --filter @langchain/langgraph build`
- `pnpm --filter @langchain/langgraph test src/graph/message.test.ts` (4 tests passed)
- `pnpm --filter @langchain/langgraph exec vitest list --mode int src/tests/chatbot.int.test.ts --filesOnly`
- `pnpm format:check`

Docs-only change. No public API changes and no changeset required.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,7 +49,7 @@ good code into the codebase.
 ### 🏭 Release process
 
 As of now, LangGraph has an ad hoc release process: releases are cut with high frequency via by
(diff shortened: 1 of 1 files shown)


christian-bromann (approved):
Thank you 👍

PR #2718, opened 2026-08-20
fix(sdk): validate protocol SSE content type

jstar0:
## Summary

A successful non-SSE response (for example, a JSON error page) could be treated as a ready protocol stream, leaving consumers with an empty event iterator. Validate the response Content-Type before resolving the stream readiness promise, and propagate the failure to the event iterator. Add regression coverage and a patch changeset.

## Verification

- `pnpm --filter @langchain/langgraph-sdk test src/client/stream/transport/http.test.ts` (17 passed)
- `pnpm exec oxfmt --check libs/sdk/src/client/stream/transport/http.ts libs/sdk/src/client/stream/transport/http.test.ts .changeset/steady-sse-validation.md`
- `pnpm exec oxlint libs/sdk/src/client/stream/transport/http.ts libs/sdk/src/client/stream/transport/http.test.ts`

### DIFF
--- a/.changeset/steady-sse-validation.md
+++ b/.changeset/steady-sse-validation.md
@@ -1,0 +1,5 @@
+---
+"@langchain/langgraph-sdk": patch
+---
+
+Reject protocol SSE stream responses with non-SSE content types before marking subscriptions ready.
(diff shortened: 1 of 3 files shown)


christian-bromann (approved):
LGTM 👍

PR #2726, opened 2026-08-21
feat(langgraph): add getRuntime helper for typescript

ilyaryabchinski:
So I was migrating my agent to typescript and realised I cannot get runtime info like in Python, so I wanted to have this possibility in the TS SDK and decided to make this PR

### DIFF
--- a/libs/langgraph-core/src/pregel/utils/config.test.ts
+++ b/libs/langgraph-core/src/pregel/utils/config.test.ts
@@ -8,6 +8,7 @@ import {
   getStore,
   getWriter,
   getConfig,
+  getRuntime,
   recastCheckpointNamespace,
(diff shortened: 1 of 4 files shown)


christian-bromann:
I don't think we want to encourage users to import a `getRuntime` method to access runtime properties. This seems non-idiomatic to TypeScript and most primitives, e.g. tools, middleware or graph nodes, get this as 2nd argument. While we are trying to stay interface compliant to Python we DO have some idiomatic differences.

I would argue you don't need this. Maybe I am missing something here, could you provide an example where you are missing this?

ilyaryabchinski:
I think you're right, I tried one more time using this second parameter and it covers everything, thank you!

Closing this.

PR #2735, opened 2026-08-23
fix(sdk): keep discovered subgraph messages out of root state

### LINKED ISSUE #2716: StreamController merges messages from a discovered depth-one subgraph into root messages
### Checked other resources

### Example Code

Create an empty directory with these two files. This uses a synthetic protocol adapter, so it requires no server, model, API key, or network access after installation.

`package.json`:

```json
{
  "name": "langgraph-root-message-repro",
  "private": true,
  "type": "module",
  "scripts": {
    "reproduce": "node reproduce.mjs"
  },
  "dependencies": {
    "@langchain/core": "1.2.9",
    "@langchain/langgraph-sdk": "1.9.30"
  }
}
```

`reproduce.mjs`:

```javascript
import assert from "node:assert/strict"
import { Client } from "@langchain/langgraph-sdk"
import { StreamController } from "@langchain/langgraph-sdk/stream"

const namespace = ["worker:00000000-0000-0000-0000-000000000001"]
const baseParams = { namespace, node: "model", timestamp: 0 }
const protocolEvents = [
  {
    type: "event",
    event_id: "event-1",
    seq: 1,
    method: "values",
    params: { ...baseParams, data: { messages: [] } },
  },
  {
    type: "event",
    event_id: "event-2",
    seq: 2,
    method: "messages",
    params: {
      ...baseParams,
      data: { event: "message-start", role: "ai", id: "child-message" },
    },
  },
  {
    type: "event",
    event_id: "event-3",
    seq: 3,
    method: "messages",
    params: {
      ...baseParams,
      data: {
        event: "content-block-start",
        index: 0,
        content: { type: "text", text: "" },
      },
    },
  },
  {
    type: "event",
    event_id: "event-4",
    seq: 4,
    method: "messages",
    params: {
      ...baseParams,
      data: {
        event: "content-block-delta",
        index: 0,
        delta: { type: "text-delta", text: "child-only output" },
      },
    },
  },
  {
    type: "event",
    event_id: "event-5",
    seq: 5,
    method: "messages",
    params: {
      ...baseParams,
      data: {
        event: "content-block-finish",
        index: 0,
        content: { type: "text", text: "child-only output" },
      },
    },
  },
  {
    type: "event",
    event_id: "event-6",
    seq: 6,
    method: "messages",
    params: { ...baseParams, data: { event: "message-finish" } },
  },
]

let rootFilter
const adapter = {
  threadId: "thread-1",
  setThreadId(threadId) {
    this.threadId = threadId
  },
  async open() {},
  async send() {},
  async *events() {},
  async getState() {
    return null
  },
  openEventStream(params) {
    rootFilter = params
    return {
      ready: Promise.resolve(),
      events: (async function* () {
        yield* protocolEvents
      })(),
      close() {},
    }
  },
  async close() {},
}

const controller = new StreamController({
  assistantId: "test-agent",
  client: new Client({ apiUrl: "http://unused.invalid" }),
  threadId: adapter.threadId,
  transport: adapter,
})

await controller.hydrationPromise
await new Promise((resolve) => setTimeout(resolve, 20))

assert.deepEqual(rootFilter.namespaces, [[]])
assert.equal(rootFilter.depth, 1)

const subgraphs = [...controller.subgraphStore.getSnapshot().values()]
const rootMessages = controller.rootStore.getSnapshot().messages

console.log({
  discoveredSubgraphs: subgraphs.map((subgraph) => subgraph.namespace),
  rootMessageIds: rootMessages.map((message) => message.id),
})

assert.equal(subgraphs.length, 1, "The namespace should be discovered as a subgraph")
assert.equal(
  rootMessages.length,
  0,
  "A message scoped to a discovered subgraph should not enter root messages",
)

await controller.dispose()
```

Run:

```shell
npm install
npm run reproduce
```

```text
{
  discoveredSubgraphs: [ [ 'worker:00000000-0000-0000-0000-000000000001' ] ],
  rootMessageIds: [ 'child-message' ]
}

AssertionError [ERR_ASSERTION]: A message scoped to a discovered subgraph should not enter root messages

1 !== 0
```

### Description

`StreamController` discovers the synthetic `worker:*` namespace as a subgraph from its namespaced `values` event. However, the `messages` events from that same non-root namespace are also assembled into `controller.rootStore.messages`.

The root pump subscribes to `[[]]` with `depth: 1`. Its message routing excludes namespaces containing `task:*` or `tools:*`, but it treats every other depth-one namespace as part of the root message projection. Consequently, an ordinary first-level subgraph such as `worker:*` can appear in a coordinator chat transcript even though its namespace is separately discoverable and can be consumed through a scoped message projection.

I expect messages belonging to a discovered non-root subgraph to remain in that subgraph's scoped projection rather than being merged into root messages solely because the subgraph is one level deep. If some orchestrator-owned depth-one namespaces must remain visible at root, the routing needs a generic way to distinguish or exclude other discovered subgraphs rather than relying only on the `task:*` / `tools:*` naming conventions.

This is observable while messages stream and can cause downstream chat runtimes that render `stream.messages` to display internal subgraph output in the main transcript.

### System Info

- `@langchain/langgraph-sdk@1.9.30` (latest stable when reproduced)
- `@langchain/core@1.2.9`
- Node.js `v24.19.0`
- npm `12.0.2`
- Linux `x86_64`

dawNotPoi:
## Summary

- exclude message events from exact namespaces already confirmed as subgraph hosts
- preserve messages from ordinary root-owned depth-one node namespaces
- add a StreamController regression test covering discovery and root routing

## Testing

- `pnpm --filter @langchain/langgraph-sdk test`
- `pnpm --filter @langchain/langgraph-sdk build`
- `pnpm lint`
- `pnpm format:check`
- `pnpm build`

Fixes #2716

### DIFF
--- a/.changeset/fix-subgraph-root-message-leak.md
+++ b/.changeset/fix-subgraph-root-message-leak.md
@@ -1,0 +1,9 @@
+---
+"@langchain/langgraph-sdk": patch
+---
+
+fix(sdk): keep discovered subgraph messages out of root state
+
+Messages scoped to a confirmed subgraph namespace no longer appear in
+`rootStore`, while ordinary root-owned depth-one nodes continue to stream
+normally.
--- a/libs/sdk/src/stream/controller.test.ts
+++ b/libs/sdk/src/stream/controller.test.ts
@@ -112,14 +112,18 @@ function inputRequestedEvent(
   } as Event;
 }
 
-function valuesEvent(messages: unknown[], seq: number): Event {
+function valuesEvent(
+  messages: unknown[],
+  seq: number,
+  namespace: string[] = []
+): Event {
   return {
     type: "event",
     event_id: `values-${seq}`,
     seq,
     method: "values",
     params: {
-      namespace: [],
+      namespace,
       timestamp: 0,
       data: {
         messages,
@@ -156,14 +160,19 @@ function lifecycleEvent(event: string, seq: number): Event {
   } as Event;
 }
 
-function messageStartEvent(id: string, seq: number, role = "ai"): Event {
+function messageStartEvent(
+  id: string,
+  seq: number,
+  role = "ai",
+  namespace: string[] = []
+): Event {
   return {
     type: "event",
     event_id: `msg-start-${id}-${seq}`,
     seq,
     method: "messages",
     params: {
-      namespace: [],
+      namespace,
       timestamp: 0,
       data: { event: "message-start", id, role },
     },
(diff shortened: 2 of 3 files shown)


rrajpuro:
@dawNotPoi Did this not address the issue?

PR #2747, opened 2026-08-26
fix(langgraph): pushMessage emits on the streamEvents v3 messages channel

t3s7r:
## Problem

`pushMessage` looks up the callback handler by the name `StreamMessagesHandler`. `streamEvents({ version: "v3" })` registers `StreamProtocolMessagesHandler` instead, so on the v3 path (which `@langchain/langgraph-api` uses to serve run streams) manually pushed messages never reach the live `messages` channel. The state-persistence half still runs, so the pushed messages only appear in `values` snapshots at node end — the live emission is lost silently.

Reproduction: a node that calls `pushMessage(new AIMessage({ id: "1", content: "First" }), config)` emits nothing on the `messages` channel under `streamEvents(..., { version: "v3" })`, while the same graph under `graph.stream(..., { streamMode: "messages" })` emits the tuple as documented.

## Fix

`pushMessage` now finds either handler and delivers through the v3 handler's `emitFinalMessage` (made public, mirroring the classic handler's public `_emit`), producing the standard `message-start` / content block / `message-finish` sequence. No behavior change on the classic path.

## Test

Adds a regression test to `message.test.ts`: pushed message must surface as `message-start`/`message-finish` events on the v3 protocol stream. Red on current main, green with the fix; the four existing tests still pass.

### DIFF
--- a/.changeset/tender-showers-mate.md
+++ b/.changeset/tender-showers-mate.md
@@ -1,0 +1,5 @@
+---
+"@langchain/langgraph": patch
+---
+
+fix(langgraph): pushMessage emits on the streamEvents v3 messages channel
--- a/libs/langgraph-core/src/graph/message.test.ts
+++ b/libs/langgraph-core/src/graph/message.test.ts
@@ -65,4 +65,31 @@ describe("pushMessage", () => {
 
     expect(values).toEqual(messages);
   });
+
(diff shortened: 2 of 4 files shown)


christian-bromann (approved):
LGTM 👍

PR #2751, opened 2026-08-27
fix(api): run.start on a cancelled thread must not fold input into a resume

t3s7r:
## Problem

`createOrResumeRun` decides input-vs-resume like this:

```ts
const isResume =
  params.input != null &&
  ((currentRun != null && currentStatus === "interrupted") ||
    hasPendingInterrupts);
```

A **cancelled** run also ends with status `"interrupted"` — the same status a genuine `interrupt()` pause has — but the thread has no pending interrupt to consume a resume value. LangGraph only delivers resume values to an `interrupt()` call, so when a client stops a run (`stop()` → `runs.cancel`) and then sends a follow-up message, that message is folded into `Command(resume: input)` that nothing reads: the message is silently dropped and the interrupted node re-runs with stale state.

Reproduction (SDK `useStream` against `langgraphjs dev`): start a run on a graph with a long-running node, `stop()` mid-node, then `submit({ messages: [...] })`. The created run has `input: null` and `command.resume` carrying the message; the thread's final state contains no trace of it.

## Fix

Key the decision on `hasPendingInterruptsForThread` alone (already computed in the same scope; it reads thread-state tasks with `subgraphs: true`):

- thread paused at a real `interrupt()` → still resumes (unchanged),
- cancelled thread → plain input, i.e. a fresh run from the checkpoint that applies the message to state.

Applied to both copies of the condition: `protocol/service.mts` and `experimental/embed/protocol.mts`.

## Test

`tests/protocol-v2/run-start-cancelled.test.mts` drives `ProtocolService.handleCommand("run.start")` with stub bindings: the cancel case (interrupted status, no pending interrupts) must submit input — red on current main; the pending-interrupt case must still produce `Command(resume)` — green before and after. `pnpm typecheck` clean; the rest of the suite unchanged (the three `fetch failed` files fail identically on clean main in my sandbox).


t3s7r:
The failing `Unit Tests` job is unrelated to this diff: `26.x` on the runners moved from Node 26.7.0 to 26.8.1 on Aug 26/27, and `tracing.test.ts > does not double stream tokens from a nested graph under tracing` fails deterministically on 26.8.x on unmodified `main` (passes on 22.x and 26.7.0). Details and local bisection across Node versions in #2752. The earlier `sdk-svelte` failure was a one-off and passed on re-run; everything this diff touches is green.

christian-bromann (approved):
LGTM 👍

PR #2753, opened 2026-08-27
ci: pin Node 26 tracing test job to 26.7.0

xianjianlf2:
## Summary
- pin the affected Ubuntu Node 26 CI matrix entry to 26.7.0
- retain Node 22/24 plus macOS and Windows matrix coverage
- document that Node 26.8.x deterministically drops the final tracing token

## Validation
- `git diff --check`
- parsed the workflow as YAML

The focused tracing test requires the full pnpm workspace install, which did not complete in this environment. This is intended as a temporary, reversible CI workaround until the upstream Node regression is resolved.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,7 +114,10 @@ jobs:
       matrix:
         # See Node.js release schedule at https://nodejs.org/en/about/releases/
         os: [ubuntu-latest]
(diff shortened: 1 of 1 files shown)


christian-bromann:
I don't think this is needed. The CI passes just fine. Thanks for the contribution though!

Issue #2752, opened 2026-08-27
CI: tracing.test.ts 'does not double stream tokens' fails on Node 26.8.x — all PR unit-test jobs red since runners moved off 26.7.0


t3s7r:
## Summary

`libs/langgraph-core` `src/tests/tracing.test.ts > does not double stream tokens from a nested graph under tracing` fails deterministically on Node **26.8.x**, which the `Unit Tests (ubuntu-latest, 26.x)` CI job now resolves to. Every PR CI run since the runners picked up 26.8.x fails this job (fail-fast then cancels the other unit-test jobs), regardless of the PR's content.

```
AssertionError: expected 'Hello world from langgrap' to be 'Hello world from langgraph'
 ❯ src/tests/tracing.test.ts:250:23
```

The streamed inner text is consistently exactly one token short.

## Evidence

- Reproduced locally on current `main` (dd287b4, no other changes), same failure on `main` as of Aug 24 (01f61f9) — so no recent langgraphjs commit caused it:
  - Node **22.20.0**: 7 passed
  - Node **26.7.0**: 7 passed
  - Node **26.8.1**: 1 failed (always the same one-token-short result)
- CI timeline: the last green `26.x` job I found (PR #2742, Aug 26 12:16 UTC) ran on **26.7.0**; runs from Aug 27 onward acquire **26.8.1**. Node 26.8.0/26.8.1 were released 2026-08-26.

First seen on the CI runs of #2751 (twice in a row, identical failure), where the diff only touches `libs/langgraph-api`.

## Notes

26.8.1 over 26.8.0 is only a version-string fix, so the behavior change is in 26.8.0. Whether this is a Node regression (timing of the final token emission under tracing callbacks) or the test depending on fragile scheduling, I haven't isolated. Until then, pinning the matrix to `26.7.0` would unblock PR CI.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

xianjianlf2:
I reproduced the failure on Node 26.8.x and prepared a focused CI workaround that pins the affected Ubuntu Node 26 matrix entry to 26.7.0, the last known-good release, while preserving the other matrix coverage. I would like to take this issue and open a PR.

PR #2762, opened 2026-08-31
fix(sdk): send checkpoint_id in runs.stream() request body

JessYanCoding:
## Summary

`RunsStreamPayload` extends `RunsInvokePayload` (`libs/sdk/src/types.ts:171`, `:48`), which declares `checkpointId` — *"Checkpoint ID for when creating a new run"* (`types.ts:70-73`). But `RunsClient.stream()` builds its wire body at `runs/index.ts:79-101` and includes `checkpoint` while omitting `checkpoint_id`.

So this type-checks, compiles, and silently runs from the thread's latest state instead of forking from the requested checkpoint:

```ts
client.runs.stream(threadId, assistantId, { input, checkpointId });
```

No error is raised, and the caller gets no signal that the checkpoint was ignored.

## Reproduction

The same `checkpointId` passed to all three methods on a real `Client`, with a stub `callerOptions.fetch` capturing the outgoing request body:

| endpoint | `checkpoint_id` in body — before | after |
|---|---|---|
| `/threads/t1/runs/stream` | **false** | true |
| `/threads/t1/runs` | true | true |
| `/threads/t1/runs/wait` | true | true |

## Why this looks like a plain omission rather than intent

1. `RunsClient.create()` (`runs/index.ts:151`) and `RunsClient.wait()` (`:240`) in the same class both send `checkpoint_id`.
2. The server validates `/runs/stream` with the same `schemas.RunCreate` as `/runs` and `/runs/wait` — `libs/langgraph-api/src/api/runs.mts:181, 257, 268, 342, 383`.
3. The field is declared on the very payload type the method accepts.

The field was simply missing from one of three otherwise identical bodies.

## The change

One line, placed exactly where `create()` has it (immediately after `checkpoint`):

```ts
checkpoint_id: payload?.checkpointId,
```

## Tests

Added `sends checkpointId as checkpoint_id in the request body` to the existing `describe("runs.stream()")` block in `src/client/runs/stream.test.ts`, next to the existing body-shape assertions and reusing that file's SSE harness.

**Reverse-verified**: with the one-line change reverted, the new test fails (`expected undefined to be '1f0aaaaa-…'`). With it applied, `pnpm test` in `libs/sdk` gives **53 files / 779 tests passing, no type errors**; `pnpm lint` and `pnpm format:check` are clean at the repo root.

A changeset is included (`@langchain/langgraph-sdk`: patch).

## Note

`stream()` also omits `langsmith_tracer`, which `create()` and `wait()` both send. I left it out to keep this change to a single field — happy to include it here or open a follow-up, whichever you prefer.

---

*This contribution was prepared with AI assistance.*


JessYanCoding:
The five red `Unit Tests` legs are unrelated to this PR.

Only one test actually fails, and it is on the `26.x` leg in a package this PR does not touch — `libs/langgraph-core/src/tests/tracing.test.ts:250`:

```
AssertionError: expected 'Hello world from langgrap' to be 'Hello world from langgraph'
Expected: "Hello world from langgraph"
Received: "Hello world from langgrap"
```

`1 failed | 1646 passed` — the final streamed token character is dropped. That is #2752, and #2753 pins the `26.x` matrix leg to Node 26.7.0. The other four legs end with `The operation was canceled`, i.e. fail-fast cancellation rather than failures of their own.

In the same run, `libs/sdk` — the only package this PR changes — reports `53 passed (53)`, and `Check linting`, `Check formatting`, `Build`, all seven `Environment Tests` and all four `Browser Tests` are green.

Happy to rebase once #2753 lands, if that would help.

christian-bromann (approved):
LGTM 👍

PR #2765, opened 2026-08-31
Fix Docker path detection on Linux

hari-dhanushkodi:
Detect Docker using the current process `PATH` before falling back to a login shell or known installation paths. This preserves Linux CI/container paths, removes the macOS-only `dscl` dependency, and adds common `/usr/local` locations to the fallback.

Added focused coverage for finding Docker through the process environment and a patch changeset.

### Verification
- `pnpm --dir libs/langgraph-cli exec vitest run tests/shell.test.mts --no-color`
- `pnpm --dir . exec oxlint libs/langgraph-cli/src/docker/shell.mts libs/langgraph-cli/tests/shell.test.mts`
- `pnpm --dir libs/langgraph-cli build`
- `pnpm --dir libs/langgraph-cli typecheck` *(blocked by pre-existing unresolved workspace packages `@langchain/langgraph-api` and `create-langgraph`; the package build, which builds workspace dependencies, succeeds)*

Made by [Open SWE](https://openswe.vercel.app/agents/01a05891-b125-766d-bd43-e1513978b9a8)

### DIFF
--- a/.changeset/lucky-paths-build.md
+++ b/.changeset/lucky-paths-build.md
@@ -1,0 +1,5 @@
+---
+"@langchain/langgraph-cli": patch
+---
+
+Detect Docker using the current process PATH on Linux and other platforms.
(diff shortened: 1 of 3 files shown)


sz-lc-mdm-captive (changes requested):
instead of just adding a fix and preserving the original behavior, i think we should drop this fix and go with #2872 : 
* it removes `docker/shell.mts` entirely, simplifying
* it aligns behavior with the python CLI implementation
* it should also fix support for windows

PR #2767, opened 2026-08-31
fix(sdk): accumulate tool-call args across block-delta events

### LINKED ISSUE #2766: [SDK] block-delta overwrites tool_call_chunk args, so streaming tool calls drop out of message.tool_calls
### Checked other resources

### Description

`applyCoreEventDelta` in `libs/sdk/src/client/stream/messages.ts` accumulates on `text-delta`, `reasoning-delta` and `data-delta`, but its `block-delta` case is a shallow spread:

```ts
case "block-delta":
  return {
    ...(current ?? {}),
    ...event.delta.fields,
  } as CoreContentBlock;
```

`block-delta` is the shape the Python v3 emitter uses for tool-call arguments, and each event carries only the newest `args` fragment. The spread therefore **replaces** the block's `args` instead of appending to it, so a `tool_call_chunk` block never holds more than one fragment.

Downstream that silently removes the call from the message. `assembledToBaseMessage` builds `tool_call_chunks` from the blocks, `AIMessageChunk`'s constructor rebuilds `tool_calls` from `collapseToolCallChunks(tool_call_chunks)`, and a group whose joined text does not `parsePartialJson` is moved to `invalid_tool_calls` — dropping out of `tool_calls` entirely. A bare fragment such as `, "end_time":` parses as nothing, so **for every delta whose fragment is not self-contained JSON, `message.tool_calls` is missing that call**. Accumulated fragments would not have this problem: a growing prefix of valid JSON parses at every length, which is the invariant `collapseToolCallChunks` is written against (see #11148 / #11150 in `langchainjs`, which are about it re-parsing exactly that accumulated string).

Two things suggest this is an oversight rather than intent:

1. `applyCoreContentDelta`, the sibling branch in the same file taken when the event carries `content` instead of `delta`, already does the right thing and even preserves `id`/`name` across fragments that omit them:

   ```ts
   case "tool_call_chunk":
   case "server_tool_call_chunk": {
     const merged = { ...target, ...delta } as Record<string, unknown>;
     if (delta.id == null && "id" in target && target.id != null) merged.id = target.id;
     if (delta.name == null && "name" in target && target.name != null) merged.name = target.name;
     merged.args = `${("args" in target ? target.args : "") ?? ""}${delta.args ?? ""}`;
     return merged as unknown as CoreContentBlock;
   }
   ```

2. `data-delta`, immediately above the `block-delta` case in `applyCoreEventDelta`, concatenates rather than replaces.

So the fix appears to be routing `block-delta` for `tool_call_chunk` / `server_tool_call_chunk` through the same concatenation, or generalising it in `applyCoreEventDelta`.

**Impact.** In a React UI reading `message.tool_calls` to render tool cards, each affected call disappears and reappears — one frame per non-parsing fragment. With parallel tool calls the fragments interleave, so several cards blink at once. We are working around it by reading the calls off `message.content` (the `tool_call_chunk` blocks keep their `id` and `name` through the overwrite) and using `tool_calls` only as the source of parsed args, but that leaves the accumulated args string unrecoverable — it was already overwritten by the time the client sees it.

I'm happy to open a PR if the direction above looks right.

### Example Code

Verbatim from an SSE capture against a Python LangGraph server (`content-block-*` on the `messages` channel, one assistant message, `args` field only):

```
seq=25  content-block-start   index=1  args=''
seq=26  content-block-delta   index=1  args='{"start_time": "2026-08-31T12:00:00Z"'
seq=27  content-block-delta   index=1  args=', "end_time":'
seq=28  content-block-delta   index=1  args=' "2026-08-31T13:00'
seq=29  content-block-delta   index=1  args=':00Z"}'
seq=32  content-block-finish  index=1  args={'start_time': '2026-08-31T12:00:00Z', 'end_time': '2026-08-31T13:00:00Z'}
```

Each delta arrives as:

```json
{"event": "content-block-delta", "index": 1,
 "delta": {"type": "block-delta",
           "fields": {"type": "tool_call_chunk",
                      "id": "toolu_019ez5oFy6Ax4S7DETSXaC8c",
                      "name": "get_topology_unhealthy",
                      "args": ", \"end_time\":"}}}
```

Observed on the assembled message:

| after | block's `args` | `message.tool_calls` |
|---|---|---|
| seq 26 | `{"start_time": "2026-08-31T12:00:00Z"` | call present (partial args) |
| seq 27 | `, "end_time":` | **call absent** — in `invalid_tool_calls` |
| seq 28 | ` "2026-08-31T13:00` | **call absent** |
| seq 29 | `:00Z"}` | **call absent** |
| seq 32 | *(finalized object)* | call present |

Expected: `args` accumulates to `{"start_time": "2026-08-31T12:00:00Z", "end_time": "2026-08-31T13:00:00Z"` and the call is present in `tool_calls` at every step.

No error is raised — the call is moved to `invalid_tool_calls` silently.

### System Info

- `@langchain/langgraph-sdk` 1.10.0 (current `latest`, published 2026-08-26); also present unchanged on `main` at `a5089cda1`
- `@langchain/core` 1.2.9
- Node v22.18.0, macOS 26.5.2 arm64
- Server: Python LangGraph, Agent Streaming Protocol v3

darkobardic:
## Summary

Fixes #2766. `applyCoreEventDelta`'s `block-delta` case is a shallow spread, so it replaces a `tool_call_chunk` block's `args` with each incoming fragment instead of appending. `block-delta` is the shape the Python v3 emitter uses for tool-call arguments, so the block never holds more than one fragment: `collapseToolCallChunks` cannot parse it, `AIMessageChunk`'s constructor moves the call to `invalid_tool_calls`, and it drops out of `message.tool_calls` for every fragment that is not self-contained JSON. In a UI that renders `tool_calls`, the tool card blinks out and back — several at once when tool calls run in parallel.

`applyCoreContentDelta`, the sibling branch taken when the event carries `content` rather than `delta`, already does the right thing: it concatenates `args` and preserves `id`/`name` across fragments that omit them (added for this same class of bug — see the comment in `applyContentDelta`). This routes `block-delta` through it for chunk blocks. `data-delta`, immediately above the `block-delta` case, likewise accumulates, so this looks like an oversight rather than intent.

## Changes

- `libs/sdk/src/client/stream/messages.ts` — in `applyCoreEventDelta`, when `current` and `event.delta.fields` are both `tool_call_chunk` / `server_tool_call_chunk`, delegate to `applyCoreContentDelta` instead of spreading. Every other block type, and a chunk's first fragment (no `current` yet), is byte-identical to before.
- One change covers both call sites: `toStreamingMessageHandle`'s event loop and `MessageAssembler.consume`, which is what `RootMessageProjection` reads.

## Test plan

- [x] New test in `libs/sdk/src/client/stream/messages.test.ts` replays the four `block-delta` fragments from #2766 verbatim and asserts, after each one, that the block's `args` equals the accumulated prefix and that the call is present in `tool_calls` with `invalid_tool_calls` empty.
- [x] Confirmed it fails on `main` — at the second fragment, `args` is `, "end_time":` rather than the accumulated string.
- [x] `pnpm --filter @langchain/langgraph-sdk test` — 53 files, 779 tests, no type errors.
- [x] `pnpm lint`, `pnpm format:check`, `pnpm --filter @langchain/langgraph-sdk build` (attw + publint clean).

Happy to adjust the shape — e.g. moving the chunk-accumulation into `applyCoreContentDelta`'s caller, or handling it in `coreContentBlockFromDelta` instead — if you'd prefer it elsewhere.


darkobardic:
Wrong premise — `block-delta.fields.args` is cumulative by contract, so this would have corrupted args for conforming servers. The fragments were produced by our own stream middleware; fixed there. Thanks @L4XB for catching it. See #2766.

PR #2778, opened 2026-09-02
fix(langgraph): type-error addNode channel name collisions

### LINKED ISSUE #2724: DX: node/channel name collisions in `addNode` are only caught at runtime — statically known collisions could be a type error
## Current behavior

`addNode` rejects a key that collides with a state channel, but only at graph-construction time:

```ts
import { Annotation, StateGraph } from "@langchain/langgraph";

const State = Annotation.Root({
  judge: Annotation<string>(),
});

new StateGraph(State).addNode("judge", (state) => ({ judge: state.judge }));
```

```
$ bun run repro.ts
error: judge is already being used as a state attribute (a.k.a. a channel), cannot also be used as a node name.
      at addNode (.../@langchain/langgraph/dist/graph/state.js:350:40)

$ tsc --noEmit
(no output, exit 0)
```

Verified with `@langchain/langgraph@1.4.12`, `@langchain/core@1.2.9`, TypeScript 6.0.3. Guard in current main: [`state.ts#L1072-L1076`](https://github.com/langchain-ai/langgraphjs/blob/12cc3f5d357dc94ac9df183ef05836ff4e83e8d2/libs/langgraph-core/src/graph/state.ts#L1072-L1076).

The check itself is clearly intentional and the message is good. The pain is *when* it fires: graphs are typically built at module top level, so the collision surfaces as a module-evaluation crash — in our case a dev watch loop kept dying on import until we traced it to a channel named `judge` and a node named `judge`. All the information needed to catch this lives in the type system already, since `StateGraph`'s generics carry the state/input/output definitions.

We also couldn't find the constraint documented — the [Graph API overview](https://docs.langchain.com/oss/javascript/langgraph/graph-api) describes the channel union and per-node private channels, but not that node names share a namespace with channel names.

## What we verified (and one trap worth flagging)

Before proposing anything we prototyped a typed signature against 1.4.12 and tested it across: literal collisions on state keys, **input/output schema keys**, legal literals, wide `string` keys, `Record<string, unknown>` schemas, the deprecated `{ channels }` constructor, per-node `options.input`, subgraph actions, and `Command`-returning actions.

The trap: a naive `K extends keyof State ? never : K` is **not** enough and would give false confidence. The runtime channel set is the union of the state, `input`, and `output` schemas, plus private channels added by earlier `addNode(..., { input })` calls. Counterexample — this passes the naive type check and still throws:

```ts
const State  = Annotation.Root({ shared: Annotation<string>() });
const Input  = Annotation.Root({ incoming: Annotation<string>() });
const Output = Annotation.Root({ outgoing: Annotation<string>() });

new StateGraph({ state: State, input: Input, output: Output })
  .addNode("incoming", fn); // keyof S is "shared" → naive check passes; runtime throws
```

One compatibility subtlety matters for any implementation: today the runtime checks the key against the channels *before* merging the current call's `options.input` schema, so a node whose name collides with its **own** input key is legal. The type check has to use the pre-call channel set, or it would silently change behavior.

## Proposal

**MVP (the actual ask):** reject collisions with the statically known constructor-time channel keys (state ∪ input ∪ output literals), across all `addNode`/`addSequence` overloads. Wide `string` keys and index-signature schemas keep passing — the runtime guard stays authoritative for anything the types can't see. Sketch that compiles and behaves as described on 1.4.12:

```ts
type KnownStringKeys<T> = string extends keyof T ? never : Extract<keyof T, string>;
// `string extends keyof T` guards the deprecated { channels } constructor and
// Record<string, ...> schemas, which surface a wide index signature.

type GraphChannelKeys<S, I, O> =
  | KnownStringKeys<S>
  | KnownStringKeys<ExtractStateType<I>>
  | KnownStringKeys<ExtractStateType<O>>;

type AllowedNodeKey<Channels extends string, K extends string> =
  string extends K ? K            // dynamic keys stay allowed
  : K extends Channels ? never    // statically known collision → compile error
  : K;
```

**Optional extension (fine to defer or drop):** accumulate a `ChannelNames` generic through the builder's return type so private channels introduced by earlier `addNode(..., { input })` calls also participate in the check. This buys full static equivalence with the runtime guard, at the cost of an extra generic on every chained call — whether that's worth the type-complexity budget is your call; the MVP alone already catches the common case.

Suggested test matrix beyond the above scenarios: union-literal keys (distributive conditionals silently narrow instead of erroring — worth a documented decision), Annotation/StateSchema/Zod state definitions, and a type-perf baseline (`tsc --extendedDiagnostics`) on long `addNode` chains.

## Docs

Independent of the type change, one sentence in the Graph API page would have saved us the debugging session: node names share a namespace with the state channel names known at that point in graph construction, and collisions throw synchronously in `addNode`.

Happy to submit a PR for the MVP tier (and the docs line) if the direction sounds right. And happy to move this to the LangChain Forum if that's the preferred venue for feature proposals.

ryanzillini:
Fixes #2724

`StateGraph.addNode` already rejects a node key that collides with a state/channel name, but only at graph-construction time. This PR makes statically knowable collisions a compile-time type error across the `addNode` / `addSequence` overloads.

## What changed

Constructor-time channel keys (the union of **state**, **input**, and **output** schema literals) are now rejected as node names by the type checker. The existing runtime guard is unchanged and stays authoritative for anything types cannot see.

The check lives on the `action` / `nodes` argument (`IfAllowedNodeKey`), not on the `key` parameter. Putting it on `key` (or constraining `K`) either blocked inferring the node name from the literal (so `ends` stole `K`) or dropped the `StateGraph` overload so `Graph.addNode` accepted the call.

- Naive `keyof State` is not used — input/output schema keys that are not in state also error (the `incoming` trap from #2724).
- Wide `string` keys, `Record` / index-signature schemas, and the deprecated `{ channels }` constructor keep compiling.
- A node whose name collides with its **own** `options.input` key remains legal (runtime checks the pre-call channel set).
- Union-literal keys (`"channel" | "ok"`) are rejected rather than silently narrowed. Documented in the new tests.
- One-sentence Graph API note: node names share a namespace with those channel names.

No extra `ChannelNames` generic — private channels from earlier `addNode(..., { input })` calls are deferred, as #2724 suggested.

## Test plan

- [x] New unit/type tests in `libs/langgraph-core/src/graph/state.addNode.channels.test.ts`
- [x] Existing `graph.test.ts` collision runtime test still throws (now also `@ts-expect-error`)
- [x] `pnpm --filter @langchain/langgraph test` on the new file plus `graph.test.ts`, `state.test.ts`, `types.test-d.ts`, `state_graph_compat.test-d.ts`, `pregel.test-d.ts`, `errors.test.ts`, and `pregel.test.ts`


jsiu93 (commented):
Issue author here. Thanks for picking this up. I went through the diff and the new test file against the scenarios listed in #2724, and the boundary is exactly the MVP tier I proposed: state ∪ input ∪ output literal keys rejected, wide `string` / index-signature / `{ channels }` keys still allowed, own-`options.input` collisions still legal, and no `ChannelNames` generic. The `pregel.test.ts` rename is a good catch too: that test was exercising the channel-collision guard rather than the duplicate-step-name guard, since `foo` is a channel there.

**One non-blocking suggestion: the error text users see.** Because the rejection branch resolves to `never`, the compiler message never mentions channels. This is what `tsc` prints against the pkg.pr.new build of this PR:

```
probe.ts(4,40): error TS2769: No overload matches this call.
  The last overload gave the following error.
    Argument of type '[(s: any) => { judge: any; }]' is not assignable to parameter of type 'never'.
probe.ts(4,41): error TS7006: Parameter 's' implicitly has an 'any' type.
```

The second line is a knock-on: once the parameter type collapses to `never`, the callback loses contextual typing, so the user gets two errors and neither says what went wrong. A branded template literal in the error branch keeps the check identical and turns the message into the diagnosis. I patched the shipped `state.d.cts` with this shape and re-ran the same probe:

```ts
type NodeChannelCollision<K extends string> =
  `Node name "${K}" is already a state channel name; pick a different node name`;

type IfAllowedNodeKey<K extends string, S, I, O, Allowed> =
  string extends K ? Allowed
  : [K & GraphChannelKeys<S, I, O>] extends [never] ? Allowed
  : Allowed extends unknown[]
    ? [error: NodeChannelCollision<K & GraphChannelKeys<S, I, O>>,
       options?: NodeChannelCollision<K & GraphChannelKeys<S, I, O>>]
    : { [P in K & GraphChannelKeys<S, I, O>]: NodeChannelCollision<P> };
```

```
probe.ts(4,40): error TS2769: No overload matches this call.
    Argument of type '(s: any) => { judge: any; }' is not assignable to parameter of type
    '"Node name \"judge\" is already a state channel name; pick a different node name"'.
```

Same result for the `options` form, the tuple-array form, and `addSequence`; legal chains still infer node names for `addEdge`. Two notes from trying it:

- The error tuple has to be fixed-length. I first tried `[error: Msg, ...rest: unknown[]]` and the `addNode("judge", fn, { ends })` form compiled without error, so that shape is unsafe.
- The object-map overload still reports the neighbouring tuple overload's message (`'judge' does not exist in type '[key: string, action: ...][]'`). That is overload ordering rather than this type, so I would leave it for a follow-up if at all.

Happy to see this land either way; the MVP as written already removes the module-evaluation crash that started the issue.

ryanzillini:
Addressed the review feedback on the type-error message.

**What changed** (`e660025`):
- `IfAllowedNodeKey` no longer rejects as `never`. Colliding calls now surface a branded template-literal diagnostic that names the channel, e.g. `Node name "judge" is already a state channel name; pick a different node name`.
- Error tuple is fixed-length (`[error, options?]`) so `addNode("judge", fn, { ends })` still type-errors (a rest `unknown[]` slot would have let that form compile).
- `key` stays a naked `K`, so `ends` / options still infer node names. No `ChannelNames` generic. Runtime guard unchanged.
- Left the object-map overload’s neighbouring-overload message alone, as suggested (follow-up if wanted).

**CI note:** Unit Tests on Node 26 were failing on an earlier run; after merging `main` into this branch they passed (along with 22/24, macOS, Windows). Latest commit will re-run CI.

Marking this ready for review. Thanks @jsiu93 for the precise suggestion.

jsiu93 (commented):
Re-reviewed at 3e84b31. Pulled the branch and ran a standalone type probe with the project's tsc 5.9.3, plus the PR's test file and graph.test.ts (29 passing).

- A colliding `addNode("judge", …)` now reports `Node name "judge" is already a state channel name; pick a different node name` instead of `never`, on addNode and on both addSequence forms.
- `addNode("judge", fn, { ends: [END] })` still errors, so the fixed-length tuple holds.
- Input-only / output-only schema keys (`incoming`, `outgoing`) are rejected on addNode and addSequence.
- Wide `string`, `Record<string, …>`, index-signature schemas, and a node colliding with its own `options.input` key all still compile.
- `ends` still infers node names: `addEdge("missing", END)` on the built graph errors as before.

Same conclusion as the first pass: the boundary matches the MVP from #2724, and the diagnostic is now actionable. LGTM from the issue author's side (GitHub does not let me submit a formal approval here); maintainer sign-off still needed for merge.

ryanzillini:
@hntrl This is ready for a maintainer look when you have a minute.

It turns the existing `addNode` channel-name collision into a compile-time error for constructor-time state, input, and output keys. CI is green. The issue author re-checked the current head at 3e84b31 and confirmed the boundary from #2724, including the diagnostic text, but cannot submit an approving review.

PR #2803, opened 2026-09-07
fix(checkpoint): stop search() prefix matching at namespace segments

### LINKED ISSUE #2721: `BaseStore.search()` namespace prefix matching crosses segment boundaries, returning sibling namespaces' data (JS port of CVE-2026-71433)
## Summary

`BaseStore` namespace prefix matching in the JS stores compares the joined namespace path as a **raw string prefix**, with no segment boundary. A read scoped to namespace `["tenant", "acme"]` also returns items stored under the unrelated sibling namespace `["tenant", "acme-corp"]`.

Since applications commonly use the namespace as a tenant/user boundary, an ordinary scoped read can return another tenant's data. No crafted input is required, and the caller gets no signal that the extra items came from elsewhere.

## This was already fixed in Python but not in JS

This is the same defect as **[CVE-2026-71433](https://advisories.gitlab.com/pypi/langgraph-checkpoint-postgres/CVE-2026-71433/)** (CVSS 5.3), *"Namespace prefix matching crosses segment boundaries in Postgres and SQLite stores"*, fixed in Python `langgraph-checkpoint-postgres` / `-sqlite` **3.1.1** via langchain-ai/langgraph#8478 (merged 2026-07-30).

The advisory's own wording describes the JS behaviour exactly:

> applications commonly use the namespace as a tenant boundary, where a read scoped to one namespace could return items belonging to another, without any crafted input.

The JS packages are **not listed as affected** in that advisory, and the fix was never ported. The two ecosystems now disagree about what a scoped read means:

| | Python (>= 3.1.1) | JavaScript (latest) |
|---|---|---|
| `search(prefix)` segment boundaries | Enforced | **Not enforced** |
| `LIKE` metacharacter handling | Escaped | Rejected at validation (see below) |
| SQLite case-sensitivity | `GLOB` | n/a |
| `*` wildcard spans | one segment | n/a |

Notably, the metacharacter half of the CVE *was* independently fixed in JS by #2512 ("reject SQL LIKE wildcards in PostgresStore namespace labels", released in `@langchain/langgraph-checkpoint-postgres@1.0.3`), about a month before the Python CVE fix. That PR modified the exact code path at issue here but addressed only the wildcard risk. Its rationale states:

> `InMemoryStore` uses literal `String.startsWith()` and is inherently safe

That is accurate with respect to wildcards, but `String.startsWith()` is precisely what allows `"tenant:acme-corp"` to match the prefix `"tenant:acme"`. So the boundary half of the defect remains open in both stores.

## Affected versions

Reproduced on the latest published releases:

- `@langchain/langgraph-checkpoint@1.1.5`
- `@langchain/langgraph-checkpoint-postgres@1.0.5`

## Affected code

**`InMemoryStore`** — `libs/checkpoint/src/store/memory.ts:158`

```js
for (const [namespace, items] of this.data.entries())
  if (namespace.startsWith(op.namespacePrefix.join(":")))
    candidates.push(...items.values());
```

**`PostgresStore`** — `libs/checkpoint-postgres/src/store/modules/search-operations.ts`, in `executeSearch` and `executeVectorSearch` (5 query sites; `dist` lines 32, 71, 117, 191, 247):

```sql
WHERE namespace_path LIKE $1     -- params: [`${namespacePath}%`]
```

**`PostgresStore.executeListNamespaces`** — `libs/checkpoint-postgres/src/store/index.ts`, both the `prefix` and `suffix` match conditions use bare `LIKE` with the joined path.

The equality path is correct and is what makes the inconsistency visible: `get` / `put` / `delete` use `namespace_path = $1` (exact), so an item can be returned by `search` and then be unreachable by `get` at the same namespace.

**Secondary bug:** `InMemoryStore.listNamespaces` uses the segment-wise `doesMatch` helper and is *not* affected, while `PostgresStore.listNamespaces` uses raw `LIKE` and *is*. The two stores therefore return different results for the same call, independent of the security concern.

## Reproduction

No database required.

```js
import { InMemoryStore } from "@langchain/langgraph-checkpoint";

const store = new InMemoryStore();

// Two unrelated tenants. "acme" is NOT a parent of "acme-corp";
// they are siblings whose names happen to share leading characters.
await store.put(["tenant", "acme"], "note", { text: "acme's own note" });
await store.put(["tenant", "acme-corp"], "secret", { text: "acme-corp CONFIDENTIAL" });

const scoped = await store.search(["tenant", "acme"]);
for (const item of scoped) {
  console.log(item.namespace, item.key, item.value);
}

// get() is correctly exact, which is what makes this inconsistent:
console.log(await store.get(["tenant", "acme"], "secret"));
```

Actual output on `1.1.5`:

```
search(['tenant','acme']) returned:
  namespace=["tenant","acme"]      key=note   value={"text":"acme's own note"}
  namespace=["tenant","acme-corp"] key=secret value={"text":"acme-corp CONFIDENTIAL"}

LEAK: 1 item(s) from another namespace
get(['tenant','acme'],'secret') => null (correct)
```

Expected: `search(["tenant","acme"])` returns only the `["tenant","acme"]` item.

## Suggested fix

Mirror langchain-ai/langgraph#8478: require either an exact namespace match, or the separator immediately after the prefix.

`InMemoryStore`:

```js
const prefix = op.namespacePrefix.join(":");
if (namespace === prefix || namespace.startsWith(`${prefix}:`)) { ... }
```

`PostgresStore` (keeps both branches indexable):

```sql
WHERE (namespace_path = $1 OR namespace_path LIKE $2)   -- $1 = prefix, $2 = prefix || ':%'
```

The same change applies to `executeListNamespaces`, and the `suffix` branch needs the mirrored treatment (exact, or `'%:' || suffix`).

Since #2512 already rejects `%`, `_`, and `\` in namespace labels for `PostgresStore`, no additional escaping is needed there. `InMemoryStore` has no such validation, but `startsWith` is literal, so it is unaffected by metacharacters.

Happy to open a PR if that would help.

## Impact for us

We hit this in production on a multi-tenant deployment: a user in team `growth` received `grep` results containing the **file contents** of team `growth-marketing`, which they are not a member of. We are shipping a local wrapper that filters `search` results to the exact namespace, but the defect is in the store contract, so every JS consumer using namespaces as a tenant boundary is exposed by default.

devtechedge:
Closes #2721

Port of CVE-2026-71433 / langchain-ai/langgraph#8478 to the JS stores.

## Bug

`InMemoryStore.search()` compared the colon-joined namespace with a raw `startsWith`. `PostgresStore` used `namespace_path LIKE prefix%`. A read scoped to `["tenant","acme"]` also returned `["tenant","acme-corp"]`. `get` is exact, so the extra item is visible on search and then unreachable.

## Why the issue-body matcher is not enough

`namespace === prefix || namespace.startsWith(prefix + ":")` closes the sibling leak and turns `search([])` into `[]` on InMemoryStore, because `[].join(":")` is `""`. Confirmed on 1.1.5: 2 items → 0, silent. Empty prefix is documented as "search everything".

`:` was also legal in a JS label (`validateNamespace` only rejected `.`, the Python separator). `put(["a","b:c"])` then `put(["a","b","c"])` collide on the joined key.

## Fix

- InMemory: match exact, or require `:` after the prefix, or empty prefix (unconstrained)
- Postgres search (5 query sites): `namespace_path = $1 OR namespace_path LIKE $2` with `$2 = path || ':%'`
- Postgres `listNamespaces`: same for prefix/suffix; `*` uses a POSIX regex so it spans one segment
- Labels may not contain `:` on put or search (write and read)

Postgres `search([])` still throws `Namespace cannot be empty` — existing, not changed.

## Tests

```
pnpm --filter @langchain/langgraph-checkpoint test
pnpm --filter @langchain/langgraph-checkpoint-postgres test src/store/modules/utils.test.ts
```

113 + 20 passed. Includes the issue repro, empty-prefix, and a segment-wise property test that never touches the joined string.

## Agent disclosure

Written by an AI agent on behalf of @devtechedge, who has reviewed this pull request.

### DIFF
--- a/.changeset/namespace-segment-boundary.md
+++ b/.changeset/namespace-segment-boundary.md
@@ -1,0 +1,8 @@
+---
+"@langchain/langgraph-checkpoint": patch
+"@langchain/langgraph-checkpoint-postgres": patch
+---
+
+fix(checkpoint): stop `search()` namespace prefix matching at segment boundaries (CVE-2026-71433 port)
+
+`InMemoryStore.search(["tenant","acme"])` no longer returns sibling `["tenant","acme-corp"]` items. Postgres `LIKE prefix%` is now exact-or-descendant (`= path OR LIKE path:%`). Namespace labels may not contain `:`. Empty prefix still means "search everything" on InMemoryStore. Closes #2721.
--- a/libs/checkpoint-postgres/src/store/index.ts
+++ b/libs/checkpoint-postgres/src/store/index.ts
@@ -28,6 +28,11 @@ import {
   StoreMigrationConfig,
 } from "./store-migrations.js";
 import { getStoreTablesWithSchema } from "./sql.js";
+import {
+  namespaceMatchRegex,
+  namespaceScopeClause,
+  namespaceScopeParams,
+} from "./modules/utils.js";
 
 export type * from "./modules/types.js";
 
@@ -326,19 +331,36 @@ export class PostgresStore extends BaseStore {
     const conditions: string[] = [];
     let paramIndex = 1;
 
-    // Add match conditions
+    // Add match conditions. Empty path constrains nothing. Paths without `*`
+    // stay index-friendly (`=` / `LIKE path:%`). `*` spans one segment and
+    // needs a POSIX regex because LIKE cannot exclude the `:` separator.
     if (matchConditions && matchConditions.length > 0) {
       for (const condition of matchConditions) {
-        if (condition.matchType === "prefix") {
-          const prefix = condition.path.join(":");
-          conditions.push(`namespace_path LIKE $${paramIndex}`);
-          params.push(`${prefix}%`);
+        if (
+          condition.matchType !== "prefix" &&
+          condition.matchType !== "suffix"
+        ) {
+          continue;
+        }
+        if (!condition.path.length) {
+          continue;
+        }
+        if (condition.path.includes("*")) {
+          conditions.push(`namespace_path ~ $${paramIndex}`);
+          params.push(namespaceMatchRegex(condition.path, condition.matchType));
           paramIndex += 1;
-        } else if (condition.matchType === "suffix") {
-          const suffix = condition.path.join(":");
-          conditions.push(`namespace_path LIKE $${paramIndex}`);
-          params.push(`%${suffix}`);
-          paramIndex += 1;
+        } else {
+          const { clause, nextIndex } = namespaceScopeClause(
+            "namespace_path",
+            paramIndex
+          );
+          conditions.push(clause);
+          const { exact, like } = namespaceScopeParams(
+            condition.path,
+            condition.matchType
+          );
+          params.push(exact, like);
+          paramIndex = nextIndex;
         }
       }
     }
--- a/libs/checkpoint-postgres/src/store/modules/search-operations.ts
+++ b/libs/checkpoint-postgres/src/store/modules/search-operations.ts
@@ -7,7 +7,11 @@ import { DatabaseCore } from "./database-core.js";
 import { VectorOperations } from "./vector-operations.js";
 import { QueryBuilder } from "./query-builder.js";
 import { SearchOptions, SearchItem } from "./types.js";
-import { validateNamespace } from "./utils.js";
+import {
+  validateNamespace,
+  namespaceScopeClause,
+  namespaceScopeParams,
+} from "./utils.js";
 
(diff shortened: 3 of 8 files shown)


hntrl (commented):
Closing in favor of #2825- we agree with the fix but adding the `:` constraint to postgres feels like an over-optimization.

Some of our design was inspired by what you had here, so we'll make sure you get properly attributed!

PR #2806, opened 2026-09-09
fix(checkpoint): handle prototype-named texts in vector indexing

To3akaRin:
When an indexed field contains a text such as `constructor`, `toString`, or `__proto__`, `InMemoryStore.put()` throws `TypeError: toEmbed[texts[0]].push is not a function` before storing the item. The text-to-embedding map is a plain object, so inherited properties bypass its array initialization.

Use a null-prototype object for this map. Regression tests cover single-value indexing, wildcard extraction, and duplicate texts within and across documents in one batch. They verify readback, one embedding per unique text, and vector search scores for every document. Includes a checkpoint patch changeset.

Validation:
- Before the fix: all 5 new regression cases fail; the 7 existing vector tests pass.
- After the fix: all 115 checkpoint tests pass, using local embeddings with no external API calls.
- Checkpoint build (including declaration generation, attw and publint), repository lint, format check, and `git diff --check` pass.

### DIFF
--- a/.changeset/safe-vector-text-keys.md
+++ b/.changeset/safe-vector-text-keys.md
@@ -1,0 +1,5 @@
+---
+"@langchain/langgraph-checkpoint": patch
+---
+
+Fix InMemoryStore vector indexing failing for texts such as `constructor`, `toString`, and `__proto__`. Preserve embedding deduplication across fields and documents.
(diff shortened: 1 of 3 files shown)


To3akaRin:
Hi maintainers, just a gentle follow-up on this small fix. All CI checks are passing, and the regression tests cover both prototype-named texts and embedding deduplication across fields and documents. When you have a chance, I'd appreciate a review. Thank you!

PR #2821, opened 2026-09-10
fix(sdk): restart root pump after stream failure

### LINKED ISSUE #2817: useStream: dead SSE pump leaves isLoading=true, error unset, and cannot reconnect
### Summary

When the SSE transport behind `StreamController` exhausts `maxReconnectAttempts`, the root pump exits silently. `isLoading` stays `true` (if a run was in flight), `error` stays `undefined`, and the pump can never be reopened on that controller. From the consumer's point of view the stream looks busy forever rather than broken.

Versions: `@langchain/langgraph-sdk` 1.10.0, `@langchain/react` 1.0.33.

### Chain of events

1. `ProtocolSseTransportAdapter.openEventStream` (`client/stream/transport/http.ts`) retries a broken fetch up to `maxReconnectAttempts` (default 5, ~20s of backoff). After the last failure it calls `streamQueue.close(toError(error))`, so the consuming iterator throws.
2. `ThreadStream.#pumpStream` (`client/stream/index.ts`) catches and calls `#failThreadWithError`, which rejects pending commands and calls `subscription.close()` on every handle. It does not mark the thread closed and fires no listener.
3. `StreamController.#startRootPump` (`stream/controller.ts`): the `for await (const event of subscription)` loop sees `done: true`, `subscription.isPaused` is `false`, so the outer `while` breaks and the pump promise resolves. Nothing threw at this layer, so the `catch` branch never runs.
4. `isLoading` is only flipped by `LifecycleLoadingTracker` on a root terminal lifecycle event, and `error` is only set by `hydrate()` or the submit coordinator. A dead transport hits neither path, so the root snapshot is never updated.
5. `#startRootPump` begins with `if (this.#rootPump != null) return`. The resolved promise is still assigned, so a later `submit()` on the same controller cannot reopen the pump. The only exits are a thread swap or `dispose()`.

### Reproduction

1. `useStream({ threadId, assistantId })` against a thread with a run in progress.
2. Kill network connectivity (or the server) for longer than the reconnect budget (~20s with defaults).
3. Restore connectivity.

Observed: `stream.isLoading === true`, `stream.error === undefined`, no further events ever arrive; `stream.submit()` does not reopen the stream.

Expected: `isLoading` drops to `false`, `error` carries the transport error, and either `submit()` or an explicit reconnect can bring the stream back.

### Suggested fix

- When `#failThreadWithError` fires (or the root pump loop ends while not paused and not disposed), set `rootStore.error` to the transport error and `isLoading` to `false`.
- Clear `#rootPump` so a later `submit()` can reopen it.
- Expose a `reconnect()` on `StreamController` and on the `useStream` return so long-lived consumers can recover without remounting the hook.

### Context

We keep several `useStream` instances mounted in a pool so switching threads is instant, and retain any instance whose `isLoading` is `true`. A dead stream is therefore pinned indefinitely precisely because it looks busy; the only recovery today is remounting the hook with a new key.

kkkhs:
## Summary

Fixes #2817.

When a ThreadStream reports a fatal stream error, StreamController now closes and clears its root pump state so the next submit opens a fresh root subscription instead of reusing the settled pump promise.

The root pump also reports an unexpected unpaused subscription close during an active run before clearing its pump state.

## Tests

- `pnpm --filter @langchain/langgraph-sdk test src/stream/controller.test.ts`
- `pnpm exec oxlint libs/sdk/src/stream/controller.ts libs/sdk/src/stream/controller.test.ts`
- `pnpm exec oxfmt --check libs/sdk/src/stream/controller.ts libs/sdk/src/stream/controller.test.ts .changeset/root-pump-restart.md`
- `pnpm --filter @langchain/langgraph-sdk build`

### DIFF
--- a/.changeset/root-pump-restart.md
+++ b/.changeset/root-pump-restart.md
@@ -1,0 +1,6 @@
+---
+"@langchain/langgraph-sdk": patch
+---
+
+Restart the root stream pump after a thread stream failure so a later `submit()`
+can receive terminal lifecycle events instead of leaving `isLoading` stuck.
--- a/libs/sdk/src/stream/controller.test.ts
+++ b/libs/sdk/src/stream/controller.test.ts
@@ -3505,6 +3505,68 @@ describe("StreamController", () => {
     await controller.dispose();
   });
 
+  it("restarts the root pump after a thread stream failure", async () => {
+    const errorListeners = new Set<(error: Error) => void>();
+    const subscriptions: ReturnType<typeof makePushableSubscription>[] = [];
+    let run = 0;
+    const submitRun = vi.fn(async () => ({ run_id: `run-${++run}` }));
+    const thread = {
+      subscribe: vi.fn(async () => {
+        const subscription = makePushableSubscription();
+        subscriptions.push(subscription);
+        return subscription;
+      }),
+      onError: vi.fn((listener: (error: Error) => void) => {
+        errorListeners.add(listener);
+        return vi.fn(() => {
+          errorListeners.delete(listener);
+        });
+      }),
+      onEvent: vi.fn(() => vi.fn()),
+      close: vi.fn(async () => undefined),
+      interrupts: [],
+      ordering: {},
+      submitRun,
+      startLifecycleWatcher: vi.fn(() => undefined),
+    } as unknown as ThreadStream;
+    const client = {
+      threads: {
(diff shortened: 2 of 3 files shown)


hntrl (commented):
Going to close this in favor of #2831 since we have some additional connection lifecycle things we want to address. Thanks for the PR! We'll make sure to attribute you as a co author

PR #2824, opened 2026-09-10
feat(langgraph): add responseSchema option to interrupt()

eliornl:
Adds `interrupt(value, { responseSchema })`. The schema (Zod or JSON Schema) is surfaced on `Interrupt.response_schema` as JSON Schema so clients (LangGraph Studio first) can render a typed form for the resume value instead of a free-form JSON box. Omit it and nothing changes.

A Zod schema also parses the resume value, and the parsed value is what `interrupt()` returns; a parse failure rejects before the value is committed, so the next resume works. A JSON Schema object is passed through to clients without validating the resume value. `@langchain/langgraph-sdk`'s `Interrupt` type gains the optional `response_schema` field.

Python counterpart: langchain-ai/langgraph#8886.

### DIFF
--- a/libs/langgraph-core/src/constants.ts
+++ b/libs/langgraph-core/src/constants.ts
@@ -1,4 +1,5 @@
 import { PendingWrite } from "@langchain/langgraph-checkpoint";
+import type { JSONSchema } from "@langchain/core/utils/json_schema";
 import {
   coerceTimeoutPolicy,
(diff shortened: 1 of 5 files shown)


hntrl (commented):
can we add a `minor` changeset?

hntrl (dismissed):

hntrl (approved):

PR #2825, opened 2026-09-10
fix(langgraph-checkpoint-postgres): anchor namespace matching to segments

byhow:
## Summary
Fix cross-namespace reads in the Postgres store: searching `["tenant", "a"]` previously also returned `["tenant", "ab"]`. All five search paths now match the exact namespace or its descendants at a `:` boundary. Namespace listings likewise respect prefix and suffix boundaries, so suffix `["alice"]` excludes `users:malice`.

Listing filters now validate labels and support `*` as exactly one segment, matching `InMemoryStore`. Embedded stars and search prefixes remain literal. LIKE patterns are escaped, and `:` is rejected inside labels to prevent ambiguous stored paths. Empty listing filters remain unrestricted; the reserved-root rule applies to prefixes, not suffixes. Includes regression tests, documentation, and a patch changeset.

The matching rules follow the Python counterpart PR https://github.com/langchain-ai/langgraph/pull/8478, using `:` instead of `.`. Wildcard listings use segment-aware regex; literal listings retain equivalent equality/LIKE matching. JavaScript's existing `%`, `_`, and backslash restrictions remain. The related [CVE-2026-71433 advisory](https://github.com/langchain-ai/langgraph/security/advisories/GHSA-47pj-3jcm-6whg) covers Python packages; it does not list this npm package.

Validation after merging `origin/main`:

- The initial 8 security regressions all fail against original implementation `039c5441`, reproducing the leaks.
- TypeScript, scoped lint, formatting, and whitespace checks pass.
- Local verification reused installed dependencies and a temporary source alias for the unbuilt checkpoint package; those temporary configs are not included.

Co-authored-by: Dev M <294291171+devtechedge@users.noreply.github.com>

PR #2828, opened 2026-09-11
fix(checkpoint): enforce InMemory namespace segment boundaries

byhow:
Fix InMemoryStore searches returning sibling namespaces whose labels share a character prefix. Searches now include only the exact namespace and descendants; empty prefixes remain unrestricted. Reject colon aliases through public search/write and direct batch operations.

The original PR remains open; this PR does not replace or close it automatically. Related report: #2721.

Validation:
- All 89 checkpoint unit tests pass.
- Four namespace regressions fail against original implementation 039c5441; the corresponding fixed tests pass.
- Bundled InMemory isolation checks pass inside Node 24 Docker.
- Scoped lint passes. Package typecheck reports six existing argument-count errors in memory-pollution.test.ts; the same errors reproduce on baseline 039c5441. Source typechecking passes.

The colon restriction is local to InMemoryStore, preserving MongoDB colon-label compatibility. This draft is ready for code review; merge still requires normal maintainer approval and CI. No anti-slop tooling is included.

Stack: this PR is based on #2825 (Postgres). It contains only the checkpoint/InMemory portion adapted from @devtechedge's #2803

Co-authored-by: Dev M [294291171+devtechedge@users.noreply.github.com](mailto:294291171+devtechedge@users.noreply.github.com)

### DIFF
--- a/libs/checkpoint/src/store/memory.ts
+++ b/libs/checkpoint/src/store/memory.ts
@@ -10,6 +10,7 @@ import {
   GetOperation,
   type IndexConfig,
   type SearchItem,
+  validateNamespace,
 } from "./base.js";
 import { tokenizePath, compareValues, getTextAtPath } from "./utils.js";
 
@@ -79,6 +80,11 @@ export class InMemoryStore extends BaseStore {
     // First pass - handle gets and prepare search/put operations
     for (let i = 0; i < operations.length; i += 1) {
       const op = operations[i];
+      if ("namespace" in op) {
+        validateNamespace(op.namespace, {
+          allowReservedRoot: true,
+        });
+      }
       if ("key" in op && "namespace" in op && !("value" in op)) {
         // GetOperation
         results.push(this.getOperation(op));
@@ -249,9 +255,22 @@ export class InMemoryStore extends BaseStore {
   }
 
   private filterItems(op: SearchOperation): Item[] {
+    if (op.namespacePrefix.length > 0) {
+      validateNamespace(op.namespacePrefix, { allowReservedRoot: true });
+    }
+    const prefix = op.namespacePrefix.join(":");
     const candidates: Item[] = [];
     for (const [namespace, items] of this.data.entries()) {
-      if (namespace.startsWith(op.namespacePrefix.join(":"))) {
+      // Exact match, or the separator immediately after the prefix, so
+      // "tenant:acme" does not also match sibling "tenant:acme-corp".
+      // Empty prefix is unconstrained (search everything). See #2721 /
+      // CVE-2026-71433. Do not use startsWith(prefix) or startsWith(prefix+":")
+      // without the empty-prefix arm — the latter turns search([]) into [].
+      if (
+        prefix === "" ||
+        namespace === prefix ||
+        namespace.startsWith(`${prefix}:`)
+      ) {
         candidates.push(...items.values());
       }
     }
--- a/libs/checkpoint/src/store/memory.ts
+++ b/libs/checkpoint/src/store/memory.ts
@@ -17,6 +17,7 @@ import { tokenizePath, compareValues, getTextAtPath } from "./utils.js";
 
 function validateMemoryNamespace(namespace: string[]): void {
   validateNamespace(namespace, { allowReservedRoot: true });
+
   if (namespace.some((label) => label.includes(":"))) {
     throw new InvalidNamespaceError(
       "Namespace labels cannot contain colons (':'), the InMemoryStore path separator."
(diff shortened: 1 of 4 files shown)


devtechedge:
Thanks for picking this up and for the co-author credit. Happy to help close out the two open items:

1. **BaseStore validation compatibility** - I can review the shared validation approach; the design in #2803 kept `search([])` unrestricted (documented "search everything" behavior), so if that's preserved here it should be compatible.
2. **The six `memory-pollution.test.ts` typecheck errors** — this PR doesn't touch that file, so I believe the argument-count errors pre-exist on `main`. I can verify and submit a small follow-up PR to fix them so they don't block this.

Also happy to rebase or close #2803 in favor of this, whichever makes review easiest - just let me know which path you'd prefer.

libs/checkpoint/src/store/memory.ts
```diff
@@ -10,9 +10,21 @@ import {
...
 import { tokenizePath, compareValues, getTextAtPath } from "./utils.js";
 
+function validateMemoryNamespace(namespace: string[]): void {
+  validateNamespace(namespace, { allowReservedRoot: true });
+
+  if (namespace.some((label) => label.includes(":"))) {
+    throw new InvalidNamespaceError(
+      "Namespace labels cannot contain colons (':'), the InMemoryStore path separator."
+    );
+  }
```
hntrl, line 25: why do we need to add this invariant?

hntrl (approved):

libs/checkpoint/src/store/memory.ts
hntrl: i now understand why

libs/checkpoint/src/tests/namespace.test.ts
```diff
@@ -192,6 +192,86 @@ describe("InMemoryStore Namespace Operations", () => {
...
+    const doc = { foo: "bar" };
+    await expect(store.put(["tenant", "acme:corp"], "k", doc)).rejects.toThrow(
+      InvalidNamespaceError
+    );
+    await expect(store.search(["tenant:acme"])).rejects.toThrow(
+      InvalidNamespaceError
+    );
+  });
+
+  it("should not leak sibling namespaces that share a string prefix (issue 2721 / CVE-2026-71433)", async () => {
```
hntrl: can we not have github meta in tests?

byhow: good point, just removed

libs/checkpoint/src/tests/namespace.test.ts
```diff
@@ -233,3 +313,20 @@ describe("InMemoryStore Namespace Operations", () => {
     expect(batchDeletedResult).toBeNull();
   });
 });
+
+
+it("rejects namespace aliases through direct batch reads, writes and deletes", async () => {
```
devtechedge: Small nit while we are in this file: the last case
(`rejects namespace aliases through direct batch reads, writes and deletes`) is
declared at file scope, after the closing `});` of the describe block, and it
builds its own `new InMemoryStore()` instead of using the shared fixture. Moving
it inside the describe and dropping the local store makes it consistent with the
other cases.

```ts
  it("rejects namespace aliases through direct batch reads, writes and deletes", async () => {
    await store.put(["tenant", "a", "notes"], "key", { own: true });

    for (const operation of [
      { namespace: ["tenant", "a:notes"], key: "key" },
      { namespace: ["tenant", "a:notes"], key: "key", value: { own: false } },
      { namespace: ["tenant", "a:notes"], key: "key", value: null },
      { namespacePrefix: ["tenant", "a:notes"] },
    ]) {
      await expect(store.batch([operation])).rejects.toThrow(/colons/);
    }

    expect((await store.get(["tenant", "a", "notes"], "key"))?.value).toEqual({ own: true });
  });
```

Not blocking and assertions are unchanged either way.

hntrl (approved):

libs/checkpoint/src/tests/namespace.test.ts
byhow: good nit 👍 addressed this in the latest push

devtechedge:
One small thing I want to confirm. I had offered earlier to rebase or close #2803 in favor of this, whichever made review easiest, and #2803 was closed a few days later.

If that was just taking me up on the offer, nothing to do and I am glad to leave it. I only want to be sure it was not closed for feedback I should learn from, since no comment came with it.

hntrl:
Hey @devtechedge! See my comment here: https://github.com/langchain-ai/langgraphjs/pull/2803#pullrequestreview-5202814764

I ended up walking back the assertion about adding `:` being an optimization, but its better for us to split up these concerns into separate PRs (see https://github.com/langchain-ai/langgraphjs/pull/2828)

devtechedge:
Thanks for the follow-up, @hntrl, and for walking that back.

Splitting the Postgres and InMemory changes into separate PRs makes sense given they ship as separate packages, and that answers my question about #2803.

Glad the InMemory fix is landing in #2828. If anything there needs another pass, I am happy to help.

PR #2829, opened 2026-09-11
fix(langgraph-checkpoint-redis): anchor namespace matching to segments

byhow:
`RedisStore` matched namespaces with RediSearch `TEXT` queries and treated the result as the answer. A `TEXT` query is an unordered, case-folded AND over tokens cut at punctuation, so a scoped read, overwrite or delete could resolve a document in a different namespace.

`["tenant", "a"]` collided with `["tenant", "A"]`, `["a", "tenant"]` and `["tenant", "a-b"]`, and `namespace.join(".")` additionally made `["a.b"]` and `["a", "b"]` indistinguishable. Because `get()` and `put()` took the first hit without checking it, `put()` could `DEL` a document in another namespace — a silent cross-tenant delete requiring no read access. Labels are user-derived and were interpolated into the query unescaped, so `["tenant", "a) | @prefix:(victim"]` could rewrite it.

## The fix

The indexed query is demoted to a candidate filter, and the exact namespace is confirmed in application code on every document before it is returned, overwritten or deleted. `search` applies that check before pagination and vector selection, so a namespace cannot leak in through a page boundary or a nearest-neighbour result.

Candidate terms are restricted to runs of `[a-zA-Z0-9_]` and non-ASCII characters, which also closes the injection vector structurally: no query syntax survives the filter. Narrowing is an optimisation only — every omission widens the candidate set and changes no result.

The whole guarantee reduces to one predicate, `isWithinNamespace`, which anchors matching at segment boundaries:

```ts
prefix === "" || found === prefix || found.startsWith(`${prefix}.`)
```

## No migration

`SCHEMAS` is byte-identical to the merge base, there is no `FT.ALTER` or `dropIndex`, and document format, key naming and required Redis permissions are unchanged. Old and new clients coexist on the same index, so upgrading the package is the complete fix. That is why this is a `patch` alongside #2825 and #2828. See [Namespace matching](https://github.com/langchain-ai/langgraphjs/blob/codex/redis-namespace-boundaries/libs/checkpoint-redis/README.md#namespace-matching).

## Failures surface instead of being swallowed

Three deliberate behaviour changes beyond namespace matching, all in the same direction.

`setup()` previously reported a failed `FT.CREATE` through `console.error` and continued, so a store whose index was never created looked healthy until its first query came back empty. It now probes the index after a failed create and throws only when the index is genuinely unusable, preserving the original error as `cause`. A concurrent creation, or an ACL that grants reads but not creates, is still tolerated — it no longer depends on matching the server's error text.

`get()` and `search()` previously turned a missing index into `null` and `[]`. They now throw and name `await store.setup()`. An empty result and an absent index are different answers, and returning the first for the second hides the misconfiguration instead of reporting it.

`put()` previously swallowed a failed lookup of the document it was about to replace and wrote anyway, which could leave the old document in place alongside the new one. A failed lookup now aborts the mutation it was guarding.

## Reviewability

The security-critical logic is arranged so a reviewer has to check few things, and so a mistake fails fast rather than in production.

`isWithinNamespace` is a pure module-level function rather than a closure inside the search path, so the boundary rule is unit-tested directly — a shared string prefix, reordered labels, a delimiter alias, case, a parent against its own child, and the empty prefix. Reverting the segment anchor fails 6 unit tests; removing `findDocument`'s identity check fails 4, including the cross-tenant overwrite. Both previously needed a container to detect.

`findDocument` and the non-vector search path had near-identical paging loops with separately written termination conditions. Both now consume one `candidatePages` generator, so when paging stops is decided in a single place and each caller only says what it wants from a page. The page size is a named constant, and its type excludes `LIMIT` so a caller cannot fight the pager.

`FilterBuilder.buildRedisSearchQuery` is removed. It built queries with the unanchored `split(/[.-]/)` this PR removes everywhere else and interpolated filter input without the character whitelist the live path uses. Nothing called it outside its own test.

One test-runner fix rides along. Vitest 4 removed `test.poolOptions` and moved its contents to top level, which it announces on every integration run, so this package's `poolOptions.forks.singleFork` has been dead since that upgrade — the single fork its comment asks for, to control cleanup of the Docker container each test starts, was not in effect. It is now the top-level `singleFork`. No other package in the repo still nests pool options.

## Validation

65 unit and 128 integration tests pass against `redis:8` and `redis/redis-stack-server:latest` via testcontainers with deterministic embeddings; typecheck, lint and format are clean.

The suite is checked against the vulnerability rather than only against the fix: mutating either half of the invariant — the segment anchor, or the exact-identity check — fails unit tests in milliseconds, and substituting the pre-stack `store.ts` fails 12 integration tests covering exact reads/updates/deletes with identical keys, delimiter aliases in both query modes, case-sensitive keys, segment isolation with vector on and off, TTL refresh across namespaces, and search pagination.

One TTL behaviour the write path never asserted is now pinned: updating a document re-applies the full configured TTL rather than inheriting what remained of the previous document's. Nothing reads a TTL, so that holds only because the replacement is written with an expiry of its own.

## Performance

`benchmarks/` is included so these can be reproduced. p50 ms over 100k documents, `redis:8`:

| | before stack | this PR |
|---|---|---|
| `get` hit | 0.28 | 0.27 |
| `get` miss, hot key | 0.25 | 0.24 |
| `search` dense | 28.50 | 30.07 |
| `search` sparse | 0.41 | 0.72 |
| `search` fallback | 0.39 | 0.70 |
| `search` + KNN, sparse | 2.68 | 2.35 |

The KNN row is faster *and* correct: at 100k the benchmark's own result check shows the pre-stack build answering a single-namespace query with `[tenant-dense.notes | tenant-10355.notes | tenant-26675.notes]`, where this PR returns only the requested namespace. The sparse and fallback rows are the cost of verification, around a third of a millisecond. Removing narrowing altogether costs roughly 80s per search at this size, which is why the candidate query is kept rather than replaced by a full scan. `benchmarks/tokenizer-probe.ts` records the RediSearch tokenizer behaviour the query rule depends on, measured rather than assumed.

The reviewability refactor was benchmarked separately against the commit before it and moved no cell by more than 0.04ms.

## Size

1,607 changed lines, above the usual 1,000 guideline. 774 are tests and 306 are the benchmark and tokenizer harness, leaving 521 lines of source change and 6 of test-runner config. The isolation suite is the evidence that the fix works and cannot sensibly be split from it.

Stack: #2825 → #2828 → this PR. Adapted from @jackjin1997's #2826; related report #2721. Literal listing semantics are unchanged.


libs/checkpoint-redis/NAMESPACE_UPGRADE.md
hntrl, line 1: ?

libs/checkpoint-redis/src/store.ts
```diff
@@ -432,18 +432,36 @@ export class RedisStore {
...
+      : ["store"]) {
+      try {
+        await this.client.ft.alter(index, {
+          "$.prefix": {
+            type: SchemaFieldTypes.TAG,
+            AS: "namespace",
+            CASESENSITIVE: true,
+          },
+        });
+      } catch (error: any) {
```
hntrl: i'm guessing this is an upsert op? what if these tags dont exist?

libs/checkpoint-redis/src/store.ts
```diff
@@ -822,7 +831,7 @@ export class RedisStore {
 
           let matches = true;
           for (let i = 0; i < options.prefix.length; i++) {
-            if (parts[i] !== options.prefix[i]) {
+            if (options.prefix[i] !== "*" && parts[i] !== options.prefix[i]) {
```
hntrl: is this a behavior change

hntrl (commented):
half rubber stamp, am not nearly familiar enough with redis data types? If we feel good that integration tests (against a newer version of redis) are covering old and new behavior then I feel fine about stamping?

some comments though

libs/checkpoint-redis/NAMESPACE_UPGRADE.md
byhow: moved to readme or i am ok with dropping this altogether

libs/checkpoint-redis/src/store.ts
byhow: Yes [on the index](https://redis.io/docs/latest/commands/ft.alter/). should be fine since we are creating before altering, and if index doesn't exist it just throws and we catch it in setup()

libs/checkpoint-redis/src/store.ts
byhow: yeah this is extraneous; this was trying to expand the wildcard matching for multiple namespaces in between

libs/checkpoint-redis/README.md
```diff
@@ -156,6 +156,31 @@ const ops = [
...
+path, so descendant searches use one exact tag without Redis wildcard-expansion
+limits. Document values and TTLs are preserved. Setup scans existing documents
+in batches and is safe to retry after interruption; it waits for indexing and
+throws if an index is still building after 60 seconds. Wait and retry setup in
+that case. Larger stores need a maintenance window for the scan and added index
+storage. Setup requires permission for FT.ALTER, FT.INFO, FT.AGGREGATE/cursors,
+EVAL, JSON.GET and JSON.SET. Run upgraded clients only: older clients issue
+unsafe TEXT queries and
+write documents without the new prefix field.
+
```
hntrl: am a little bit nervous that we're adding a change that requires `.setup()`. If we're using old configs are people going to implicitly break on the next upgrade?

libs/checkpoint-redis/src/store.ts
hntrl: do we need to rm then

libs/checkpoint-redis/src/store.ts
```diff
@@ -428,22 +429,112 @@ export class RedisStore {
...
+    while (
+      await this.client.eval(
+        `local info = redis.call('FT.INFO', ARGV[1])
+       for i = 1, #info, 2 do
+         if (type(info[i]) == 'table' and info[i].ok or info[i]) == 'indexing' then return tonumber(info[i + 1]) end
+       end
+       return redis.error_reply('Missing index status')`,
+        { arguments: [index] }
+      )
+    ) {
```
hntrl: dont love this

jackjin1997:
Came at `store.ts` from a different angle and landed on the same KNN block, so one data point that may help size this PR.

The pagination defect the rewrite fixes is reproducible on `main` on its own, independently of namespace matching. On `main` the vector path asks for `[KNN ${limit}]` and then pages the result with `LIMIT { from: offset, size: limit }`, so `offset >= limit` pages into a candidate set that is already exhausted. Seeding 7 matching items and walking them with `limit: 3` gives page sizes `[3, 0, 0]` instead of `[3, 3, 1]` — everything past the first page is silently empty, which is the failure mode where a caller stops paging and quietly misses stored memories. `const wanted = offset + limit` plus the widen-and-retry loop here fixes exactly that, and the added `SORTBY __embedding_score ASC` also settles the ordering, which [the Redis docs](https://redis.io/docs/latest/develop/ai/search-and-query/vectors/) call out as required for deterministic paging.

Raising it because none of that is visible from the title, and it is worth keeping if this PR ever gets asked to narrow to namespace anchoring. I have a small container-free unit test that pins the page-size behaviour by faking `ft.search` with the documented KNN/LIMIT semantics — happy to hand it over if it is useful, or to open it separately against whatever lands.

For reference, the sibling backends already behave this way: MongoDB requests `limit + offset` before `$skip`/`$limit`, Postgres groups before `LIMIT ... OFFSET`, and `InMemoryStore` skips with `ix < offset`. Redis was the only one truncating first.

libs/checkpoint-redis/src/store.ts
```diff
@@ -339,6 +295,91 @@ const SCHEMAS = [
   },
 ];
 
+/**
+ * RediSearch's default stopword list. These are never indexed, so using one as
+ * a query term would exclude every document instead of narrowing the search.
+ */
+const REDISEARCH_STOPWORDS = new Set([
```
byhow: what if the stop words are updated? that is not gonna work if stop words are different from different redis instances either.

libs/checkpoint-redis/src/store.ts
```diff
@@ -339,6 +295,91 @@ const SCHEMAS = [
...
+ * single term: control characters and the backslash. A term read off a label
+ * holding one was never indexed under that spelling.
+ */
+const FUSES_TERMS = /[\u0000-\u001f\u007f\\]/;
+
+/**
+ * Term boundaries: ASCII punctuation and space. Letters, digits, "_" and every
+ * non-ASCII character are term content, so "café" is a term of its own.
+ */
+const NOT_TERM_CHARACTER = /[^a-zA-Z0-9_\u0080-\uffff]+/;
```
byhow: one off manual prune is also not ideal

libs/checkpoint-redis/src/store.ts
```diff
@@ -421,76 +486,156 @@ export class RedisStore {
...
-        LIMIT: { from: 0, size: key === "" ? 100 : 1 },
+  /**
+   * Candidate rows for `query`, a page at a time.
+   *
+   * Callers confirm the exact namespace themselves, so a page may hold nothing
+   * they want and the row they are after can sit on any later page. Paging
+   * runs until RediSearch reports the candidate set exhausted; stopping
+   * earlier is the caller's business.
+   */
+  private async *candidatePages(
```
byhow: why async generators

libs/checkpoint-redis/src/store.ts
```diff
@@ -339,6 +303,51 @@ const SCHEMAS = [
...
+ * "_" and non-ASCII. Namespace labels reach the query from the caller, so a
+ * term built from anything else could close the clause and write its own --
+ * `@prefix:(`, `|` and `)` are all query syntax. Splitting on this makes that
+ * structurally impossible rather than escaped-and-hopefully-correct.
+ *
+ * It is not a model of RediSearch's tokeniser and does not need to be. A term
+ * that turns out not to be indexed simply finds nothing, and a search that
+ * finds nothing widens rather than reporting an empty namespace.
+ */
+const UNSAFE_IN_TERM = /[^a-zA-Z0-9_\u0080-\uffff]+/;
```
byhow: why do we need the unsafe in term here

libs/checkpoint-redis/src/store.ts
```diff
@@ -377,18 +386,92 @@ export class RedisStore {
...
+    } catch {
+      // Deliberately ignored; `probeExactNamespaceField` decides what is used.
+    }
+  }
+
+  async setup(): Promise<void> {
+    // Create store index
+    await this.ensureIndex(
+      SCHEMAS[0].index,
+      SCHEMAS[0].schema as Record<string, unknown>,
```
byhow: parse, dont validate or cast

libs/checkpoint-redis/src/store.ts
```diff
@@ -421,76 +504,209 @@ export class RedisStore {
...
+
+  /**
+   * Candidate rows for `query`, a page at a time.
+   *
+   * Callers confirm the exact namespace themselves, so a page may hold nothing
+   * they want and the row they are after can sit on any later page. Paging
+   * runs until RediSearch reports the candidate set exhausted; stopping
+   * earlier is the caller's business.
+   */
+  private async *candidatePages(
```
byhow: why async generators here

libs/checkpoint-redis/src/store.ts
```diff
@@ -421,76 +504,209 @@ export class RedisStore {
...
+  ): Promise<{ id: string; value: StoreDocument } | undefined> {
+    const candidates = this.namespaceQuery(prefix, false);
+    // The empty string is a usable key, but `@key:{}` is not a legal TAG
+    // filter, so there the candidate query has to carry the lookup alone.
+    const query =
+      key === ""
+        ? candidates
+        : `(${candidates}) (@key:{${this.escapeTagValue(key)}})`;
+
+    for await (const page of this.candidatePages("store", query))
```
byhow: are you building like a mini redis here or what is going on

libs/checkpoint-redis/src/store.ts
```diff
@@ -434,70 +463,101 @@ export class RedisStore {
...
+    };
+
+    const first = await run(narrowed);
+    // Keeping nothing is not the same as matching nothing. A narrowing that
+    // matched documents cannot have skipped others, so "none of them were
+    // yours" is the final answer. Only a query that matched nothing at all
+    // is ambiguous enough to be worth asking again without it.
+    if (first.kept.length > 0 || first.matched > 0 || narrowed === widened)
+      return first.kept;
+    return (await run(widened)).kept;
```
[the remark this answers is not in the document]
byhow: Confirmed and fixed, though not here — `namespaceTerms` no longer derives terms from fragments of a label. A label is taken apart only across characters RediSearch is measured to split on, and one carrying a character that neither splits nor stays inside a token (backslash, newline, carriage return) is skipped whole, which widens the candidate set rather than naming tokens the index never stored.

Verified on a container: `get` and `search` for `["tenant","a\\b"]` both return that document while `["tenant","a-b"]` keeps its own. Covered by an integration test over twelve labels that reduce to the same terms, and by one that re-derives the character classification from the engine on every run so the three groups cannot drift.

libs/checkpoint-redis/src/store.ts
```diff
@@ -659,80 +703,91 @@ export class RedisStore {
...
+        // rejected here -- before anything fetches the document it points
+        // at. Only survivors cost a round trip.
+        const nearest = async (query: string, k: number) => {
+          const results = await this.client.ft.search(
+            "store_vectors",
+            `(${query})=>[KNN ${k} @embedding $BLOB]`,
+            {
+              PARAMS: { BLOB: vectorBytes },
+              DIALECT: 2,
+              LIMIT: { from: 0, size: k },
```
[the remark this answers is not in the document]
byhow: Could not reproduce. On Redis Stack a `KNN 8` query returned `__embedding_score` ascending — 0.00e+0, 1.19e-7, 4.77e-7, 1.07e-6, 2.03e-6, 3.10e-6, 4.53e-6, 6.08e-6 — so the neighbours already arrive in distance order and slicing after filtering keeps the ranking.

Leaving it unchanged rather than adding a `SORTBY` on that evidence. If you have a case where the order differs — a different algorithm or dialect, perhaps — that would be worth seeing, since the page really would be wrong.

libs/checkpoint-redis/src/store.ts
```diff
@@ -659,80 +703,91 @@ export class RedisStore {
...
+            );
+        };
+
+        let hits = await nearest(queryStr, Math.max(wanted, 1));
+        // Too few of the nearest belonged to the caller: either the namespace
+        // is sparse among the vectors, or the narrowing missed it. Looking
+        // further out with no narrowing at all covers both. KNN cannot page,
+        // and RediSearch refuses a `k` above MAXSEARCHRESULTS, so this widens
+        // to a bounded ceiling rather than doubling until it fits.
+        if (hits.length < wanted) hits = await nearest("*", 1000);
```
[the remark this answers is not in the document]
byhow: Confirmed and fixed. Reproduced exactly as described: 1200 vectors under `tenant.A` and one under `tenant.a`, and `search(["tenant","a"], { query, limit: 1 })` returned `[]`.

The retry now asks for as many neighbours as the server will serve rather than a fixed thousand, so the caller's document is reached. Scoping before KNN would be better and needs an exact tag on the vector index, which is deliberately out of this patch; the ceiling is `MAXSEARCHRESULTS`, and beyond it the search still under-reports. A unit test fails if the ceiling is shrunk back.

libs/checkpoint-redis/src/store.ts
```diff
@@ -815,36 +871,21 @@ export class RedisStore {
...
+        )
+          continue;
+        if (
+          options?.suffix &&
+          !isWithinNamespace(
+            parts.slice().reverse().join("."),
+            options.suffix.slice().reverse().join(".")
+          )
+        )
+          continue;
```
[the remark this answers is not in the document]
byhow: Confirmed and reverted. All three cases reproduced: prefix `["tenant.a"]` resolved what `["tenant","a"]` holds, prefix `[""]` matched every namespace, and suffix `["a.tenant"]` matched after reversal.

`listNamespaces` is back to `main`'s segment-by-segment comparison. Routing it through `isWithinNamespace` was to avoid stating the rule twice, but flattening the filter reintroduced the aliasing this PR exists to remove, and the method is not otherwise part of the fix. An integration test now covers the three aliases.

PR #2830, opened 2026-09-11
fix(langgraph-checkpoint-mongodb): enforce namespace boundaries

byhow:
MongoDBStore joins namespace labels with `/` for uniqueness and vector filtering, so `["tenant", "a/b"]` can collide with `["tenant", "a", "b"]`. Ordinary searches and exact reads already use namespace arrays.

This focused fix rejects `/` inside labels on writes (including direct batches), verifies vector candidates against the original namespace array before returning them, and wraps listing labels in `$literal`. Existing slash-containing records remain readable and deletable. No migration API, data rewrite, new encoding fields, or index rebuild is introduced.

The validation change has a major changeset. Existing slash labels must be renamed before updates; old encoding collisions can still cause short vector pages or duplicate-key errors until cleaned up. See `libs/checkpoint-mongodb/namespace-upgrade.md`. The vector check prevents out-of-scope values from being returned even before cleanup.

Validation: 26 unit tests and four MongoDB Atlas Local integration tests pass, covering ordinary/vector segment boundaries, write rejection, same-key distinct namespaces, literal prefix/suffix listing, and both directions of legacy vector collisions without changing data or indexes. Source/test typechecks and scoped lint pass. Live vector tests use deterministic manual embeddings; automatic server embeddings have unit coverage only.

Stack: #2825 → #2828 → #2829 → this PR. Related: #2721; contributor context: #2803 and #2826. MongoDB behavior was reproduced independently; the Python advisory is not automatically an advisory for this npm package. This draft does not close those reports or contributor PRs.


libs/checkpoint-mongodb/namespace-upgrade.md
hntrl, line 1: ?

libs/checkpoint-mongodb/src/tests/namespace-isolation.int.test.ts
```diff
@@ -0,0 +1,150 @@
+import { getEnvironmentVariable } from "@langchain/core/utils/env";
+import { MongoClient } from "mongodb";
+import { afterAll, beforeAll, expect, it } from "vitest";
+import { MongoDBStore } from "../store.js";
+
+// Use a dedicated local Atlas deployment; each suite owns a unique database.
+const client = new MongoClient(
+  getEnvironmentVariable("MONGODB_URL") ??
+    "mongodb://127.0.0.1:57017/?directConnection=true"
+);
```
hntrl, line 10: is this through testcontainers? do we just always assume a local atlas instance is available?

.changeset/mongodb-namespace-encoding.md
```diff
@@ -0,0 +1,5 @@
+---
+"@langchain/langgraph-checkpoint-mongodb": major
```
hntrl, line 2: not major?

PR #2831, opened 2026-09-11
feat(sdk): expose stream connection lifecycle

### LINKED ISSUE #2817: useStream: dead SSE pump leaves isLoading=true, error unset, and cannot reconnect

ramon-langchain:
Fixes #2817

### Summary

- expose `onConnected` after initial and restored SSE/WebSocket connections
- include the scheduled delay in `onReconnect` so UIs can render retry state
- restart a settled root event pump on later submissions and clear `isLoading` when an active stream closes fatally
- forward lifecycle callbacks through React, Vue, Svelte, and Angular stream hooks

### Testing

- `pnpm --filter @langchain/langgraph-sdk test -- src/client/stream/transport/http.test.ts src/client/stream/transport/websocket.test.ts src/stream/controller.test.ts` (800 tests passed; no type errors)
- `pnpm exec oxlint <changed source files>`
- `pnpm exec oxfmt --check <changed files>`
- `pnpm turbo build:internal --filter=@langchain/langgraph-sdk --filter=@langchain/react --filter=@langchain/vue --filter=@langchain/svelte --filter=@langchain/angular` (11 tasks successful; existing fixture TS2883 diagnostics were emitted during declaration generation)

Made by [Open SWE](https://openswe.vercel.app/agents/094c8665-f733-5261-a25f-17392be76491) · openai:gpt-5.6-sol (high)

## References
- Plan: https://openswe.vercel.app/agents/094c8665-f733-5261-a25f-17392be76491/plan

Co-authored-by: Huangshuo Kuang <141250392+kkkhs@users.noreply.github.com>

### DIFF
--- a/.changeset/stream-connection-lifecycle.md
+++ b/.changeset/stream-connection-lifecycle.md
@@ -6,4 +6,8 @@
 "@langchain/angular": minor
 ---
 
-Expose successful stream connections through `onConnected`, include the scheduled delay in `onReconnect`, and allow stream controllers to restart their root pump after a fatal transport failure.
+Expose connection lifecycle callbacks for built-in streaming transports.
+
+`onConnected` runs after the initial SSE or WebSocket connection becomes usable and after every successful reconnect. Its payload distinguishes an `initial` connection from a `reconnected` connection and includes the reconnect attempt number.
+
+`onReconnect` now also receives the scheduled `delayMs`, allowing applications to display accurate retry state before the next connection attempt. React, Vue, Svelte, and Angular stream hooks forward both callbacks.
--- a/.changeset/stream-connection-lifecycle.md
+++ b/.changeset/stream-connection-lifecycle.md
@@ -1,0 +1,9 @@
+---
+"@langchain/langgraph-sdk": minor
+"@langchain/react": minor
+"@langchain/vue": minor
+"@langchain/svelte": minor
+"@langchain/angular": minor
+---
+
+Expose successful stream connections through `onConnected`, include the scheduled delay in `onReconnect`, and allow stream controllers to restart their root pump after a fatal transport failure.
--- a/libs/sdk/src/client/stream/transport/websocket.ts
+++ b/libs/sdk/src/client/stream/transport/websocket.ts
@@ -150,8 +150,7 @@ export class ProtocolWebSocketTransportAdapter implements TransportAdapter {
     this.onReconnected = handler;
   }
 
-  async open(): Promise<void> {
-    const reconnecting = this.reconnectInFlight != null;
+  async open(reconnecting = false): Promise<void> {
     if (this.closed) {
       throw new Error("Protocol WebSocket transport is closed.");
     }
@@ -416,7 +415,7 @@ export class ProtocolWebSocketTransportAdapter implements TransportAdapter {
       }
 
       try {
-        await this.open();
+        await this.open(true);
         if (this.onReconnected) {
           await this.onReconnected();
         }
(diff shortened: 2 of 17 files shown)


.changeset/stream-connection-lifecycle.md
```diff
@@ -0,0 +1,9 @@
+---
+"@langchain/langgraph-sdk": minor
+"@langchain/react": minor
+"@langchain/vue": minor
+"@langchain/svelte": minor
+"@langchain/angular": minor
+---
+
+Expose successful stream connections through `onConnected`, include the scheduled delay in `onReconnect`, and allow stream controllers to restart their root pump after a fatal transport failure.
```
hntrl: could you separate the root pump issue into a separate changeset? also fine to be a little bit more verbose here

hntrl (dismissed):

libs/sdk/src/client/stream/transport/websocket.ts
```diff
@@ -150,8 +150,7 @@ export class ProtocolWebSocketTransportAdapter implements TransportAdapter {
     this.onReconnected = handler;
   }
 
-  async open(): Promise<void> {
-    const reconnecting = this.reconnectInFlight != null;
+  async open(reconnecting = false): Promise<void> {
```
hntrl: why are you adding this parameter here? if theres a race condition with scheduling vs. sending the callback is there a way we can just reorder those steps?

hntrl (approved):

PR #2843, opened 2026-09-16
fix(sdk): reconcile remotely resolved interrupts

### LINKED ISSUE #2842: SDK v2: hydrated interrupt stays pending after another client resumes the thread
### Checked other resources

- [x] Descriptive title provided.
- [x] LangGraph/LangChain documentation searched.
- [x] GitHub issues searched; related reports found, but no duplicate of this case.
- [x] The standalone reproduction isolates the failure to the published SDK.
- [x] The failure reproduces on the latest stable SDK release, `1.11.0`.

### Example Code

This standalone reproduction uses the published SDK's `Client` and two `StreamController` instances with the built-in HTTP/SSE transport against a small, in-process protocol fixture. The fixture models an interrupted thread, accepts an `input.respond` command from one controller, and broadcasts the resumed run's state and completion to both controllers. All identifiers and state are synthetic; no credentials, model, or external server are required.

In an empty directory, install the dependencies, save the code below as `repro.mjs`, and run it:

```sh
npm init -y
npm install --save-exact --ignore-scripts --no-audit --no-fund @langchain/langgraph-sdk@1.11.0 @langchain/core@1.2.11
node repro.mjs
```

```javascript
import assert from "node:assert/strict";
import { createServer } from "node:http";
import { setTimeout as delay } from "node:timers/promises";
import { StreamController } from "@langchain/langgraph-sdk/stream";

const threadId = "00000000-0000-4000-8000-000000000001";
const interruptId = "0123456789abcdef0123456789abcdef";
let state = {
  values: { messages: [], phase: "approval" },
  next: ["approval"],
  tasks: [{ id: "approval-task", name: "approval",
    interrupts: [{ id: interruptId, value: { question: "Continue?" } }] }],
  checkpoint: { thread_id: threadId, checkpoint_ns: "", checkpoint_id: "cp-1" },
  parent_checkpoint: null,
  metadata: { step: 1 },
};
const streams = new Set();
const events = [];
let seq = 0;
const matches = (channels, event) =>
  channels.includes(event.method.split(".")[0]);
function frame(response, event) {
  response.write(`id: ${event.seq}\nevent: ${event.method}\ndata: ${JSON.stringify(event)}\n\n`);
}
function emit(method, data) {
  const event = { type: "event", seq: ++seq, event_id: `event-${seq}`,
    method, params: { namespace: [], data } };
  events.push(event);
  for (const stream of streams)
    if (matches(stream.channels, event)) frame(stream.response, event);
}
async function waitFor(condition, description) {
  const deadline = Date.now() + 5000;
  while (!condition()) {
    if (Date.now() > deadline) throw new Error(`Timed out: ${description}`);
    await delay(5);
  }
}
const server = createServer(async (request, response) => {
  const path = new URL(request.url, "http://fixture.invalid").pathname;
  let body = "";
  for await (const chunk of request) body += chunk;
  const json = body ? JSON.parse(body) : {};
  const send = (value) => {
    response.writeHead(200, { "Content-Type": "application/json" });
    response.end(JSON.stringify(value));
  };
  if (path.endsWith("/state")) return send(state);
  if (path.endsWith("/history")) return send([state]);
  if (path.endsWith("/stream/events")) {
    response.writeHead(200, { "Content-Type": "text/event-stream" });
    response.write(": connected\n\n");
    const stream = { response, channels: json.channels };
    streams.add(stream);
    response.on("close", () => streams.delete(stream));
    for (const event of events)
      if (event.seq > (json.since ?? 0) && matches(stream.channels, event))
        frame(response, event);
    return;
  }
  if (path.endsWith("/commands") && json.method === "input.respond") {
    assert.equal(json.params.interrupt_id, interruptId);
    state = { ...state, values: { messages: [], phase: "working" },
      next: ["model"], tasks: [{ id: "model-task", name: "model", interrupts: [] }],
      checkpoint: { ...state.checkpoint, checkpoint_id: "cp-2" },
      metadata: { step: 2 } };
    send({ type: "success", id: json.id, result: { run_id: "run-2" },
      applied_through_seq: seq });
    setImmediate(() => {
      emit("lifecycle", { event: "running" });
      emit("values", state.values);
    });
    return;
  }
  response.writeHead(404).end();
});
await new Promise(resolve => server.listen(0, "127.0.0.1", resolve));
const apiUrl = `http://127.0.0.1:${server.address().port}`;
const client = new Client({ apiUrl, apiKey: null, streamProtocol: "v2",
  callerOptions: { maxRetries: 0 } });
const options = { client, assistantId: "agent", threadId,
  maxReconnectAttempts: 0 };
const observer = new StreamController(options);
const responder = new StreamController(options);
let reopened;
const ids = controller => controller.rootStore.getSnapshot().interrupts.map(i => i.id);
try {
  await Promise.all([observer.hydrationPromise, responder.hydrationPromise]);
  await waitFor(() => streams.size === 4, "both clients subscribe");
  assert.deepEqual(ids(observer), [interruptId]);
  await responder.respond("continue", { interruptId });
  await waitFor(() => observer.rootStore.getSnapshot().values.phase === "working"
    && observer.rootStore.getSnapshot().isLoading, "observer receives running state");
  console.log("During resumed run:", JSON.stringify({
    observer: ids(observer), responder: ids(responder),
    observerIsLoading: observer.rootStore.getSnapshot().isLoading }));
  assert.deepEqual(ids(responder), []);
  state = { ...state, values: { messages: [], phase: "done" }, next: [], tasks: [] };
  emit("values", state.values);
  emit("lifecycle", { event: "completed" });
  await waitFor(() => observer.rootStore.getSnapshot().values.phase === "done"
    && !observer.rootStore.getSnapshot().isLoading, "observer receives completion");
  // Keep the streams open and allow asynchronous terminal reconciliation to settle.
  await delay(1000);
  const serverState = await client.threads.getState(threadId);
  assert.deepEqual(serverState.tasks, []);
  reopened = new StreamController(options);
  await reopened.hydrationPromise;
  console.log("After completion:", JSON.stringify({
    observer: ids(observer), responder: ids(responder), reopened: ids(reopened),
    serverTasks: serverState.tasks }));
  assert.equal(observer.rootStore.getSnapshot().error, undefined);
  assert.deepEqual(ids(reopened), []);
  assert.deepEqual(ids(observer), [], "Observer must clear the resolved interrupt");
} finally {
  await Promise.all([observer.dispose(), responder.dispose(), reopened?.dispose()]);
  for (const stream of streams) stream.response.end();
  server.closeAllConnections();
  await new Promise(resolve => server.close(resolve));
}
```

The setup above was verified in a fresh temporary npm project. The final assertion checks the expected behavior and fails on the affected SDK. The streams remain open for one second after completion to allow asynchronous terminal reconciliation to settle. The test waits for observed state changes before making assertions; it does not invoke or replace private SDK methods.

```text
During resumed run: {"observer":["0123456789abcdef0123456789abcdef"],"responder":[],"observerIsLoading":true}
After completion: {"observer":["0123456789abcdef0123456789abcdef"],"responder":[],"reopened":[],"serverTasks":[]}

AssertionError [ERR_ASSERTION]: Observer must clear the resolved interrupt
+ actual - expected

+ [
+   '0123456789abcdef0123456789abcdef'
+ ]
- []
```

There is no transport or server error; `observer.rootStore.getSnapshot().error` remains `undefined`.

### Description

An already-hydrated interrupt remains in the observing controller's `interrupts` after a different controller answers it and the resumed run completes.

Steps exercised by the reproduction:

1. Hydrate two controllers for the same interrupted thread. Both see interrupt A.
2. Call `respond()` through the second controller.
3. The first controller receives live `lifecycle: running` and updated `values`. Its `isLoading` becomes true, but interrupt A remains.
4. The first controller receives the final values and `lifecycle: completed`; `isLoading` becomes false, but interrupt A still remains.
5. An independent `getState()` reports no pending tasks, and a new controller hydrated from that state has an empty interrupt list. The original observer still contains A.

**Expected:** the observer reconciles its pending interrupts with the completed run's authoritative state and removes A. It should not require a local response, a new interrupt, or recreating the controller to discard an already-resolved interrupt.

**Actual:** the responding controller and a newly hydrated controller agree with the server, but the already-connected observer retains the resolved interrupt in `interrupts` despite receiving the resumed run's state and completion.

This reproduces without historical lifecycle replay, dropped connections, or any new `input.requested` event. It is an SDK protocol-level reproduction; the fixture does not execute a LangGraph graph.

**Relevant implementation / related fixes**

- Local `respond()` marks the targeted interrupt resolved, but that knowledge stays in the responding controller.
- In SDK `1.11.0`, `#settleParkedInterruptsOnTerminal()` returns early when `#pendingHydrationInterruptEvents` is empty. Here A was already present during hydration, and no new interrupt arrives, so there are no parked interrupts to trigger that reconciliation.
- #2813 fixes newly raised interrupts after passive rejoin by reconciling parked events. This case concerns removal of an interrupt that was already hydrated.
- #2808 covers locally resolved interrupts versus remote state. The observer in this reproduction never submits a response locally.

The same standalone HTTP fixture also reproduces the failure on SDK `1.9.29`.

### System Info

- Node.js `24.19.0`
- npm `12.0.2`
- Latest release tested: `@langchain/langgraph-sdk@1.11.0`
- Clean test environment: `@langchain/core@1.2.11`, `@langchain/protocol@0.0.19`
- Also reproduced on `@langchain/langgraph-sdk@1.9.29`
- Built-in SSE transport, `streamProtocol: "v2"`

jackjin1997:
## Summary
Fixes #2842
- reconcile hydrated interrupts when a terminal lifecycle event comes from a run resumed by another client
- keep polling authoritative thread state until a remotely completed run is committed, while preserving interrupted-state reconciliation
- add regression coverage for the cross-client completion race

## Testing
- `pnpm --dir libs/sdk exec vitest run src/stream/controller.test.ts` (69 passed)
- `pnpm exec oxlint libs/sdk/src/stream/controller.ts libs/sdk/src/stream/controller.test.ts`
- `pnpm exec oxfmt --check libs/sdk/src/stream/controller.ts libs/sdk/src/stream/controller.test.ts .changeset/calm-threads-reconcile.md`
- `pnpm --dir libs/sdk build`

## AI Disclosure
This bug was identified and the fix was implemented with AI assistance.


jackjin1997:
The first CI run caught a regression in the live nested-interrupt browser test: terminal reconciliation was considering every currently displayed interrupt, including one raised live after hydration.

Commit `9cffa2d` narrows that path to IDs seeded by hydration, so live nested interrupts remain visible while stale hydrated interrupts are still reconciled. I reran the focused Chrome browser suite (`stream.nested-interrupts.test.tsx`: 3/3 passed), the controller suite (69/69 passed), oxlint, and oxfmt locally.

PR #2844, opened 2026-09-16
fix(sdk): surface message-finish response metadata on streamed messages

zduric-langchain:
Fixes LG-690

## Summary

`stop_reason` did not reach AI messages assembled from a live stream, so a client could not tell a message that ends the turn apart from one about to call a tool until the run finished or the page reloaded.

Two things caused it:

- `MessageAssembler.consume` read the finish payload from `data.responseMetadata`, which is the `@langchain/core` in-process `MessageFinishEvent` spelling. The wire event spells it `data.metadata`, so `AssembledMessage.finishMetadata` was always `undefined`. `MessageFinishData` declares neither key and intersects `Extensible = Record<string, any>`, so nothing caught it.
- `assembledToBaseMessage` never read `finishMetadata` at all. It set `response_metadata` to a fixed `{ output_version: "v1" }` on every AI message.

Fixing either alone changes nothing, so both are in here.

## Changes

- `readFinishMetadata` accepts both spellings and merges them, `metadata` winning. Applied in `MessageAssembler.consume` and in `StreamingMessage.#assembleMessage`, which had the same mismatch.
- `finishMetadata` threaded through `AssembledToMessageInput` into `response_metadata`, with `output_version` still pinned last so a provider-supplied `output_version` such as `responses/v1` cannot win.
- A legacy top-level `reason` still maps to `finish_reason`, but loses to `metadata` when both arrive. The protocol dropped that field before this code was written, so it has been inert against current servers.

This matches what `langchain_core.language_models.chat_model_stream` already does for the same protocol events, which is why the Python SDK was unaffected.

## Testing

Parameterized regression test over both wire spellings in `libs/sdk/src/client/stream/messages.test.ts`. It fails on both parameters without the change.

Full SDK suite: 802 passed across 53 files. The one `langsmith_tracer` typecheck error in `src/client/stream/index.test.ts` is pre-existing and reproduces identically on `main`.

Verified end to end by replaying a real captured event stream through a patched build: `stop_reason` comes through as `end_turn` where it was previously absent.

## Known gaps, not addressed here

`assembledToBaseMessage` still does not derive `model_name` / `model_provider` from `message-start` metadata, and puts token usage on `additional_kwargs.usage` rather than `usage_metadata`. Both make a streamed message differ from the checkpointed one. Separate asks.


ccurme (approved):

zduric-langchain:
fixed in a parallel pr

PR #2848, opened 2026-09-17
feat(sdk): back useStream's enqueue strategy with a server-backed queue adapter

thushanth-bengre-langchain:
* backs `useStream`'s `enqueue` multitask strategy with real, durable server-side runs instead of an in-memory client-only queue
  * opt-in via `serverQueue: true` (built-in transport) or by setting `serverQueue` directly on a custom `AgentServerAdapter` and omitting it keeps today's behavior unchanged
  * queued submissions become server-accepted runs immediately (`multitaskStrategy: "enqueue"`), hydrated from the Runs API on load, and cancelled server-side through `cancelQueued`/`clearQueue`


libs/sdk/src/client/stream/transport.ts
```diff
@@ -136,4 +137,15 @@ export interface AgentServerAdapter extends TransportAdapter {
...
+  /**
+   * Backs `useStream`'s `"enqueue"` multitask strategy with real, durable
+   * runs instead of an in-memory client-only queue. Omit it and
+   * `"enqueue"` keeps today's client-only defer behavior, unchanged.
+   *
+   * Only opted into explicitly here or via `serverQueue: true` on
+   * `useStream`; never inferred from a transport's ability to reach a
+   * real Agent Server.
+   */
+  serverQueue?: Pick<RunsClient, "create" | "list" | "cancel">;
```
hntrl: serverQueue feels like a little bit of awkward naming? I was originally thinking the relationship is queue -> {local, lsd}

I also don't know if the "interface of truth" is the runs client, just so we isolate things and keep local vs. SDK clients nice and tidy it should be its own interface? e.g. server queue talks to RunsClient in its own impl, client queue talks to in memory maps in its own impl

libs/sdk/src/client/stream/transport.ts
```diff
@@ -136,4 +137,15 @@ export interface AgentServerAdapter extends TransportAdapter {
...
   >;
+
+  /**
+   * Backs `useStream`'s `"enqueue"` multitask strategy with real, durable
+   * runs instead of an in-memory client-only queue. Omit it and
+   * `"enqueue"` keeps today's client-only defer behavior, unchanged.
+   *
+   * Only opted into explicitly here or via `serverQueue: true` on
+   * `useStream`; never inferred from a transport's ability to reach a
+   * real Agent Server.
```
hntrl: comment should be updated in accordance with the interface comments im mentioning below

libs/sdk/src/client/runs/index.test.ts
```diff
@@ -94,3 +94,47 @@ describe("runs.cancelMany", () => {
...
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  function parseFetchCall() {
+    const [url] = fetchMock.mock.calls[0];
+    return { url: new URL(url) };
+  }
+
+  it("sends an array-valued param (select) as repeated query entries, not one JSON string", async () => {
```
hntrl, line 104: nice

libs/sdk/src/client/runs/index.test.ts
```diff
@@ -94,3 +94,47 @@ describe("runs.cancelMany", () => {
...
   });
 });
+
+describe("runs.list", () => {
+  let fetchMock: MockFetch;
+
+  beforeEach(() => {
+    fetchMock = createMockFetch();
+    overrideFetchImplementation(fetchMock);
+    (globalThis as any).fetch = fetchMock;
```
hntrl: is it possible to just set the fetch override on the client class itself? is that an option we expose

libs/sdk/src/stream/controller.ts
```diff
@@ -650,6 +657,9 @@ export class StreamController<
       if (this.#disposed || this.#currentThreadId !== hydratedThreadId) return;
       threadExists = state != null;
       threadActive = isThreadStateActive(state);
+      // Refined to `false` below once `state.tasks` is parsed, if it turns
+      // out this "active" state is actually just a parked interrupt.
+      threadActiveWithoutInterrupt = threadActive;
```
hntrl: this feels like one of those things that are pretty easily going to fall out of date since theres a lot of client updates we submit..

could you explain what this is for? it might be justified, I'm not totally sure

libs/sdk/src/stream/dispatch-config.ts
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * The platform expects `config.configurable.thread_id` on every run
+ * dispatch; we set it last so user-supplied values can't accidentally
+ * override the active thread id (which would route the run to a
+ * different thread).
+ */
+export function bindThreadConfig(
+  config: unknown,
+  threadId: string
+): Record<string, unknown> {
```
hntrl, line 11: why do we have to propagate thread id through runnable?

libs/sdk/src/stream/queue-adapter-agent-server.ts
```diff
@@ -0,0 +1,297 @@
...
+    ]);
+    try {
+      const run = await this.#runs.create(threadId, this.#assistantId, {
+        input: values,
+        config: bindThreadConfig(options?.config, threadId),
+        metadata: options?.metadata,
+        checkpointId: options?.forkFrom,
+        multitaskStrategy: "enqueue",
+      });
+      if (this.#pendingCancel.delete(id)) {
```
hntrl, line 134: we only cancel when another enqueue value is sent? is the idea that sending cancels across also sends enqueues

libs/sdk/src/stream/queue-adapter-local.ts
hntrl, line 1: i'm surprised its that easy

libs/sdk/src/client/stream/transport.ts
thushanth-bengre-langchain: Removed serverQueue for setting local/server.
Made it derived from `transport`. Split the shared `Pick` into 2 named types, so each adapter contract is explicit instead of one shape.

libs/sdk/src/client/stream/transport.ts
thushanth-bengre-langchain: Comment updated

libs/sdk/src/client/runs/index.test.ts
thushanth-bengre-langchain: Fixed this. This test now passes fetch directly instead od mutating global fetch state.

libs/sdk/src/stream/controller.ts
thushanth-bengre-langchain: We needed a way to set `isLoading` on hydration, but couldn't just use `threadActive`.
If there is an interrupt, we cannot set `isLoading` to true. So this was to track that combined value. 

But I found a simpler way to do this. Now using `hasActiveInterrupts` instead.

libs/sdk/src/stream/dispatch-config.ts
thushanth-bengre-langchain: This is not a new logic. Just moved it to it's own file so the enqueue path can use it. Server uses `config.configurable.threadId` to read which threads checkpoint to update. so every run needs it.

libs/sdk/src/stream/queue-adapter-agent-server.ts
thushanth-bengre-langchain: This is solving a race condition between enqueue and cancel. Cancel needs a `run_id`, which is only set once the async enqueue responds. So if a run that's in flight for enqueue gets cancelled in the meantime, we record it in pendingCancel, and once enqueue resolves, we check if that id is up for cancellation If it is, cancel the run instead of letting it get into the queue as if nothing happened.

libs/sdk/src/stream/queue-adapter-local.ts
thushanth-bengre-langchain: Yes. Since it's just in-memory array, no race conditions to handle here.

libs/sdk/src/client/stream/transport.ts
thushanth-bengre-langchain: Update: Reverted the "derived from transport" part here.

This is a real breaking change if we derived from transport. Users will get server capability without asking for one and I believe this should be opt-in. 
If they are using a proxy to talk to runs API, they have to proxy the create, list, cancel (`POST`/`GET /threads/{id}/runs`, `POST /threads/{id}/runs/{run_id}/cancel` API endpoints otherwise they will see 404 failures without knowing why.

So for this version, the only source of truth will be `queue: local(default) | server` on `useStream`.
If that is not set, fall back to existing behavior(local in-memory queue) but no errors. 

runs client is not the "interface of truth" and the local and server-backed implementations each have their own interface.

hntrl (approved):
retroactive stamp- we merged early to cut rc

